PR #14, opened 2015-02-16
Deprecate base_path in body in the publishing context

jamiecobbett:
The documentation implies that the base_path must be included in the request
body for write requests as well as in the URL. This is not true and in fact if
supplied in the body it is discarded by content-store, otherwise it is ignored
- the base_path in the URL is used instead.

Leading developers to believe that it is required but not performing any
validation on it seems likely to be a source of errors, so let's get rid of it.

### DIFF
--- a/formats/metadata.json
+++ b/formats/metadata.json
@@ -3,7 +3,6 @@
   "type": "object",
   "additionalProperties": false,
   "required": [
-    "base_path",
     "format",
     "publishing_app",
     "rendering_app",
@@ -11,9 +10,6 @@
     "locale"
   ],
   "properties": {
-    "base_path": {
-      "type": "string"
-    },
     "content_id": {
       "type": "string",
       "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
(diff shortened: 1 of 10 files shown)


jamiecobbett:
One wrinkle is that the `base_path` will be included in the response body to a PUT.

lib/govuk_content_schemas/frontend_schema_generator.rb
```diff
@@ -39,14 +39,15 @@ def internal?(property_name)
 
   def required_properties
     return [] unless @publisher_schema.schema.has_key?('required')
-    @publisher_schema.schema['required'].reject { |property_name| internal?(property_name) }
+    (@publisher_schema.schema['required'] - INTERNAL_PROPERTIES) + ['base_path']
```
vinayvinay: good optimisation! :+1:  this could be repeated in `frontend_properties` as well, and then we could get rid of the `internal?` method.

formats/metadata.json
```diff
@@ -11,9 +10,6 @@
     "locale"
   ],
   "properties": {
-    "base_path": {
-      "type": "string"
-    },
```
vinayvinay: maybe i'm missing something, but if metadata.json is supposed to be the master definition, and shared across publishers and frontend [^1], shouldn't we be preserving `base_path` here? at least as an optional field?

[1] https://github.com/alphagov/govuk-content-schemas#publisher-schema-defined-using-component-parts

PR #22, opened 2015-03-06
[DRAFT] Document contract testing

heathd:
this is a draft of documentation about contract testing for feedback

/cc @tommyp @danielroseman

PR #23, opened 2015-03-09
Disallow unspecified properties in finders

jamiecobbett:
Allowing unspecified properties leaves the way open for typos in field names,
undocumented fields and other errors. I found some problems in contacts-admin
after making this change.

I've also added an example for a Contact finder as they are a bit different.

cc @evilstreak @edds @tommyp

### DIFF
--- a/formats/finder/publisher/details.json
+++ b/formats/finder/publisher/details.json
@@ -12,6 +12,9 @@
       "type": "string",
       "additionalProperties": false
     },
+    "email_signup_enabled": {
+      "type": "boolean"
+    },
     "filter": {
       "description": "This is the fixed filter that scopes the finder",
       "type": "object",
--- a/formats/finder/publisher/details.json
+++ b/formats/finder/publisher/details.json
@@ -19,6 +19,12 @@
       "properties": {
         "document_type": {
           "type": "string"
+        },
+        "organisations": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       }
     },
--- a/formats/finder/publisher/details.json
+++ b/formats/finder/publisher/details.json
@@ -13,8 +13,14 @@
       "additionalProperties": false
     },
     "filter": {
+      "description": "This is the fixed filter that scopes the finder",
       "type": "object",
-      "additionalProperties": false
+      "additionalProperties": false,
+      "properties": {
+        "document_type": {
+          "type": "string"
+        }
+      }
     },
     "facets": {
       "type": "array",
@@ -41,6 +47,12 @@
           "name": {
             "type": "string"
           },
+          "preposition": {
+            "type": "string"
+          },
+          "short_name": {
+            "type": "string"
+          },
           "type": {
             "type": "string"
           },
@@ -65,6 +77,12 @@
           }
         }
       }
+    },
+    "show_summaries": {
+      "type": "boolean"
+    },
+    "summary": {
+      "type": "string"
     }
   }
 }
--- a/formats/finder/publisher/details.json
+++ b/formats/finder/publisher/details.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
-  "additionalProperties": true,
+  "additionalProperties": false,
   "required": [
     "document_noun",
     "filter",
@@ -14,13 +14,13 @@
     },
     "filter": {
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": false
     },
     "facets": {
       "type": "array",
       "items": {
         "type": "object",
-        "additionalProperties": true,
+        "additionalProperties": false,
         "required": [
           "key",
           "filterable",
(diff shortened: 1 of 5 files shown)


evilstreak:
Whilst I approve of the intent to use stricter interfaces, I'm keen to avoid a repeat of the govuk_content_models bump and upgrade dance we have with the legacy publishing stack. I'd be interested to see the "Suggested workflows" docs added to the README to see how you're expecting this to be used.

tekin:
@evilstreak some additional docs from @heathd on contract testing have just been added here, although it still has gaps: https://github.com/alphagov/govuk-content-schemas/pull/22.

I also endorse this change. I think it's important that we're very explicit about what can and can't be included with a given format - the utility of the contract testing will be somewhat lost if there is ambiguity in the form of a given format. Because we don't have an actual explicit gem dependency, the dance won't be nearly as bad, and also has scope for the gradual and independent migration of a format (e.g. adding new (optional) fields to the schema for a format here in `govuk-content-schemas` before updating both the publishing and the frontend application to generate/handle the new data separately/sequentially).

formats/finder/publisher/details.json
```diff
@@ -12,15 +12,30 @@
...
     },
+    "email_signup_enabled": {
+      "type": "boolean"
+    },
     "filter": {
+      "description": "This is the fixed filter that scopes the finder",
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": false,
+      "properties": {
```
tekin: This should also include a "format" property (the [policies finder](https://github.com/alphagov/policy-publisher/blob/master/lib/policies_finder_publisher.rb#L62-L64) filters on `"format": "policy"`) otherwise the [spec for the finder](https://github.com/alphagov/policy-publisher/blob/master/spec/lib/policies_finder_publisher_spec.rb) will fail.

jamiecobbett:
@evilstreak I've made a stab at workflows here: https://github.com/alphagov/govuk-content-schemas/pull/25

tekin:
I *think* the finders schema will need further updating to match reality - the frontend app is expecting fields related to beta messaging which are not present here in the schema.

jamiecobbett:
@tekin/@tommyp if you're correct, that suggests that the publishing app (specialist-publisher?) isn't exercising all the behaviours when validating against the schema.

tekin:
No, I don't think `specialist-publisher` is wired into the content schemas workflow at present.

tommyp:
As it currently stands, Finders require several fields. All the existing Finders (published from Specialist Publisher) link off to Specialist Frontend which fetches the Finder Content Item and needs a piece of extra information (a format name to link back to the Finder). The Finder also has a bunch of optional fields that Finder Frontend is set up to deal with (`beta`, `beta_message`, `summary`) if they aren't present. Specialist Publisher doesn't do contract testing yet.

Contacts Admin and Policy Publisher are now both publishing what is a Finder which very little variation from the existing Finder format. Do we end wanting to have a slightly different schema for each or should we follow this approach with naming the extra properties requires?

Also, was I meant to follow the approach in this PR with the schema or was I meant to only list the required fields and allow anything else to go?

If it's the former, can I suggest we get some documentation about _writing_ a schema as my first pass was a complete stab in the dark at it.

tekin:
I've just added a schema for the "policy" format (https://github.com/alphagov/govuk-content-schemas/pull/27). 

As you say, it's essentially a finder, but it's also a distinct "format", so should be spec'd separate with its own schema as currently, govuk-content-schemas works based on the "format" field, which is distinct in this case.

This does mean there is going to be some duplication across these format schemas, but it makes it clearer what each format is, and also makes it possible to evolve them separately and makes it easier for the appropriate frontend app to test against the different variation examples, based on the format.

We definitely need to evolve the documentation around working with govuk-content-schemas as it's far from clear at present. As you say @tommyp, it wasn't clear if the schema should list all the fields and disallow additional ones. That is what this pull request aims to address. The README and docs have recently been updated (there's now a page for [How to add a new content format](https://github.com/alphagov/govuk-content-schemas#how-to-add-a-new-content-format) and also [Suggested workflows](https://github.com/alphagov/govuk-content-schemas#suggested-workflows), but we can (and should) definitely keep working on improving these.

tekin:
If we are saying that unspecified properties should not be allowed, maybe the schema validation should complain if either the `details` hash or the `links` array contains have `additionalProperties` set to `true`?

tommyp:
@tekin I've had a chat with @jamiecobbett and @heathd about this and it sounds like there needs to some further conversation about this.

heathd:
@tekin I don't think we should have the `make` task or build process enforce the requirement to set `additionalProperties` false, as there will probably be some exceptions. We'll continue to work on improving the documentation, and see if we can improve the feedback/output from the make task.

heathd:
given that this build is passing with all the downstream contract tests, I'd like to go ahead and merge it.

If there are issues relating to fields which are produced by `specialist-publisher` when it publishes finders, we can resolve those when the contract tests are set up for specialist publisher.

evilstreak:
@heathd :+1: Go ahead

PR #24, opened 2015-03-09
Add base policies finder as an example

tekin:
This is mainly so that we have a finder example that filters on format.

### DIFF
--- a/formats/finder/frontend/examples/policies.json
+++ b/formats/finder/frontend/examples/policies.json
@@ -1,0 +1,37 @@
+{
+   "base_path": "/government/policies",
+   "format":"finder",
+   "title":"Policies",
+   "description":"",
+   "updated_at": "2015-02-20T10:31:55.529Z",
+   "public_updated_at":"2015-03-09T14:09:25.000+00:00",
+   "locale":"en",
+   "details":{
+      "document_noun":"policy",
+      "email_signup_enabled":false,
+      "filter":{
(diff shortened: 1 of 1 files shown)


formats/finder/frontend/examples/policies.json
```diff
@@ -0,0 +1,37 @@
...
+   "title":"Policies",
+   "description":"",
+   "updated_at": "2015-02-20T10:31:55.529Z",
+   "public_updated_at":"2015-03-09T14:09:25.000+00:00",
+   "locale":"en",
+   "details":{
+      "document_noun":"policy",
+      "email_signup_enabled":false,
+      "filter":{
+         "format":"policy"
```
jamiecobbett: Is `format` distinct from `document_type`? [contacts-admin](https://github.com/alphagov/contacts-admin/blob/master/app/presenters/contacts_finder_presenter.rb#L50) and [specialist-publisher](https://github.com/alphagov/specialist-publisher/blob/master/finders/metadata/cma-cases.json#L7) use `document_type`. If they are describing the same thing, we should standardise.

tekin: Ooops, looks like my local copy of policy-publisher was slightly out of date - this was recently changed: https://github.com/alphagov/policy-publisher/commit/556776f5d7d3f52b9e877daf1c9da6156de29466. Closing this as the policies finder no longer seems that special.

PR #27, opened 2015-03-10
Policy format

tekin:
This adds the "policy" format to the schema, with two examples: one each of a "Policy area" and a "Policy programme". There is little material difference between these two examples at present, but they will diverge soon once we are including the link back to the parent "area" from the "programme". This hasn't been implemented yet, at least not in the representation of the policy being sent to content store.

Next steps are to update the policy-publisher app to generate policies that match this schema and also to test against it. Same for finder-frontend.


formats/policy/publisher/schema.json
```diff
@@ -0,0 +1,200 @@
...
+      }
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+      }
+    }
+  }
+}
```
rboulton: Missing newline at end of file.

PR #29, opened 2015-03-13
Make examples have distinct base_paths

jamiecobbett:
This means that all of the existing examples can be uniquely identified by the
base_path. This makes it easier to tell which is which, and also means that the
base_path can be used as a key in a hash or test name.


heathd:
would be nice to add a validation/requirement that the base paths are unique

PR #30, opened 2015-03-16
Update policy schema to include people

tekin:
Adds "people" to the list of valid links on a policy


heathd:
I think you should have made `people` an optional link rather than a required one, no?

tekin:
Cheers @heathd, now fixed.

PR #35, opened 2015-03-19
Add nation_applicability to Policies

tommyp:
This commit updates the schema for Policies to have nation applicability. The `nation_applicability` hash contains an array of strings which the nations that the Policy applies to along with an array of hashes with the nation and alt_poicy_url for each nation that the policy isn't applicable to.

### DIFF
--- a/formats/policy/frontend/examples/policy_area.json
+++ b/formats/policy/frontend/examples/policy_area.json
@@ -19,7 +19,16 @@
       "signup_link": "",
       "summary": "The government believes that the current benefits system is too complex, and there are insufficient incentives to encourage people on benefits to start paid work or increase their hours.",
       "show_summaries": false,
-      "facets":[]
+      "facets":[],
+      "nation_applicability": {
+        "applies_to": [
+          "england",
+          "northern_ireland",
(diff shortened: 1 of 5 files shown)


tekin:
Is this a good point to add a separate example so that we have one for a standard policy (applies to all nations, probably called just `policy.json`) and one for a policy that applies only to specific nations (say `inapplicable_policy.json`)? Those examples can be used in the schema tests to ensure the frontend properly handles both variants.

tommyp:
Since we have an 2 examples here, one for Policy Area and one for Programme, would it be easy to just make one apply to all nations and the other apply to only one?

tekin:
@tommyp I think keeping those two as they are (representing the common case for each variant) is probably a good idea to ensure we have good coverage in the examples and add a separate one to represent a policy with inapplicable nations. I don't think it matters if this is a policy area or programme, just that it's appropriately named (e.g. `policy_with_inapplicability.json`)

tommyp:
Updated the existing examples to be applicable to all nations and added an additional example which is only applicable to certain nations.

PR #36, opened 2015-03-20
Add minimal policy area example

edds:
One which doesn't include so many links.

### DIFF
--- a/formats/policy/frontend/examples/minimal_policy_area.json
+++ b/formats/policy/frontend/examples/minimal_policy_area.json
@@ -1,0 +1,24 @@
+{
+   "base_path": "/government/policies/minimal-benefits-reform",
+   "title": "Minimal benefits reform",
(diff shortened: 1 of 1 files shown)


formats/policy/frontend/examples/minimal_policy_area.json
```diff
@@ -0,0 +1,38 @@
...
+   "details": {
+      "document_noun": "document",
+      "email_signup_enabled": false,
+      "filter": {
+         "policies": [
+            "benefits-reform"
+         ]
+      },
+      "human_readable_finder_format": "Policy",
+      "signup_link": "",
```
elliotcm: I don't think we need these on policies.

edds: Fair. I have made the minimal even more minimal.

PR #37, opened 2015-03-20
Add document_collections to links

vinayvinay:
https://trello.com/c/pfvFRgg5

`government-frontend` doesn't display document collection membership next to 'Part of:' in the metadata section in the header or footer as shown here:
https://www.gov.uk/government/case-studies/clever-engineering-macrete-bridges-the-technology-gap

this field should carry that bit of information so that it gets displayed on government-frontend.

### DIFF
--- a/formats/case_study/frontend/examples/archived.json
+++ b/formats/case_study/frontend/examples/archived.json
@@ -63,6 +63,15 @@
         "web_url": "https://www.gov.uk/government/case-studies/terence-age-33-stoke-on-trent",
         "locale": "en"
       }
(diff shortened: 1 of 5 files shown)


jamiecobbett:
I think the build error is because you edited the generated schema.json files by hand (I did exactly the same thing), so you can rebase the two commits into one.

We do have a ticket in the backlog to split the generated files into a separate directory to make it more obvious that they are generated.

vinayvinay:
@jamiecobbett: i've squashed commits, thanks!

PR #40, opened 2015-04-15
Make public_updated_at required for renderable content

jamiecobbett:
This is currently optional in content-store, making it required here is a route
to making it required there (and then backfilling the data).

We have been thinking of public_updated_at as "the updated date which users
see" and because some formats don't display this (eg gone/redirect items), it
has been optional.

However, it seems to me that it isn't about what the user sees, but when there
was last a (significant) semantic change to the content at the URL. Therefore,
everything should have a public_updated_at.

As an example, it being optional meant that a placeholder without a
public_updated_at was added by publisher which then broke an expectation in
email-alert-service and stopped it from consuming any more messages from the
queue.

To make the Whitehall tests pass for "coming soon" items, we'd need to do something like: https://github.com/alphagov/whitehall/compare/supply_public_updated_at_for_coming_soon?expand=1

### DIFF
--- a/formats/case_study/frontend/schema.json
+++ b/formats/case_study/frontend/schema.json
@@ -5,7 +5,8 @@
   "required": [
     "base_path",
     "format",
-    "locale"
+    "locale",
+    "public_updated_at"
   ],
   "properties": {
     "base_path": {
--- a/formats/case_study/publisher/schema.json
+++ b/formats/case_study/publisher/schema.json
@@ -7,7 +7,8 @@
     "publishing_app",
     "rendering_app",
     "update_type",
-    "locale"
+    "locale",
+    "public_updated_at"
   ],
   "properties": {
     "base_path": {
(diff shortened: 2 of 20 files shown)


rboulton:
I agree with the logic behind this change.  Currently failing the whitehall contract tests, though.

jamiecobbett:
@bradleywright do you have an opinion on this?
> We have been thinking of public_updated_at as "the updated date which users
see" and because some formats don't display this (eg gone/redirect items), it
has been optional.

> However, it seems to me that it isn't about what the user sees, but when there
was last a (significant) semantic change to the content at the URL. Therefore,
everything should have a public_updated_at.

bradleywright:
@jamiecobbett I broadly agree with your reasoning. The logic of `public_updated_at` is mostly driven by Whitehall's change notes, happy to make it a standard.

jamiecobbett:
Merging this Whitehall change would fix that build: https://github.com/alphagov/whitehall/pull/2106

jamiecobbett:
All builds are now passing

PR #41, opened 2015-04-15
Dummy server for example content

heathd:
This is a self-contained rack app which will read example files from the
example folder.

It can be run either using the `startup.sh` shell script, or as a rack
up directly using rackup and the `config.ru` file.

We ensure that file changes propogate immediately by doing the
following:

- Set the `Cache-control: no-cache` header to ensure that the content
  store client does not cache the fetched items.
- Always read the file from disk so that changes propogate immediately.

It runs on port 3068 (the content store port) and serves the examples using the urls in the example files themselves. Example urls are required to be unique.

### DIFF
--- a/dummy_content_store/README.md
+++ b/dummy_content_store/README.md
@@ -1,0 +1,20 @@
+# Dummy content store
+
+The dummy content store allows you to run a frontend using the examples from
+the `govuk-content-schemas` repo. You will not need to run the content store
+itself.
+
+To run the dummy content store:
+
+```sh
+$ ./dummy_content_store/startup.sh
+```
+
+by default it will use the examples from the `formats` folder.
+
+To use a different folder set the `EXAMPLES_PATH` environment variable:
+
+```sh
+$ EXAMPLES_PATH=/path/to/my/examples ./dummy_content_store/startup.sh
+```
+
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source "https://rubygems.org"
 gem 'json-schema'
 gem 'rake'
 gem 'rspec'
+gem 'rack'
+gem 'rack-contrib'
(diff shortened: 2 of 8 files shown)


dummy_content_store/README.md
```diff
@@ -0,0 +1,20 @@
+# Dummy content store
```
jamiecobbett: Could you add a link from the README to this page?

heathd:
@jamiecobbett link added to README

vinayvinay:
:+1:

dsingleton:
Just had a quick play with this locally and managed to get `government-frontend` to render [/government/case-studies/get-britain-building-carlisle-park](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/case_study/frontend/examples/case_study.json), just running:

- `govuk-content-schemas$ ./dummy_content_store/startup.sh`
- `government-frontend$ PLEK_SERVICE_STATIC_URI=assets.digital.cabinet-office.gov.uk ./startup.sh`

get britain building carlisle park case study gov uk

My experience of getting this running was that it took a while to find an example from schemas that could be rendered, i think it's only case studies at the moment (i went to try `manuals`, but we've no examples yet, and `finders` have a further dependency on search). 

I also had to dig around finding the specific path(s) that are in the dummy content store - a `rake routes` like tool would be super useful.

None of that is a blocker for merging this, but worth considering in how we make this easy for folks to use day to day. 

:+1: from me.

heathd:
Great feedback @dsingleton, thanks!

For listing the examples, another option I had considered was to have the dummy app render a page with a list of them. It could even generate the URLs based on the rendering_app using plek. Need to experiment with that to see if it would work reliably.

dsingleton:
> For listing the examples, another option I had considered was to have the dummy app render a page with a list of them. It could even generate links to view the examples in the respective frontend apps in the dev environment by feeding the rendering_app to plek. Need to experiment with that to see if it would work reliably.

Yep, something along those lines would be lovely.

PR #42, opened 2015-04-16
Schema for email-alert-signup

tommyp:
This commit adds a schema for the new email-alert-signup format along with an example. It currently only contains 2 items in the details hash: `tags` and `breadcrumbs`.

### DIFF
--- a/formats/email_alert_signup/publisher/details.json
+++ b/formats/email_alert_signup/publisher/details.json
@@ -1,0 +1,33 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "tags"
+  ],
+  "properties": {
+     "tags": {
+       "type": "object"
+    },
+    "breadcrumbs": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "link",
+          "title"
+        ],
+        "properties": {
+          "link": {
+            "type": "string",
+            "format": "uri"
+          },
+          "title": {
+            "type": "string"
+          }
+        }
+      }
+    }
+  }
+}
--- a/formats/email_alert_signup/publisher/links.json
+++ b/formats/email_alert_signup/publisher/links.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {},
+  "definitions": {
+    "guid_list": {
+      "type": "array",
+      "items": {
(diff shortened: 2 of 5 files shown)


formats/email_alert_signup/publisher/details.json
```diff
@@ -0,0 +1,34 @@
...
+          "title"
+        ],
+        "properties": {
+          "link": {
+            "type": "string",
+            "format": "uri"
+          },
+          "title": {
+            "type": "string",
+            "pattern": "^[a-zA-Z0-9-]+$"
```
jamiecobbett: I think the tests are failing because this regex doesn't allow spaces. We don't have validations on titles normally, so maybe it can just be dropped?

tommyp: The tests are falling because I forgot to remove a regex that used extended pattern matchers which the validator doesn't support. Fair point about removing this though. I've amended the commit.

jamiecobbett: OK - I was referring to:
```
  The property '#/details/breadcrumbs/0/title' value "An extra bank holiday" did not match the regex '^[a-zA-Z0-9-]+$' in schema file:///home/jenkins/workspace/govuk_content_schemas_branches/formats/email_alert_signup/frontend/schema.json#
```
https://ci-new.alphagov.co.uk/job/govuk_content_schemas_branches/109/console

But it's all fixed now, so no worries.

tommyp: Oh. My mistake. Good call.

formats/email_alert_signup/publisher/links.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "related"
```
tekin: Is this really required to generated an email alert?

formats/email_alert_signup/publisher/links.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "properties": {
```
tekin: Looking at the [email-alert-frontend pull request](https://github.com/alphagov/email-alert-frontend/pull/1), it doesn't look like any of these properties are actually needed to generate an email alert. If so, is there another reason to include `related`, `organisations` and `people`? If there isn't, we shouldn't include them in the schema.

formats/email_alert_signup/publisher/links.json
tommyp: Well no, but I thought requiring it would be good so we know which page you're signing up to?

formats/email_alert_signup/publisher/links.json
tommyp: I thought that the idea was that we set to `additionalProperties` to `false` and specify each possible key to avoid typos?

tekin:
:+1:

PR #43, opened 2015-04-17
Add format for "placeholder" items

jamiecobbett:
We have around ~~149,161~~ 153,236 placeholder items, but we don't have a schema for the
format. I've developed this against the placeholders produced by (mainstream)
publisher and whitehall. Some small changes will be required for publisher to
supply additional fields like "locale", but Whitehall's placeholders are already valid against this schema.

Once this is merged, I'll open PRs to test Whitehall's placeholders and to add schema tests to Publisher.

### DIFF
--- a/formats/placeholder/frontend/schema.json
+++ b/formats/placeholder/frontend/schema.json
@@ -1,0 +1,147 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "format",
+    "locale"
+  ],
+  "properties": {
+    "base_path": {
+      "type": "string"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "format": {
+      "type": "string",
+      "enum": [
+        "placeholder"
+      ]
+    },
+    "public_updated_at": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "locale": {
+      "type": "string"
+    },
+    "need_ids": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "details": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "change_note": {
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "tags": {
+          "type": "object",
+          "properties": {
+            "browse_pages": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "primary_topic": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "additional_topics": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "topics": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          }
+        }
+      }
+    },
+    "links": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "available_translations": {
+          "$ref": "#/definitions/frontend_links"
+        }
+      }
+    },
+    "updated_at": {
+      "type": "string",
+      "format": "date-time"
+    }
+  },
+  "definitions": {
+    "route": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "path",
+        "type"
+      ],
+      "properties": {
+        "path": {
+          "type": "string"
+        },
+        "type": {
+          "enum": [
+            "prefix",
+            "exact"
+          ]
+        }
+      }
+    },
+    "guid_list": {
+      "type": "array",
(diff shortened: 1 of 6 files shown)


formats/placeholder/publisher/schema.json
```diff
@@ -0,0 +1,145 @@
...
+      }
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+      }
+    }
+  }
+}
```
rboulton: Shouldn't this have a newline at the end of the file?

jamiecobbett: These files are generated by the `make` task. It should be possible to change the code which generates them to add a trailing newline, but that's outside of the scope.

rboulton: Ah, no problem, then.

formats/placeholder/frontend/schema.json
```diff
@@ -0,0 +1,147 @@
...
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "format": {
+      "type": "string",
+      "enum": [
+        "placeholder"
```
tekin: This will mean that any placeholder items with the prefix format (e.g. "placeholder_organisation") will not pass validation. Can this be made a pattern match instead?

formats/placeholder/publisher/details.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
```
alext: Does this mean that placeholder items are restricted in what can be in the details hash?  I'm not sure that's necessarily what we want.

A placeholder item could readily have the full details that the eventual non-placeholder format will have, and only be using the placeholder variant to prevent routes being registered during the transition phase.

formats/placeholder/frontend/schema.json
jamiecobbett: Good idea.

jamiecobbett: Unfortunately, that would require a code change as the `format` enum is [automatically added based upon the format](https://github.com/alphagov/govuk-content-schemas/blob/master/lib/govuk_content_schemas/schema_combiner.rb#L23). Given that you're creating a precedent for variable values for `format`, would you be happy to try it in a separate PR?

tekin: This isn't a new precedence, there is already be a bunch of content items with the "placeholder_FORMAT"-style of format. For example, see https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api_presenters/placeholder.rb

formats/placeholder/publisher/details.json
jamiecobbett: Generally, we disallow undefined keys for good usability reasons and it would be a shame to lose that. For Whitehall case studies, we are doing something very similar by sending the item with a format of "case_study", but a rendering_app of "whitehall-frontend", rather than "government-frontend". Maybe we can defer what you suggest until we need it?

formats/placeholder/frontend/schema.json
jamiecobbett: Ah, you're right, sorry. I need to add another schema validation to my Whitehall branch.

jamiecobbett: OK, finally got it working in 400b4c5.

jamiecobbett:
Any more for any more?

formats/placeholder/frontend/schema.json
```diff
@@ -0,0 +1,146 @@
...
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "format": {
+      "type": "string",
+      "pattern": "^(placeholder|placeholder_.+)$",
+      "description": "Should be of the form 'placeholder_my_format_name'. 'placeholder' is allowed for backwards compatibility."
```
tekin: I may be wrong, but 'placeholder' isn't just there for backwards compatibility AFAIK. It's just as valid as 'placeholder_FORMAT'. The main difference is that the latter allows the content-register to identify the content and add it to the register. I can see situations where we may want to push placeholder content out without this happening. @alext?

formats/placeholder/frontend/schema.json
```diff
@@ -0,0 +1,146 @@
...
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "format": {
+      "type": "string",
+      "pattern": "^(placeholder|placeholder_.+)$",
```
tekin: :+1:

tekin:
Minor comment on the description for the format, but otherwise looks good to me :+1:.

PR #51, opened 2015-04-24
Drop email alert signup enabled from Finders

tommyp:
As all Finders that want it now have email alert signup being live, we don't care about keeping this feature hidden. This commit removes it from the schema and the example that had it.

### DIFF
--- a/formats/finder/frontend/examples/cma-cases.json
+++ b/formats/finder/frontend/examples/cma-cases.json
@@ -11,7 +11,6 @@
     "beta": false,
     "beta_message": null,
     "document_noun": "case",
-    "email_signup_enabled": true,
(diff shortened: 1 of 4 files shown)


dsingleton:
It [looks like](https://ci-new.alphagov.co.uk/job/govuk_specialist_publisher_schema_tests/74/console) `email_signup_enabled` is still being set (but to false).

I think it's this line causing it: https://github.com/alphagov/specialist-publisher/blob/master/app/presenters/finder_content_item_presenter.rb#L45

tommyp:
@dsingleton That's cause https://github.com/alphagov/specialist-publisher/pull/475 needs merged first.

PR #52, opened 2015-04-24
Add new facets for policy finder example

dsingleton:
Policy finders should show result metadata including
the date updated, the organisations it's from, and
the document type.

This matches the display on existing policy filter
views, that will be replaced by the policy finder
eg, https://www.gov.uk/government/policies/fulfilling-the-commitments-of-the-armed-forces-covenant/activity

This adds a new facet option (result_metadata_name_key)
that will be documented in the finder-frontend repo

### DIFF
--- a/formats/policy/frontend/schema.json
+++ b/formats/policy/frontend/schema.json
@@ -99,6 +99,9 @@
                     }
                   }
                 }
+              },
+              "result_metadata_name_key": {
+                "type": "string"
               }
             }
           }
--- a/formats/policy/frontend/examples/policy_area.json
+++ b/formats/policy/frontend/examples/policy_area.json
@@ -29,6 +29,28 @@
           "key": "government_name",
           "display_as_result_metadata": true,
           "filterable": false
+        },
+        {
+          "key": "public_timestamp",
+          "short_name": "Updated",
+          "type": "date",
+          "display_as_result_metadata": true,
+          "filterable": false
+        },
+        {
+          "key": "organisations",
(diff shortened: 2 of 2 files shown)


dsingleton:
cc @tommyp and @evilstreak for finders expertise. 

(the doc for the new field is added in https://github.com/alphagov/finder-frontend/pull/191)

formats/policy/frontend/schema.json
```diff
@@ -99,6 +99,9 @@
                     }
                   }
                 }
+              },
+              "result_metadata_name_key": {
```
tommyp: This should be in the publisher schema too.

dsingleton: Yep, i'm putting that PR together now.

I'd assumed the process was get the schema agreed first - then other PRs (though i cheated with the FF one, as it's got more than just a schema update.

tommyp: I'm happy enough for the changes to go through in one PR. That's the way I've always done it.

dsingleton: I totally misunderstand, thought you mean the schema the policy publisher publishes.

Makes sense now, i'll add that.

evilstreak: This is probably better off grouped with the `display_as_result_metadata` key a few properties up.

dsingleton:
Updated PR here: https://github.com/alphagov/govuk-content-schemas/pull/53

PR #53, opened 2015-04-27
Add 'result_metadata_name_key' finder facet option

dsingleton:
For rummager fields which are objects/hashes, this option
specifies which key contains the value to use for display.

Existing spec-pub finders (cma, etc), use the 'label' key
by convention, which was hard coded into finder-frontend,
but some fields used in policy finders (like organisation)
use a different key for the display value.

This option allows policy finders to fix that through the
schema, and updates example policy content to use it and
adds two other options that will be used in policy finders.

Policies are a based on the finders schema, so the change
is duplicated in both.

### DIFF
--- a/formats/policy/frontend/examples/policy_area.json
+++ b/formats/policy/frontend/examples/policy_area.json
@@ -29,6 +29,28 @@
           "key": "government_name",
           "display_as_result_metadata": true,
           "filterable": false
+        },
+        {
+          "key": "public_timestamp",
+          "short_name": "Updated",
+          "type": "date",
+          "display_as_result_metadata": true,
+          "filterable": false
+        },
+        {
+          "key": "organisations",
+          "short_name": "From",
+          "type": "text",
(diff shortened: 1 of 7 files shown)


formats/policy/frontend/examples/policy_area.json
```diff
@@ -29,6 +29,28 @@
           "key": "government_name",
           "display_as_result_metadata": true,
           "filterable": false
+        },
```
rboulton: Did you mean to include these changes in this PR?  It doesn't look related to the rest of the PR.

dsingleton: I did, but they should have been a separate commit for clarity.

I'm about to update the PR so these are the _only_ changes. Hold on to you :tophat:

dsingleton:
Closing this and taking a different approach for `result_metadata_name_key`, that doesn't use schemas.

Will open a new PR that just adds the updates content examples for the new metadata fields.

PR #57, opened 2015-04-29
Extract dummy content store to a gem

heathd:
We wanted to add new features to `dummy_content_store`, and it felt like a good time to pull it out into a [separate gem](https://github.com/alphagov/govuk-dummy_content_store), so that we avoid overloading the `govuk-content-schemas` repo.

This commit removes the dummy content store code and updates the readme to point to the new repo.

Screenshot of the examples index from the new gem:

screen shot 2015-04-29 at 15 54 54


jamiecobbett:
I just noticed that this doesn't remove the entries in Gemfile that are no longer needed, sorry.

PR #62, opened 2015-05-06
Add schema for manual_sections

jamiecobbett:
A separate schema will be required for HMRC manual sections.

### DIFF
--- a/formats/manual_section/publisher/details.json
+++ b/formats/manual_section/publisher/details.json
@@ -1,0 +1,50 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "body",
+    "manual",
+    "organisations"
+  ],
+  "properties": {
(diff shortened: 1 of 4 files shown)


formats/manual_section/publisher/details.json
```diff
@@ -0,0 +1,50 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "body",
+    "manual",
+    "organisations"
+  ],
+  "properties": {
```
danielroseman: Looks like the exporter is currently sending empty arrays for `breadcrumbs` and `child_section_groups` for the manual_sections format, which would be invalid according to this schema.

jamiecobbett: Yep, I've removed them in the PR: https://github.com/alphagov/specialist-publisher/commit/17ef9957a7bb5cd0e60785bd83e86acd1c23ae7a

PR #63, opened 2015-05-06
Add 'result_metadata_name_key' to finder schemas

dsingleton:
This is used on policy finders and the finder of policies.

It's causing https://github.com/alphagov/policy-publisher/pull/60 to fail at the moment.

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -121,6 +121,9 @@
               "type": {
                 "type": "string"
               },
+              "result_metadata_name_key": {
+                "type": "string"
+              },
               "allowed_values": {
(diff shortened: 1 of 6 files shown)


tommyp:
Can you update https://github.com/alphagov/finder-frontend/blob/master/docs/finder-content-item.md to explain what this new key does?

dsingleton:
Yeah, will do, could have sworn i had last time.

(longer term it might make more sense for that doc to live in the same repo as schemas, and be linked to from finder-frontend, fewer PRs for simple changes)

tommyp:
True. Especially since the schemas may be self documenting but can still be a bit confusing as to what the Apps actually do with the Content Items.

PR #66, opened 2015-05-14
Add format type to email alert signup

tommyp:
As we want to make Email Alert Signup work even when there are duplicate
topic names (thanks Govdelivery) we want to prepend a `format_type`
string to the topic name to prevent duplications across different format
on GOV.UK. This commit adds it to the schema and examples.


tommyp:
Open to suggestions if people think `format_type` is too close to `format`?

tommyp:
After a chat with several people I'm going to close this and take a different approach.

PR #68, opened 2015-05-14
Add default order to Finders

tommyp:
For the Policies Finder, ordering by last update doesn't make sense as
the content being updated within each policy doesn't bubble up to update
the policy itself. This commit adds a `default_order` to the Finders and
Policy schemas along with adding an example to `/finders` of the Policy
Finder which uses this.

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -73,6 +73,9 @@
         "logo_path": {
           "type": "string"
         },
+        "default_order": {
+          "type": "string"
+        },
         "filter": {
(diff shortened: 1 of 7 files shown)


dsingleton:
I'm not sure `default_order` is the best term. It's not really a default, in that thats the only value it can be, it's more like `order`, but thats a really vague term. Possibly `sort_by`, `order_by` or similar?

rboulton:
On the other hand... I think it's quite likely we'll want to allow users to control the sort order at some point in future, in which case `default_order` becomes a good name for this.

PR #69, opened 2015-05-18
Allow description to be included in linked item expansion

alext:
Necessary to allow rendering mainstream browse pages (eg https://www.gov.uk/browse/benefits) from the content-store.

This also includes an update to make the `title`, `base_path` and `locale` fields mandatory in the expansion.

Once this is merged, I'll raise a follow-up PR on content-store to include this field in the expansion.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -227,6 +227,9 @@
           "title": {
             "type": "string"
           },
+          "description": {
+            "type": "string"
+          },
           "base_path": {
             "type": "string"
           },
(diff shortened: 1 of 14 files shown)


jamiecobbett:
Do all the existing items in content-store satisfy this spec? We have a rake task to check them against the frontend schema in content-store: https://github.com/alphagov/content-store/blob/master/lib/tasks/check_content_items_against_schema.rake

alext:
This only affects the link expansion done by the content-store itself, and is therefore defined more by the code than the data.  All renderable items have to have a title, and a locale (it default to 'en'), and the link expansion only expands renderable items (having said that, I notice that at present it's excluding redirects, but not other non-renderable formats like gone items.  I'll raise a PR to fix that).

The added description attribute is optional, and therefore won't be invalid if not present.

rboulton:
I think the description should maybe be required (but allowed to be blank, of course), but Alex and I had an offline conversation about this; even if we want it to be required, it will be much easier to migrate to having it by merging this, making the content store always populate it, and then making it required.

PR #70, opened 2015-05-18
Add feed_enabled boolean for Finders

tommyp:
For some Finder that don't have results order by when they were last
updated, a feed doesn't really make sense. This commit adds a
feed_enabled boolean to the Finder schema to allow this to be set.


tommyp:
Closing to rename this to `feed_disabled`.

PR #71, opened 2015-05-18
Add feed_disabled boolean for Finders

tommyp:
For some Finder that don't have results order by when they were last
updated, a feed doesn't really make sense. This commit adds a
feed_disabled boolean to the Finder schema to allow this to be set.


tommyp:
Closing this as we won't be defining this in the schemas.

PR #76, opened 2015-05-27
Add curated links to mainstream browse pages

tijmenb:
Some mainstream browse pages will be getting curated lists of links, instead of a flat list.

This commit adds an optional `groups` array to the `mainstream_browse_page` schema. Specification for this is copied from the `topic` schema.

FAO @rboulton 

Trello: https://trello.com/c/dsgpdVKG

### DIFF
--- a/formats/mainstream_browse_page/frontend/examples/curated_level_2_page.json
+++ b/formats/mainstream_browse_page/frontend/examples/curated_level_2_page.json
@@ -1,0 +1,31 @@
+{
+    "base_path": "/browse/benefits/entitlement-with-list",
+    "description": "When and how benefit payments are made, benefits adviser and benefit fraud",
+    "details": {},
+    "format": "mainstream_browse_page",
+    "details": {
+      "groups": [
+        {
+          "name": "Fraud",
+          "contents": [
+            "https://www.gov.uk/benefit-fraud",
+            "https://www.gov.uk/national-benefit-fraud-hotline",
+            "https://www.gov.uk/benefit-integrity-centres"
+          ]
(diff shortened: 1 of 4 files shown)


formats/mainstream_browse_page/frontend/examples/curated_level_2_page.json
```diff
@@ -0,0 +1,29 @@
+{
+    "base_path": "/browse/benefits/entitlement-with-list",
+    "description": "When and how benefit payments are made, benefits adviser and benefit fraud",
+    "details": {},
+    "format": "mainstream_browse_page",
+    "details": {
+      "groups": [
+        {
+          "name": "Oil rigs",
```
rboulton: It would be better to pick some examples of content which actually makes sense for "benefits entitlement", so that people don't get distracted when reviewing this by the odd content.  Also, so that merging the results from this API with the results from the content-api works.

rboulton:
This looks good, except having more plausible example data would make the examples clearer.

I believe the test failures are unrelated, but that needs checking (and the appropriate people need to be told/reminded they've broken the build).

rboulton:
Merged, having verified that the test failures are unrelated.

heathd:
Richard and I figured out that the failures were due to the fact that the branch hadn't been rebased, so when the finder tests were running against this particular commit, they were missing some schema changes needed by the latest finder code. 

After this branch was merged into master, the tests passed.

Something to watch out for in future (if there's a test failure on a branch build of govuk-content-schemas which you can't explain, make sure your branch is rebased against master).

PR #77, opened 2015-05-27
Add `beta` attribute to topic 

tijmenb:
Topics can be in beta now, so we need a boolean. It can also be null/empty, since older topics don't have the `beta` label set. Used the `finder` schema as an example ([source](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/finder/publisher/details.json#L11)).

Trello: https://trello.com/c/H9UMp051/139-allow-topic-subtopics-to-be-labelled-as-beta


rboulton:
Adding an example of a topic which has "beta" set would be good.

PR #81, opened 2015-06-16
Correct comments about browse page links fields

rboulton:
We've changed the semantics of these a bit, so the descriptions needed
to be updated.  The main change is that we're always going to send
active_top_level_browse_page, except for on the root browse page.

### DIFF
--- a/formats/mainstream_browse_page/publisher/links.json
+++ b/formats/mainstream_browse_page/publisher/links.json
@@ -8,11 +8,11 @@
       "$ref": "#/definitions/guid_list"
     },
     "active_top_level_browse_page": {
-      "description": "The top-level browse page (for second-level browse pages only).",
+      "description": "The top-level browse page which is active",
       "$ref": "#/definitions/guid_list"
     },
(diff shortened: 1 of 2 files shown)


alext:
You need to run make, and commit the changes to the generated schema.json files as well.

formats/mainstream_browse_page/publisher/links.json
```diff
@@ -8,11 +8,11 @@
       "$ref": "#/definitions/guid_list"
     },
     "active_top_level_browse_page": {
-      "description": "The top-level browse page (for second-level browse pages only).",
+      "description": "The top-level browse page which is active",
       "$ref": "#/definitions/guid_list"
     },
     "second_level_browse_pages": {
-      "description": "All 2nd level browse pages for the top_level_browse_page (or self if this is a top-level browse page)",
+      "description": "All 2nd level browse pages under active_top_level_browse_page",
```
alext: Nitpick: "under **the** active_top_level_browse_page"

PR #82, opened 2015-06-16
Add filter facets for type and org to policy example

dsingleton:
These are being included by policy-publisher when it published
policy content items.

Related: https://github.com/alphagov/policy-publisher/pull/82


dsingleton:
Closing this because it's breaking tests in finder-frontend, in a very annoying way.

alext:
If you rebase this against master, that should fix the collections-publisher test failure.

PR #85, opened 2015-06-22
Add support for access limiting

danielroseman:
Adds an `access_limited` object to the metadata, which appears as the top level in all generated schemas. This currently consists of a single value, `users`, which is a list of user IDs.

(Note that user IDs generated from signon are inconsistent; some are GUIDs and some are what looks like hashed strings, so there is no attempt to enforce a format here.)

Also adds an example of access limiting in a case study.

### DIFF
--- a/formats/metadata.json
+++ b/formats/metadata.json
@@ -60,6 +60,19 @@
       "items": {
         "$ref": "#/definitions/route"
       }
+    },
+    "access_limited": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [ "users" ],
+      "properties": {
+        "users": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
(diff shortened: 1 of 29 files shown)


bradleywright:
Seems reasonable to me.

formats/metadata.json
```diff
@@ -60,6 +60,19 @@
       "items": {
         "$ref": "#/definitions/route"
       }
+    },
+    "access_limited": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [ "users" ],
+      "properties": {
+        "users": {
```
jamiecobbett: I would have preferred to have it named "user_uids" or similar so that is explicit what it contains and to make it cleaner if we needed to migrate to passing around more data about each user.

PR #88, opened 2015-06-23
Add children links to topic.

alext:
This will allow us to render topic pages entirely from the
content-store.


tijmenb:
Would it be useful to add `children` to one of the examples too?

alext:
Good call.  I'll add that.

alext:
Added...

PR #91, opened 2015-06-29
Add documents_per_page to Finders

tommyp:
As we are introducing Pagination to Finders we need to be able to
specify the amount per page of the results. This commit adds it to both
the Finder and Policy formats and adds it to the Policy Area example.

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -70,6 +70,9 @@
           "type": "string",
           "additionalProperties": false
         },
+        "default_documents_per_page": {
+          "type": "integer"
+        },
         "logo_path": {
           "type": "string"
         },
--- a/dist/formats/finder/publisher/schema.json
+++ b/dist/formats/finder/publisher/schema.json
@@ -116,6 +116,9 @@
           "type": "string",
           "additionalProperties": false
         },
+        "default_documents_per_page": {
+          "type": "integer"
+        },
         "logo_path": {
           "type": "string"
         },
--- a/dist/formats/policy/frontend/schema.json
+++ b/dist/formats/policy/frontend/schema.json
@@ -49,6 +49,9 @@
         "document_noun": {
           "type": "string"
         },
+        "default_documents_per_page": {
+          "type": "integer"
+        },
         "email_signup_enabled": {
           "type": "boolean"
         },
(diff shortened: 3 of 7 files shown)


alext:
This seems like a presentation concern.  Does it really belong in the content-store?

tommyp:
This is primarily for Policy pages but I can see this needing to be configurable for each Finder since they all have different users with different stakeholders.

bradleywright:
I agree with @alext - if there's not an existing need for configurable pagination I'd rather the frontends just chose a sensible default (eg 10 for policies).

edds:
We know that many different finders are going to want different pagination settings. Just in policy we know that the main policy finder doesn't want any pagination and the individual policies are going to want pagination. We have tried, where possible to keep finder-frontend a completely agnostic and configurable. To implement this in finder-frontend we would need to start parsing base paths and then matching them against rules to work out how they should be paginated.

The content store is also designed to be the single place where frontends can ask for configuration, it makes sense that presentational concerns are going to end up in there.

tommyp:
An additional point to note is that per-Finder settings being held within the code base was the previous state of Finder Frontend until post firebreak. Moving all of these sorts of settings to within the Content Item keeps us away from parsing base paths, etc as @edds mentioned.

rboulton:
I think this _is_ a presentational concern to some extent, but I think that's fine to be in the content store; there's already lots of presentation information there (eg, ordering of facets), and it's certainly better there than having special cases in finder-frontend.

I have two concerns:

 - might we want to allow the number of results per page to be set by the user in future?  This seems plausible, so perhaps this should be called `default_documents_per_page` instead (to go with the `default_order` setting we already have). This would also be less confusing if we implement slightly fuzzier pagination strategies in future (such as avoiding almost empty pages by showing 11 results on the page if there are only 11 results left, even when the pagination size is 10).

 - Should this be a required field?  It would be good to avoid having a fallback in finder-frontend, so that there is only one place a developer needs to go to find out what the pagination setting is.

tommyp:
@rboulton:

1. Fair point about a User deciding on pagination. Naming it as `default_documents_per_page` here makes sense.

2. We could easily make it a required field, but that then means we need to deploy the Finder Frontend changes with a default, republishing all the Finders and then removing the default. At a certain point republishing all the Finders is going to get too cumbersome. There's already ~230 Finders. At what point do we stop republishing them all?

rboulton:
Making it required would be best done by a separate later pull request anyway, so isn't a blocking consideration for this PR.  I'll be happy to merge this once the naming is changed.

tommyp:
Updated this to now be `default_documents_per_page`.

PR #92, opened 2015-07-09
Add child ordering to mainstream browse page

jackscotti:
Currently, all 2nd level mainstream browse columns have the label "A to Z" on them. When a page has been curated, we shouldn't display this label, because it won't be true.
Therefore, collections,  needs to know the child ordering type for a top level mainstream browse pages in order to display related labels.

https://trello.com/c/bcTrheDZ/244-don-t-show-a-to-z-on-mainstream-browse-columns-which-aren-t-in-alphabetical-order

### DIFF
--- a/formats/mainstream_browse_page/frontend/examples/top_level_page.json
+++ b/formats/mainstream_browse_page/frontend/examples/top_level_page.json
@@ -1,7 +1,9 @@
 {
     "base_path": "/browse/benefits",
     "description": "Includes tax credits, eligibility and appeals",
-    "details": {},
+    "details": {
+      "2nd_level_ordering": "alphabetical"
+    },
     "format": "mainstream_browse_page",
     "links": {
       "top_level_browse_pages": [
--- a/formats/mainstream_browse_page/publisher/details.json
+++ b/formats/mainstream_browse_page/publisher/details.json
@@ -3,6 +3,9 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
+    "2nd_level_ordering": {
+      "enum": [ "alphabetical", "curated" ]
+    },
     "groups": {
       "type": "array",
       "items": {
--- a/dist/formats/mainstream_browse_page/frontend/schema.json
+++ b/dist/formats/mainstream_browse_page/frontend/schema.json
@@ -41,6 +41,12 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "2nd_level_ordering": {
+          "enum": [
+            "alphabetical",
+            "curated"
+          ]
+        },
         "groups": {
           "type": "array",
           "items": {
--- a/dist/formats/mainstream_browse_page/publisher/schema.json
+++ b/dist/formats/mainstream_browse_page/publisher/schema.json
@@ -87,6 +87,12 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "2nd_level_ordering": {
+          "enum": [
+            "alphabetical",
+            "curated"
+          ]
+        },
         "groups": {
           "type": "array",
           "items": {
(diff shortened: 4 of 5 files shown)


formats/mainstream_browse_page/publisher/details.json
```diff
@@ -3,6 +3,9 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
+    "child_ordering": {
+      "type": "string"
```
alext: Instead of a string, this would be better as an `enum`.  See https://github.com/alphagov/govuk-content-schemas/blob/master/formats/redirect/publisher/schema.json#L21-L23 for example.

formats/mainstream_browse_page/frontend/examples/top_level_page.json
```diff
@@ -1,7 +1,9 @@
 {
     "base_path": "/browse/benefits",
     "description": "Includes tax credits, eligibility and appeals",
-    "details": {},
+    "details": {
+        "child_ordering": "curated"
+      },
```
alext: It would be better to add another example with curated ordering, and leave this one for the common case.

formats/mainstream_browse_page/publisher/details.json
```diff
@@ -3,6 +3,9 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
+    "child_ordering": {
```
alext: I think this would be better named `2nd_level_ordering` as it will need to be present on both the top-level and 2nd-level browse pages.

jackscotti: At the moment, both 1st and 2nd level pages can access the `child_ordering` parameter in `collections-publisher`. 
`child_ordering` refers to the children of the page itself, I therefore do not think that `2nd_level_ordering` is the best name because we would be ending up with 2nd level pages with a `2nd_level_ordering` param that refers to an eventual 3rd level layer.

formats/mainstream_browse_page/frontend/examples/top_level_page.json
jackscotti: Makes sense. Will do.

formats/mainstream_browse_page/publisher/details.json
jackscotti: We have already a validation in place within `collections-publisher` and I thought that was enough. But doing what you suggested means future-proofing in case of some other apps will publish to `content-store`. Thanks, will do.

alext: This schema forms the contract for what data is passed between the apps, and therefore it's important that this is as specific as needed.  Any validation in the apps is secondary.

formats/mainstream_browse_page/publisher/details.json
alext: This field will need to be present in the content-items for both top-level and 2nd-level pages (so that we can hide the label for both pages).  For a 2nd-level page, if the field was called `child_ordering` it would suggest that it specifies the ordering for the children of the current page (ie the 3rd-level pages), when in fact it specifies the ordering of the siblings of the current page.  That's why I think `2nd_level_ordering` is a better name for ti in this context.

jackscotti: Got it. :+1:  Will do.

jackscotti:
@alext, you merged it already but I was wondering, wouldn't this need to be called `second_level_ordering`? How am I going to refer to this parameter if Ruby does not allow variables starting with a number?

rboulton:
I'm pretty sure the ruby side of things can be worked around - but I do think it would be good to rename this to "second_level_ordering", for consistency with the existing "second_level_browse_pages" link field name.

alext:
Sounds good on the naming front.

As for strings vs symbols, I prefer to use string keys for data that's going to be JSON encoded to make things clearer.  Symbols will be converted to strings in the round-trip to JSON and back, and therefore the source hash doesn't match the JSON decoded version of it.

PR #95, opened 2015-07-16
Specialist document schema

heathd:
A first pass at specialist document schemas ([trello ticket](https://trello.com/c/x7TPNRfG/262-initial-version-of-specialist-document-schemas))

I did the following:

1. dumped out examples from the `rendered_specialist_document` mongo collection
2. made a first pass at a schema by converting the [finder schemas](https://github.com/alphagov/specialist-publisher/tree/master/finders/schemas) from specialist-publisher into json-schema format using a script
3. massaged the examples and schemas be semantically clear and consistent

Some detail on steps 2 and 3:

## 2. Converting finder schemas to json-schema

The [finder schemas](https://github.com/alphagov/specialist-publisher/tree/master/finders/schemas) already describe all of the metadata fields of each specialist document type. It seemed error prone to convert these manually, so I wrote a script to do this.

A typical usage might be:
```
$ ./bin/convert_finder_schemas ../specialist-publisher/finders/schemas/*.json > formats/specialist_document/publisher/facets.json
```

this will write a set of `definitions` into the facets.json file.

```json
{
  "definitions": {
    "aaib_report_metadata": {
      "type": "object",
      "properties": {
        "document_type": {
          "type": "string",
          "enum": [
            "aaib_report"
          ]
        }

        ...

      }
    }
  }
}
```

these can then be incorporated into the `definitions` section of the `details.json` file.

they can then be used to define the metadata element using a [`oneOf` rule](http://json-schema.org/latest/json-schema-validation.html#anchor88):

```json
  "definitions": {
    "any_metadata": {
      "type": "object",
      "oneOf": [
        { "$ref": "#/definitions/aaib_report_metadata" },
        { "$ref": "#/definitions/cma_case_metadata" },
        { "$ref": "#/definitions/countryside_stewardship_grant_metadata" },
        { "$ref": "#/definitions/drug_safety_update_metadata" },
        { "$ref": "#/definitions/european_structural_investment_fund_metadata" },
        { "$ref": "#/definitions/international_development_fund_metadata" },
        { "$ref": "#/definitions/maib_report_metadata" },
        { "$ref": "#/definitions/medical_safety_alert_metadata" },
        { "$ref": "#/definitions/raib_report_metadata" },
        { "$ref": "#/definitions/vehicle_recalls_and_faults_alert_metadata" }
      ]
    }
  }
```
This also has the advantage that it should be fairly easy to update
these definitions if the finder schemas change.

The converter has two sets of hard-coded information which are not possible to determine automatically:

1. the mapping from the filename of the finder schema to the `document_type`
2. identifying whether a field allows only a single value (ie. single-select) or a list of values (ie. multi-select)

### Metadata labels

The `rendered_specialist_document` included both a machine-readable and human-readable version of each metadata field, e.g.

```
"report_type" : "investigation-report",
"report_type_label" : "Investigation report",
```

however, the `_label` values are not actually used by the specialist-frontend. Instead, when rendering a specialist-document it fetches both the [finder object and the specialist document object](https://github.com/alphagov/specialist-frontend/blob/master/app/controllers/specialist_documents_controller.rb#L10) from the content store and uses the finder object to determine the human readable labels.

Therefore I've removed all of the `_label` fields and do not allow them in a specialist document content item.

### Metadata sub-element

I've moved all 'metadata' fields from the top-level of the `details` hash to their own `metadata` sub-element. This feels semantically cleaner and also makes it easier to validate them against the automatically converted definitions.

### Interaction with rummager

I discussed with @rboulton whether there were any considerations to ensure we could index the content-item. Richard confirmed that there will be presenter objects which convert the content item into a data structure appropriate for rummager, and therefore there was no need to retain any form of backward compatibility with the pre-existing data structure.

### Removed fields

the following metadata fields do not appear to be needed/used by specialist-frontend and were removed:

- version_number
- created_at
- updated_at
- minor_update
- bulk_published

### Description vs summary

The rendered_specialist_document had a `summary` field. In the examples I have used the `description` field instead, assuming that they serve the same purpose semantically.

### published_at

specialist-frontend appears to rely upon the `published_at` field, but this seems to have the same content/semantics as `public_updated_at`. I'm retaining `public_updated_at` because this is a standard top-level field. 

### Optional vs blank fields

Some fields, eg. `closed_date` contained empty string `""` in the `rendered_specialist_document`. I opted to disallow empty string for these fields. The field itself should be absent if the value is not set, rather than representing it as an empty string.

## Other considerations

I have not made any of the metadata fields required. This seems ok for now, but we may want to tighten things up by making the appropriate metadata fields required in the future.


bin/convert_finder_schemas
```diff
@@ -0,0 +1,93 @@
...
+  end
+end
+
+converter = GovukContentSchemas::FinderSchemaConverter.new(
+  document_type_mapper: DocumentTypeMapper.new,
+  select_field_multiplicity_identifier: SelectFieldMultiplicityIdentifier.new
+)
+converted = converter.call(*ARGV)
+
+puts JSON.pretty_generate(converted)
```
jamiecobbett: Nitpick: trailing newline

spec/lib/finder_schema_converter_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      Dir[File.dirname(__FILE__) + "/../fixtures/finder_schemas/*.json"]
+    }
+
+    it "combines all of the facets from each file" do
+      expect(converted['definitions'].keys).to eq(['aaib-reports_metadata', 'drug-safety-updates_metadata'])
+      expect(converted['definitions']['aaib-reports_metadata']['properties'].keys).to eq(["document_type", "aircraft_category", "report_type", "date_of_occurrence", "aircraft_type", "location", "registration"])
+      expect(converted['definitions']['drug-safety-updates_metadata']['properties'].keys).to eq(["document_type", "therapeutic_area", "first_published_at"])
+    end
+  end
+end
```
jamiecobbett: Nitpick: trailing newline

formats/specialist_document/frontend/examples/aaib-reports.json
```diff
@@ -0,0 +1,51 @@
...
+			{
+				"text" : "Summary:",
+				"level" : 2,
+				"id" : "summary",
+				"headers" : [
+					{
+						"text" : "Download report:",
+						"level" : 3,
+						"id" : "download-report",
+						"headers" : [ ]
```
danielroseman: I think the same should apply here as for empty strings: the field should be left out completely if it is empty.

formats/specialist_document/publisher/links.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  // "required": [
```
danielroseman: These should be removed rather than commented out.

formats/specialist_document/frontend/examples/aaib-reports.json
heathd: hmm, ok we could do that but it might mean some more bit of fiddling with the structured header extractor code to make it produce the right output https://github.com/alphagov/specialist-publisher/blob/52193f0b23986e7bc83c5ec2973e37d65f6b27f8/lib/specialist_document_header_extractor.rb#L15

I wonder if it's worth it?

any opinions @dsingleton @rboulton ?

rboulton: I have very few opinions in this area.  I'm mildly in favour of there being only one representation for an empty list, so consider me "+0.1" for @danielroseman's point.

formats/specialist_document/publisher/links.json
heathd: done

spec/lib/finder_schema_converter_spec.rb
heathd: done

bin/convert_finder_schemas
heathd: done

formats/specialist_document/frontend/examples/aaib-reports.json
bradleywright: Given `[]` is truthy in both Ruby and JavaScript I think it'd be preferable for this to be absent if not set.

heathd: ok done

PR #96, opened 2015-07-17
Base paths need a leading slash.

danielroseman:
In the specialist document examples, all the base_path values need to start with a slash for consistency and compatibility with the dummy-content-store.

### DIFF
--- a/formats/specialist_document/frontend/examples/aaib-reports.json
+++ b/formats/specialist_document/frontend/examples/aaib-reports.json
@@ -42,7 +42,7 @@
 			}
 		]
 	},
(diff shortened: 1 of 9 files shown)


jamiecobbett:
How come this didn't cause a build failure of govuk-content-schemas?

alext:
> How come this didn't cause a build failure of govuk-content-schemas

Currently the schema only mandates that the `base_path` is a string. It probably needs to be tightened up to only accept absolute paths.

heathd:
I think this change is ok as it stands, I'll make a separate PR to tighten up the base_path specification

PR #97, opened 2015-07-17
Strict base path specification

heathd:
`base_path` should always contain an absolute path. At present the schema accepts any arbitrary string.

This definition was taken from this [gist](https://gist.github.com/dpk/4757681) which is based on [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.3).

I also use the same `absolute_path` definition to validate:

- the `base_path` in the frontend representation
- the `base_path` which occurs in the links section of the frontend representation

## Other possible follow ups

The `manual`, `manual_section`, `hmrc_manual` and `hmrc_manual_section` formats all contain "base_path" fields. We may want to use this to specify those as well.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -10,7 +10,7 @@
   ],
   "properties": {
     "base_path": {
-      "type": "string"
+      "$ref": "#/definitions/absolute_path"
     },
     "title": {
       "type": "string"
@@ -200,6 +200,10 @@
     }
   },
   "definitions": {
+    "absolute_path": {
+      "type": "string",
+      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?$"
+    },
     "route": {
       "type": "object",
       "additionalProperties": false,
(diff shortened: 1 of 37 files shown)


heathd:
cc @jamiecobbett @alext

jamiecobbett:
:+1: 

> The manual, manual_section, hmrc_manual and hmrc_manual_section formats all contain "base_path" fields. We may want to use this to specify those as well.

Yes, I think that's worth doing.

alext:
Looks good, although that regex is pretty inscrutable.  Do we need it to cover so many cases.  I think we could make it more restrictive than RFC3986 (possibly just alphanumeric and `-`).  What do you think?

heathd:
@alext I think its better to match exactly what is allowed by the official spec. I think given that the regexp is defined only once in a named definition, it's clear what it's purpose is.

heathd:
@alext I'm thinking this because one day we might use the schemas to do runtime validation, so although a more-strict schema would be fine for use in contract testing, I think there's value in having a more accurate specification

alext:
I suppose the question I was asking is: what do we want to allow in a base-path.  At the moment we try to keep them all to alphanumeric and `-`.  It seems that it would be good to enforce that (it's always easier to relax a restriction that it is to tighten it...).

I see your point about this being for contract testing though, if we decide to do this, we'd need to tighten the validations in the content-store at the same time.

heathd:
@jamiecobbett ok I've tightened up the restrictions for those other formats as well (cc @jennyd)

PR #98, opened 2015-07-20
Remove change_note from specialist documents

heathd:
The `change_note` field contains the same content as the last entry in
the `change_history` (the last entry is the most recent one).

It's superfluous to duplicate that information in an additional
top-level field, and doing so introduces potential for confusion and
errors.

At present it's believed that the email alert api would use the
`change_note` field when listening on the message queue, however
specialist documents are not sent to the email-alert-api in that way and
we do not intend to do that in the future.

### DIFF
--- a/dist/formats/specialist_document/frontend/schema.json
+++ b/dist/formats/specialist_document/frontend/schema.json
@@ -43,8 +43,7 @@
       "required": [
         "body",
         "metadata",
-        "change_history",
-        "change_note"
+        "change_history"
       ],
       "properties": {
         "body": {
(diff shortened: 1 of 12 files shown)


danielroseman:
Should we remove this field from other formats too, eg case_study?

heathd:
@danielroseman I didn't look at case study, nor the implications of whether it's in use or not.

heathd:
But should be fine to do a PR to remove it and if the test passes to merge it

heathd:
See #100

PR #100, opened 2015-07-20
Remove change note from case_study format

heathd:
it's not used by the frontend


rboulton:
case studies can be tagged to topics.  Therefore, merging this one _would_ break any email alerts sent via email-alert-service for case studies which are tagged to topics.  See https://github.com/alphagov/email-alert-service/blob/7390cb1bd3d98deba1f50545ad7b9e95557a5574/email_alert_service/models/email_alert.rb#L44

heathd:
ok @rboulton thanks 

reflecting on this, we could actually set up contract tests to verify that the stuff sent to email-alert-service is what it needs/expects. What do you think @danielroseman?

Closing this PR

PR #103, opened 2015-07-22
Improve cma case examples

elliotcm:
This allows us to do contract testing on the
specialist documents where the related finder is
required (for example, specialist frontend).

Part of https://trello.com/c/GhoOlUBa/266-specialist-frontend-reads-data-from-content-store

Finder frontend contract tests will fail until this is merged: https://github.com/alphagov/finder-frontend/pull/226


formats/finder/frontend/examples/cma-cases.json
```diff
@@ -5,139 +5,281 @@
...
-  "updated_at": "2015-03-13T10:43:49.764Z",
-  "public_updated_at": "2015-02-12T15:43:13.000+00:00",
+  "updated_at": "2015-07-21T12:45:38.225Z",
+  "public_updated_at": "2015-04-30T10:41:26.000+00:00",
   "details": {
     "beta": false,
-    "beta_message": null,
     "document_noun": "case",
+    "facets": [{
+      "allowed_values": [{
```
danielroseman: It's a quibble but these have been reformatted in a way that makes them much harder to read for a human. The fields are in alphabetical order rather than putting the important fields (key, name) first; and the indentation makes it more difficult to scan down and see where the elements start and end.

elliotcm: I agree, but the output from the content store is all on a single line.  I've tried two JSON reformatters and this is how they reflow.

danielroseman: `python -m json.tool filename.json` does a good job.

elliotcm: @danielroseman Found a better one, updated.

PR #104, opened 2015-07-28
topic and mainstream_browse_page now use paths

alext:
collections-publisher now sends details of curated list contents as
an array of base_paths as opposed to api_urls. Update the schema and
examples accordingly. This also validates that the values given are all
valid absolute URL paths.


tijmenb:
:+1:

PR #105, opened 2015-07-31
Allow Finders to be Alpha

tommyp:
Same as the Beta, this commit adds an `alpha` boolean and an option
`alpha_message` on the details of the Finder. In Finder Frontend this is
then used to render the Alpha Banner.

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -46,6 +46,26 @@
         "facets"
       ],
       "properties": {
+        "alpha": {
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        "alpha_message": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
         "beta": {
           "anyOf": [
             {
--- a/dist/formats/finder/publisher/schema.json
+++ b/dist/formats/finder/publisher/schema.json
@@ -92,6 +92,26 @@
         "facets"
       ],
       "properties": {
+        "alpha": {
+          "anyOf": [
+            {
+              "type": "boolean"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        "alpha_message": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
         "beta": {
           "anyOf": [
             {
--- a/formats/finder/publisher/details.json
+++ b/formats/finder/publisher/details.json
@@ -8,6 +8,18 @@
     "facets"
   ],
   "properties": {
+    "alpha": {
+      "anyOf": [
+        { "type": "boolean" },
+        { "type": "null" }
+      ]
+    },
+    "alpha_message": {
+      "anyOf": [
+        { "type": "string" },
+        { "type": "null" }
+      ]
+    },
     "beta": {
       "anyOf": [
         { "type": "boolean" },


tijmenb:
When I added the beta boolean I didn't think of this case. In retrospect I would've added a `phase_state` field instead, with "beta" and "alpha" as allowed values. Would it be better to introduce that now and deprecate the `beta`?

bradleywright:
Agree with Tijmen.

dsingleton:
Just merged https://github.com/alphagov/finder-frontend/pull/228#issuecomment-127157549, but thought the same about a phase state field. If we're standardising it here, is it worth making it part of the base content item, rather than finder specific?

bradleywright:
@dsingleton I think so.

tommyp:
What's the process for adding something into the base content item?

tommyp:
I haven’t anything to base content item and I thought the idea is that every app publishes the base part as the same. So my question is more about if it’s modifying the base do I need to modify every app?

edds:
I believe the base is all defined in here: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/metadata.json which is then used to generate all the other app schemas.

tommyp:
I get the process for doing it _here_, but won't that then make every App which publishes to the Content Store fail?

bradleywright:
I would have thought it'd be either allowed to be non-existent, or one of `alpha` or `beta`. That way apps not sending it could carry on doing so and never be in a phase.

alext:
Content-store will need to be updated to understand this filed though.

heathd:
As @alext points out, it's going to be a some extra work in content store to make this field a 'general' one in the base metadata. 

I'm also doubtful that this is a piece of information that's likely to be applicable to other formats. For example it's unlikely that we'd mark an individual AAIB report or Case Study as 'beta'. Generally the 'alpha/beta' phases labels are applicable to a whole format, rather than an individual document. The finders are a slightly unusual case in that a single finder in some way represents a format.

I'd suggest that the `phase` field should stay in the `details` section of finders for now, and if another format comes along which needs it, then we can look at promoting it to be a base metadata field.

edds:
I don't really mind if this gets put into the base object or kept in the details hash. However, a quick look through some of the places we are using beta labels the following are conditional based on the content: [manuals](https://github.com/alphagov/manuals-frontend/blob/d3b0f8ff97edaea3fb29a4c23978f44399b3f512/app/views/manuals/_beta_label.html.erb), [collections](https://github.com/alphagov/collections/blob/822a957789a3a7960fb49dce396ba824362259fa/app/views/topics/_subtopic.html.erb#L33-L38), [specialist-frontend](https://github.com/alphagov/specialist-frontend/blob/3f688412ac99cc776ed2b247041c0472bb072b53/app/views/specialist_documents/show.html.erb#L2-L7), [frontend](https://github.com/alphagov/frontend/blob/aed183cf3ed6a1e77cf3ec11f7dd6c238a7557cf/app/views/root/_base_page.html.erb), and [finder-frontend](https://github.com/alphagov/finder-frontend/blob/6a0fb226f7f978da37dcba105d8fb6b71820a21a/app/views/finders/show.html.erb#L4-L16). I think there is pretty good prior art to say that we are going to need the ability to conditionally show the alpha or beta  messaging in frontends based on the content being displayed.

bradleywright:
We've previously had discussions about marking content in transition as "beta" while the org moves, so it's definitely been used on a per document level as well as per format.

heathd:
ok, I hadn't been aware of all those uses, I stand corrected.

tommyp:
Closing this as it has been superseded by https://github.com/alphagov/govuk-content-schemas/pull/108

PR #106, opened 2015-08-06
Add schema for special route format.

elliotcm:
This format is used to reserve & register URLs for a given base path.
These will be typically registering special-case routes such as

- static files (robots.txt, google verification pages)
- legacy prefixes (/government, /info)
- /search, /, etc

reducing the chance of major routes such as /government being accidentally
overwritten when publishing other content.

These documents will be published by the relevant owning applications on deploy.

The hope is that many of these routes will be handled by a more traditional
publisher in the future, but there is no current plan for implementing that.

Note that this schema has a restricted set of properties at the metadata level,
and no details or links hashes at all. Therefore, it is not constructed by
combining sub-schemas. In addition, there is no frontend version, because
redirects are never seen by the frontend: everything is handled by the router.

https://trello.com/c/dgrgZWTb/290-write-schemas-for-non-content-route-items


tommyp:
Can we get an example?

formats/special_route/publisher/schema.json
```diff
@@ -0,0 +1,37 @@
...
+      "type": "string"
+    },
+    "routes": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/route"
+      }
+    }
+  }
+}
```
danielroseman: If this schema isn't going to include the metadata or details files, it needs to provide the definitions of `absolute_path` and `route` directly.

elliotcm: I thought about that, but this one doesn't: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/redirect/publisher/schema.json#L13

elliotcm:
@tommyp Example gist: https://gist.github.com/elliotcm/cddd6ea1f4e3989009bd
from https://gov-uk.atlassian.net/wiki/pages/viewpage.action?pageId=34209920

tommyp:
@elliotcm Should we add that to the Repo?

elliotcm:
@tommyp I was basing mine on the redirects schema which doesn't have any examples, because the examples are explicitly for the frontend rather than the publishers.  I don't mind adding a publisher example if that's a thing we do now.

tommyp:
Maybe I'm misunderstanding, but from what I thought there should be an example if you can visit `/api/content#{base_path}` and see the content item?

elliotcm:
You can always do that because it's direct access to the content store, but none of the frontends will actually use these items.

@danielroseman should we be extending this repo to include publisher examples for these special cases?  If so in what way?

formats/special_route/publisher/schema.json
heathd: hmm, that was an oversight as well. I guess the fact is that these schemas are not getting used anywhere (neither within this repo, nor by publishing clients), so the bug was not picked up.

Maybe there's a case for adding support for publisher examples as well, I'd be happy to talk through with you what might be involved @elliotcm, or pair on it if you'd like.

elliotcm: OK.  It looks like we're negotiating specifics of implementation rather than design at this point, so I'll carry on with the ticket to implement this in the various apps today, and we can sort this PR tomorrow when we're both in the office.

danielroseman:
LGTM but someone who knows makefiles better than me might want to take a look at that file.

elliotcm:
I am not one of those someones, alas.

heathd:
if you want to test it you could do a `make clean` and then `make` and check that it builds all the things that you would expect it to.

PR #107, opened 2015-08-10
Add #/links/topics to HMRC Manuals

mikejustdoit:
HMRC Manuals're in the process of becoming 'linkable' to topics.

It was decided that we should publish a manual's list of topics as part of its
`links` data, as an array of topic content_ids.

### DIFF
--- a/formats/hmrc_manual/publisher/links.json
+++ b/formats/hmrc_manual/publisher/links.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "topics": {
+      "$ref": "#/definitions/guid_list"
+    }
+  },
+  "definitions": {
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


formats/hmrc_manual/publisher/links.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "topics": {
+      "anyOf": [
+        { "$ref": "#/definitions/guid_list" },
+        { "type": "null" }
+      ]
```
alext: I don't think this `anyOf` stuff is necessary.  These fields aren't required by default (`requiredProperties` isn't set on the `links` object by default.)

mikejustdoit: Yep, @alext, from the looks of existing schemas it I thought so too. But then the schema validation specs in `hmrc-manuals-api` failed (https://github.com/alphagov/hmrc-manuals-api/blob/master/spec/models/publishing_api_manual_spec.rb#L24) with `Errors: ["The property '#/links/topics' of type NilClass did not match the following type: array in schema file:///var/govuk/govuk-content-schemas/dist/formats/hmrc_manual/publisher/schema.json#"]` which I did not understand, but the `"anyOf"` fixed

alext: We should never be sending `nil` in this field.  It should either be an array of GUIDs , or it should be not set at all.

mikejustdoit: OK, I officially don't understand these schemas

mikejustdoit: Right, maybe I've found the problem

mikejustdoit:
These changes originally added the field as an `anyOf` `guid_list` or `null`, but I've now updated it to allow just a `guid_list`

tommyp:
@alext You OK with this being merged now?

PR #108, opened 2015-08-11
Add Phase for alpha or beta

tommyp:
This PR adds a phase attribute for a Content Item. It can only be present with
`alpha` or `beta` as the attributes. This can then be used by the Rendering
Application to switch functionality or visual features based on the phase.

### DIFF
--- a/formats/finder/frontend/examples/all_policies.json
+++ b/formats/finder/frontend/examples/all_policies.json
@@ -1,0 +1,37 @@
+{
+  "base_path": "/government/policies/all",
+  "title": "All policy content",
+  "description": "",
+  "format": "finder",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "2015-07-28T13:31:24.082Z",
+  "public_updated_at": "2015-07-28T13:29:02.000+00:00",
+  "phase": "alpha",
+  "details": {
+    "document_noun": "documents",
+    "facets": [],
+    "filter": {},
+    "reject": {
+      "policies": [
+        "_MISSING"
+      ]
+    },
+    "show_summaries": false
+  },
+  "links": {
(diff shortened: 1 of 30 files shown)


formats/finder/frontend/examples/all_policies.json
```diff
@@ -0,0 +1,1109 @@
+{
+  "base_path": "/government/policies/all",
+  "title": "All policy content",
+  "description": "",
+  "format": "finder",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "2015-07-28T13:31:24.082Z",
+  "public_updated_at": "2015-07-28T13:29:02.000+00:00",
+  "phase": "alpha",
```
jamiecobbett: This an enormous example to include just to demonstrate the "phase" key. Could we use a shorter one?

tommyp: I thought I'd add the only real world example which will use it (so far)?

tommyp: But fair enough, I'll get rid of the `facets` since it seems to be the majority of the example.

heathd:
@tommyp after removing the 'phase' from `frontend_links_definition.json` you'll also need to re-run `make` to regenerate the frontend `schema.json` files.

tommyp:
@heathd Good point. Done!

alext:
IMO, content-store should be updated to support this field before this is merged.

tijmenb:
https://github.com/alphagov/content-store/pull/138 has been merged [and deployed](https://github.com/alphagov/content-store/compare/deployed-to-production...master), so I guess this is good to merge too, right?

tommyp:
@tijmenb Yeah. This can be merged now.

PR #109, opened 2015-08-13
Add analytics identifier to root metadata

edds:
So that we can easily expand the analytics identifiers in the links hash
add it to the root metadata.

The use case for this is that we have many things in the content store
that have short codes we use in analytics (to get around character
limits in dimensions). Any page which gets linked to any of these
objects should have those short codes exposed on the frontend. Adding
these codes to the links hash means that frontends won't need to make a
query to the content store for every object in the links hash. Having
the identifiers in the root object rather than in the details hash means
that we don't need to load the entire details hash to generate the links
hash objects.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -37,6 +37,9 @@
         "type": "string"
       }
     },
+    "analytics_identifier": {
+      "type": "string"
+    },
     "details": {
       "type": "object",
       "additionalProperties": false,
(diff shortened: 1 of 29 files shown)


jamiecobbett:
Presumably there'll be a follow-up to remove them from the details hash?

alext:
This field should be added to the content-store before this is merged.

jamiecobbett:
Merge when ready :+1:

edds:
Now that content-store has been updated to handle this are we good to merge this?

PR #110, opened 2015-08-13
Auto-map from file name to document type in convert script

robmckinnon:
* Automate mapping from file name to document type when mapping not already provided in schema convert script.
* Log when implicit mapping has occurred, and inform how to override by adding an explicit mapping.
* Reduces the need to re-edit the script for future document types that adhere to usual file naming convention.
* Remove mappings that follow the usual naming convention.

### DIFF
--- a/bin/convert_finder_schemas
+++ b/bin/convert_finder_schemas
@@ -24,11 +24,9 @@ class DocumentTypeMapper
     "vehicle-recalls-and-faults-alert.json" => "vehicle_recalls_and_faults_alert"
   }
 
-  def call(filename)
-    file_name = File.basename(filename)
-    begin
-      MAPPING.fetch(file_name)
-    rescue Exception => e
+  def call(filepath)
+    file_name = File.basename(filepath)
+    MAPPING.fetch(file_name) do |file_name|
       document_type = file_name.chomp('s.json').gsub('-','_')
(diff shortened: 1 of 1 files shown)


elliotcm:
We could probably remove the standard ones from the mapping now, right?

robmckinnon:
Ok, I can remove standard ones. I'll add to this PR.

bin/convert_finder_schemas
```diff
@@ -34,7 +33,16 @@ class DocumentTypeMapper
   }
 
   def call(filename)
-    MAPPING.fetch(File.basename(filename))
+    file_name = File.basename(filename)
+    begin
+      MAPPING.fetch(file_name)
+    rescue Exception => e
```
danielroseman: The [style guide](https://github.com/alphagov/styleguides/blob/master/ruby.md#exceptions) discourages the use of exceptions for things like this. Rather, we can just look up the mapping, and check if a value is returned:

    document_type = MAPPING[filename]
    if document_type.nil?
       ....

robmckinnon: Ok, I'll change now.

jamiecobbett: I agree with @danielroseman.

As an aside, rescuing `Exception` is bad practice as it is the root class of the error class hierarchy, so this code would rescue things like `SystemExit`. Rescuing `KeyError` would be the way to go.

PR #112, opened 2015-08-21
Add topic tags to HMRC Manuals

mikejustdoit:
This change adds 'tags' under 'details', and 'topics' under those 'tags'.

This is the format expected by Email Alert Service in the Current World, as
demonstrated in the Case Study schema.

### DIFF
--- a/dist/formats/hmrc_manual/frontend/schema.json
+++ b/dist/formats/hmrc_manual/frontend/schema.json
@@ -146,6 +146,19 @@
               }
             }
           }
+        },
+        "tags": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "topics": {
+              "description": "Slugs of the Manual's topics. An example of a topic slug is 'business-tax/vat'.",
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          }
         }
       }
     },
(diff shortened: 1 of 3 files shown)


jennyd:
Could you mention in the commit message that the topics in the details hash are slugs, not content_ids? The schema alone doesn't make that clear.

Since we're also going to send the topics' content_ids in the links hash, we need to add that field to the schema too.

jamiecobbett:
@jennyd json-schema also provides the `description` property for describing fields in the schema.

mikejustdoit:
@jennyd, I've edited the commit message to explain what's expected in that new field. And added a `description` (thanks @jamiecobbett, I hadn't noticed that before).

Is that `description` clear enough?

The `links` were already merged (https://github.com/alphagov/govuk-content-schemas/commit/b869cee746fff67b8b1fc2e6d29f138ab5fb226e), but if we're happy with the description text, I'll open another pull request and add it to the links too

jennyd:
@jamiecobbett thanks, I'd forgotten about `description`.

@mikejustdoit since we use 'slug' to refer to different parts of the path in different applications, could you add an example topic slug to the description? Without context it's not obvious whether a slug should be `topic/business-tax/vat`, `business-tax/vat` or `vat`.

Sorry, I didn't spot that the `links` part was already done. The topics in `links` have a pattern which makes it clearer that they're expected to be content_ids, but having a description there as well can't hurt :)

mikejustdoit:
@jennyd, I've added an example topic slug as suggested

PR #113, opened 2015-08-27
Validate that redirect destinations are paths

jamiecobbett:
content-store rejects absolute URLs, so the schema should too.

### DIFF
--- a/formats/redirect/publisher/schema.json
+++ b/formats/redirect/publisher/schema.json
@@ -52,7 +52,8 @@
   "definitions": {
     "absolute_path": {
       "type": "string",
-      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?$"
+      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?$",
+      "description": "A path only. Query string and/or fragment are not allowed."
     }
   }
 }
--- a/formats/redirect/publisher/schema.json
+++ b/formats/redirect/publisher/schema.json
@@ -37,13 +37,13 @@
         ],
         "properties": {
           "path": {
-            "type": "string"
+            "$ref": "#/definitions/absolute_path"
           },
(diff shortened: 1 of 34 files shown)


rboulton:
Did you mean to say "content-store requires absolute URLs" - I think your comment is the wrong way round, but the code is right.

jamiecobbett:
@rboulton I think that I've got it right - it requires absolute paths, and so does this validation.

However I think this will reject paths with querystrings, which isn't what we want for redirect destinations.

rboulton:
I was saying that your comment on this PR description said "rejects absolute URLs", but should have said "requires".

I think you're right that this will reject paths with querystrings, which we don't want.  Not sure of the best fix.

alext:
The validation for redirect destinations in the content-store is less strict that for other paths.  Specifically, it allows query-strings and fragments - https://github.com/alphagov/content-store/blob/master/app/models/registerable_redirect.rb#L28-L29.

I think this validation here is therefore too strict.

jamiecobbett:
@alext thanks. Rebased to allow destinations to include fragments and querystrings.

formats/redirect/publisher/schema.json
```diff
@@ -52,7 +52,13 @@
   "definitions": {
     "absolute_path": {
       "type": "string",
-      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?$"
+      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?$",
+      "description": "A path only. Query string and/or fragment are not allowed."
+    },
+    "absolute_fullpath": {
```
alext: I'm not convinced by the name of this definition, but finding it hard to come up with a better one.  Maybe `absolute_path_with_query_or_fragment`...

alext:
Minor point about naming, otherwise looks good to me.

formats/redirect/publisher/schema.json
jamiecobbett: I agree. Naming of parts/combinations of parts of URLs is difficult. Here's where I got the idea for this name:
http://api.rubyonrails.org/classes/ActionDispatch/Request.html#method-i-fullpath

boffbowsh:
:+1: Looks ok to me but I really know nothing about this repo at present

PR #114, opened 2015-09-03
Allow service design phase to be 'live'

alext:
Given live is one of the [service design phases](https://www.gov.uk/service-manual/phases) it makes sense for it to be one of the allowed options. This will allow publishing apps to be simpler, avoiding examples like https://github.com/alphagov/collections-publisher/blob/master/app/presenters/tag_presenter.rb#L60-L63

Before this is merged, the [content-store validations](https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L61-L64) will need to be updated to allow this value. I'd also suggest that the field should be defaulted to 'live' when not explicitly set in an incoming request.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -41,10 +41,12 @@
       "type": "string"
     },
     "phase": {
+      "description": "The service design phase of this content item - https://www.gov.uk/service-manual/phases",
       "type": "string",
       "enum": [
         "alpha",
-        "beta"
+        "beta",
+        "live"
       ]
     },
     "details": {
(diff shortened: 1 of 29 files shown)


rboulton:
This seems like a good idea.  When you say "defaulted to 'live'", are you referring to the frontend representation?

alext:
> When you say "defaulted to 'live'", are you referring to the frontend representation?

I meant in the content-store database (and therefore also in all "downstream" representations). This is similar to the locale field which defaults to `en`.

alext:
I've opened a PR on content-store to implement this suggestion - alphagov/content-store#146

PR #116, opened 2015-09-04
Add asylum support decision schema and examples

robmckinnon:
Add schema and examples for asylum support decisions - a new specialist document type.

### DIFF
--- a/dist/formats/specialist_document/frontend/schema.json
+++ b/dist/formats/specialist_document/frontend/schema.json
@@ -132,6 +132,9 @@
           "$ref": "#/definitions/aaib_report_metadata"
         },
         {
+          "$ref": "#/definitions/asylum_support_decision_metadata"
+        },
+        {
           "$ref": "#/definitions/cma_case_metadata"
         },
         {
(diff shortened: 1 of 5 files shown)


tommyp:
Can you split the commit adding the examples so that the Finder and Specialist Document examples are added in their own commits? 

Although Finders and Specialist Documents are tied for Asylum Support Decisions, they can be independent from each other. It may make more sense to have them added in their own commits with some information about the format such as:

- what it is
- who uses it
- any information about this format which may be useful

robmckinnon:
Happy to do a generic description for each tribunal example in a commit message. @tomwm has offered to help write them.

However we do note that in the past Finder and Specialist Document examples have been added in bulk in a single commit without this prose about the format and usage. e.g. bulk commit here: https://github.com/alphagov/govuk-content-schemas/commit/ffef69efe2c9ceb59a94019e31846106a79bd87a

tomwm:
Finders:
- x tribunal publishes decisions that describe the outcome of a tribunal case
- tribunal finders are used by public, professionals and the tribunals themselves to search for past decisions 
- published decisions form a part of case law and are used for research purposes and to prepare for current or future cases.

Specialist documents:
- An example of a decision from x tribunal

robmckinnon:
@tommyp As suggested, the Finder and Specialist Document examples are now in separate commits, with contextual commit messages covering usage.

tommyp:
@robmckinnon Ace! Thanks you. 👍

PR #117, opened 2015-09-07
Add "Gone" format

tijmenb:
From the [content-store docs](https://github.com/alphagov/content-store/blob/0e96e934206abd54e057e63a7eb6beaf71d21f60/doc/gone_item.md
):

> To represent content that is no longer available, the content store supports items with a format of "gone". These will cause a gone route to be setup in the router so that the item returns a 410 HTTP status.

This item is based on [the redirect format](https://github.com/alphagov/govuk-content-schemas/tree/master/formats/redirect).

This format will be used for testing the creation of gone routes for draft content-items in https://github.com/alphagov/collections-publisher/pull/138.

Related Trello: https://trello.com/c/ybdnGMhl

### DIFF
--- a/formats/gone/publisher/schema.json
+++ b/formats/gone/publisher/schema.json
@@ -1,0 +1,54 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "format",
+    "publishing_app",
+    "update_type",
+    "routes"
+  ],
+  "properties": {
+    "base_path": {
+      "$ref": "#/definitions/absolute_path"
+    },
+    "content_id": {
+      "type": "string",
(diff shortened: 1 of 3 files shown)


formats/gone/publisher/schema.json
```diff
@@ -0,0 +1,54 @@
...
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "path",
+          "type"
+        ],
+        "properties": {
+          "path": {
+            "type": "string"
```
alext: This should be an `absolute_path` like `base_path` is...

jamiecobbett:
:+1: for intent

formats/gone/publisher/schema.json
tijmenb: I didn't find any entry for `routes/properties/path` with a type other than `string`. Would it be better to be consistent?

PR #119, opened 2015-09-08
Move definition for `guid_list` to metadata.json

tijmenb:
I noticed the `guid_list` pattern being repeated in many formats. We need that for the tagging, so it seemed good to move this up to metadata.json. This removes the duplication, and add `guid_list` to formats that previously didn't have it. This should not be a problem. For formats that previously had a `guid_list` defined only the position in the dist json's is changed.

@rboulton @heathd I'm still getting acquainted with these schemas - are changes like this good, or is there a point in the duplication?

For: https://trello.com/c/MjwwKvPd

Commit to review: https://github.com/alphagov/govuk-content-schemas/commit/4fe544ac9cce061b21b4b62d7924ca18f3cbd537 (the other is the result of `make`)


jamiecobbett:
Looks good to me :+1:

rboulton:
LGTM too - duplication in `dist/*` is fine - duplication elsewhere is best avoided if possible.

PR #120, opened 2015-09-09
Make sure we can tag every format

tijmenb:
Currently only some formats can have tags in the links hash. This PR adds `topics`, `mainstream_browse_pages` and `policy_areas` to all formats. 

It does so by adding a `base_links.json` file, which is merged into the `links` hash. To provide this functionality I've first [refactored the schema combiner a bit](https://github.com/alphagov/govuk-content-schemas/commit/af577609fe4ce304dee3aa6a92c0458660387fee). The [second commit](https://github.com/alphagov/govuk-content-schemas/commit/7222cebc6ba63e9e798d5763bdd15c47f0e9eaed) then makes the actual change. 

The last commit is regeneration noise, so best reviewed per-commit.

Trello: https://trello.com/c/MjwwKvPd


alext:
While we're here, it's probably worth adding description fields to the base links.

rboulton:
:+1:, and would be :100: with description fields.

tijmenb:
What should the descriptions be?

screen shot 2015-09-09 at 16 43 40

Seems kinda obvious.

rboulton:
Great descriptions :)

PR #121, opened 2015-09-14
Expose content_id in frontend representations.

alext:
This includes the content_id in the frontend representation of content
items - both in the main item, and in the expanded links.

It is becoming necessary for some of the frontend applications to know
about content_ids (eg email_alert_frontend needs the topic content_ids
in order to send them to email-slert-api to create the link to
govdelivery topics).

Trello: https://trello.com/c/7K6FySvv

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -12,6 +12,9 @@
     "base_path": {
       "$ref": "#/definitions/absolute_path"
     },
+    "content_id": {
+      "$ref": "#/definitions/guid"
+    },
     "title": {
       "type": "string"
     },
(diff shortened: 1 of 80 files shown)


jamiecobbett:
Code looks good. Is there already consensus about the intent?

alext:
One thing to note:

I've made the content_id field mandatory in the links expansion. This seemed reasonable given that items are linked by content_ids, and therefore it shouldn't be possible to have a linked item without a content_id.

There's a snag with this though.  Currently the available_translations expansion [returns only self](https://github.com/alphagov/content-store/blob/master/app/models/content_item.rb#L159) for an item without a content_id. I'm not sure this behaviour is relied on anywhere. Nor does it seem correct. I think we should therefore update content-store to not return available_translations for an item without a content_id.

rboulton:
I think we have sufficient consensus: at a minimum, @bradwright and @elliotcm have been involved in discussions where we've said we need to do this at some point.

Regarding the issue with translations; I agree that behaviour doesn't seem correct.  I don't think that should block merging this PR, though, but it would be good to address soon.

Also, we should "soon" not have any content in content-store without a content ID

PR #125, opened 2015-10-07
Tweak example datetime format...

tuzz:
We're using this example to generate the contract
tests in gds-api-adapters. By default, Rails formats
JSON with a particular datetime format. This didn't
match the one in the example.

Both of these formats are valid with the same RFC:
http://tools.ietf.org/html/rfc3339

This is a simple fix that resolves the problem.

### DIFF
--- a/formats/special_route/publisher/examples/robots.txt.json
+++ b/formats/special_route/publisher/examples/robots.txt.json
@@ -4,7 +4,7 @@
   "title": "Instructions for crawler robots",
   "description": "robots.txt provides rules for which parts of GOV.UK are permitted to be crawled by different bots.",
   "format": "special_route",
-  "public_updated_at": "2015-07-30T13:58:11+00:00",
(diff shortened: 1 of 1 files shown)


alext:
I'm not sure I understand why this is necessary. As it stands, it matches what the publishing apps generate because they call `Time.now.iso8601` - https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/publishing_api/special_route_publisher.rb#L29

It would be good to keep the examples matching what's actually produced by the publishing apps.

heathd:
@alxt does it matter if they are identical provided that the time formats conform to schema (which specifies [`date-time`](http://spacetelescope.github.io/understanding-json-schema/reference/string.html#built-in-formats)?

tuzz:
We've discussed this and we're going to take a different approach.

Thanks

PR #131, opened 2015-10-08
Allow running dummy-content-store from this repo

dsingleton:
The behaviour of the `govuk-dummy_content_store` gem used to live in this repo, but was split out, however, it's still most common to want to run the dummy content store in combination with this repo.

This adds the gem as a dependency, so it's easy to run the dummy content store, and requires fewer repos (just the FE and the dummy content store) which is much simpler for less technical folk to run.

I've also added some documentation to explain how to do this.

(I chatted to @heathd  about this in advance, and he was happy with the change.)

This also adds support for running a Heroku instance of the dummy content store, which will be configured to run `master` and update on change, so there is a publicly addressable version of the content-store that people can use, rather than running their own version.

This is primarily for the ideal frontend/design workflow, where the ideal situation is running a single FE application, and pointing at 'live' instances of `static` and `content-store`, which should be the only dependencies. (but it may be useful to other folks too)


docs/running-frontend-against-examples.md
```diff
@@ -0,0 +1,23 @@
...
+The dummy content store allows you to run a frontend using the examples in this repo.
+
+It has the same API as the real [content-store](https://github.com/alphagov/content-store.)
+
+You will not need to run [content-store](https://github.com/alphagov/content-store] itself.
+
+From this repo, run:
+
+```
+dummy_content_store
```
heathd: Maybe write this as:

```
$ bundle exec dummy_content_store
```

- the `$` prefix is an idiomatic way of indicating that this is something that you type into a command line shell
- I think it's helpful to remind people to run `bundle exec` as well, although it may work without (depending on your setup), using `bundle exec` will always work

heathd:
:+1: yay just a teeny comment on the readme (and needs a rebase)

docs/running-frontend-against-examples.md
dsingleton: Good point. I've updated that to be use the bundle convention.

dsingleton:
The `govuk-dummy-content-store` gem needs a new release, so closing this until we've done that, i'll reopen with the new version.

PR #132, opened 2015-10-12
Don't expected Dir.glob to return ordered filenames

AndrewVos:
Depending on the OS, the results of Dir.glob might not be in the order
you expect them. http://ruby-doc.org/core-2.1.2/Dir.html#method-c-glob

This causes this test to fail sometimes.


jamiecobbett:
That old :chestnut:

PR #135, opened 2015-10-15
Add a schema for service manual publisher guide

tadast:
We're working on [Service Manual Publisher](https://github.com/alphagov/service-manual-publisher) and are close to getting a prototype of a guide format merged in there. 

[One of the tests](https://github.com/alphagov/service-manual-publisher/commit/73f991e85ed7363a3816c2cac2bcc6c79572b7a2) relies on changes in this branch, I believe this schema represents the service manual guide format quite well and no major changes will need to be done, so this should be a good time to merge.

/cc @KushalP

### DIFF
--- a/dist/formats/service_manual_guide/frontend/schema.json
+++ b/dist/formats/service_manual_guide/frontend/schema.json
@@ -1,0 +1,192 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "format",
+    "locale",
+    "public_updated_at"
+  ],
+  "properties": {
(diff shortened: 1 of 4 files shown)


tadast:
Rebased against master and addressed the comments in [this commit](https://github.com/alphagov/govuk-content-schemas/commit/42527e3047e36d32224117df2369c36d874af75d#commitcomment-13850098).

KushalP:
You'll need to make a PR to bump the version of this gem.

alext:
> You'll need to make a PR to bump the version of this gem.

That's not the case for this repo.  It's not released as a gem. Things just test against master at the moment.

The related gem is merely the test helpers, and doesn't contain the schemas themselves.

PR #139, opened 2015-10-30
Add schemas for travel-advice (index and country pages)

alext:
I believe this contains everything necessary to render these pages. I'd appreciate feedback on whether there's anything I've missed, or whether any of this should be modelled differently.

### DIFF
--- a/formats/travel_advice_index/publisher/links.json
+++ b/formats/travel_advice_index/publisher/links.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+     "organisations": {
+       "$ref": "#/definitions/guid_list"
+     },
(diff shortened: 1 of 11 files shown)


rboulton:
@danielroseman Is anyone from your team looking at this? It now passes all the schema checks (I rebased against master).

danielroseman:
We've put this aside until we're completely finished with the write api documentation, we'll be getting back to it soon.

formats/travel_advice_index/publisher/links.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+     "organisations": {
```
tijmenb: `organisations` is included `base_links.json` (https://github.com/alphagov/govuk-content-schemas/pull/153), so we don't need it here anymore.

PR #140, opened 2015-10-30
Add parent link to email_alert_signup schema

mobaig:
As part of the effort to convert email alerting to rely on content IDs
rather than slugs, this change adds a field to email alert signup pages
that provides a reference to the parent piece of content. This makes it
easier for the email-alert-frontend to obtain the correct content ID to
send on to the email-alert-api.

### DIFF
--- a/dist/formats/email_alert_signup/publisher/schema.json
+++ b/dist/formats/email_alert_signup/publisher/schema.json
@@ -139,6 +139,11 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "parent": {
+          "description": "The parent content item",
+          "$ref": "#/definitions/guid_list",
+          "maxItems": 1
+        },
         "mainstream_browse_pages": {
           "description": "Powers the /browse section of the site. These are known as sections in some legacy apps.",
           "$ref": "#/definitions/guid_list"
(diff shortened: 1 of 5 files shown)


dist/formats/email_alert_signup/publisher/schema.json
```diff
@@ -139,6 +139,13 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "parent": {
+          "type": "array",
+          "maxItems": 1,
```
tommyp: Would there be a situation when a signup page might have more than 1 parent?

mobaig: Not that I've seen thus far. Also seems a little counter-intuitive. I've just realised I can use the pre-existing `guid` or `guid_list` definitions here instead. I could use the latter and call this property `parents` (plural), but can't help but feel this is a case of YAGNI.

alext: We should definitely be using the existing definitions (see any other format with links for example).

Also, "parent" is probably right name.  We've already used that for topics - https://github.com/alphagov/govuk-content-schemas/blob/master/formats/topic/publisher/links.json

mobaig: @alext @tommyp updated!

PR #141, opened 2015-11-09
Make content_id a required field and update coming soon format

elliotcm:
- Makes content_id required for all formats
- Adds a links hash for the coming soon format

### DIFF
--- a/formats/coming_soon/publisher/links.json
+++ b/formats/coming_soon/publisher/links.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "can_be_replaced_by"
+  ],
+  "properties": {
+    "can_be_replaced_by": {
+      "$ref": "#/definitions/guid_list"
+    }
+  }
+}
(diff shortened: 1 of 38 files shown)


formats/coming_soon/publisher/links.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "replaced_content_item"
+  ],
+  "properties": {
+    "replaced_content_item": {
```
jamiecobbett: I understand we want consistency, but this name is confusing for coming_soon items because they are _replaced by_ the item that I think this will link to.

elliotcm: I agree, though it's less that we want consistency so much as we programmatically need it.  Two options: 1) this is a temporary format, or 2) can you think of a name that's more flexible?

jamiecobbett: `will_be_replaced_by_content_item`? `precursor_to_content_item`?

elliotcm: After much discussion, we're going with `can_be_replaced_by` since that's a clearer description of what it's actually for.

elliotcm:
@jamiecobbett updated.

PR #142, opened 2015-11-10
Add `replaced_content_item` for redirects

tijmenb:
`replaced_content_item` is the optional content_id of the item this redirect is replacing.

It allows publishing-api to add special behaviour for putting back redirects.

It will also be added to the coming_soon format (https://github.com/alphagov/govuk-content-schemas/pull/141).

/cc @elliotcm

### DIFF
--- a/dist/formats/redirect/publisher/schema.json
+++ b/dist/formats/redirect/publisher/schema.json
@@ -47,6 +47,17 @@
           }
         }
       }
+    },
+    "links": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "replaced_content_item": {
+          "description": "The optional content_item that this redirect has replaced.",
+          "$ref": "#/definitions/guid_list",
+          "maxItems": 1
+        }
+      }
     }
   },
   "definitions": {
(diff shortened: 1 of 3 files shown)


dist/formats/redirect/publisher/schema.json
```diff
@@ -59,6 +70,16 @@
       "type": "string",
       "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?(\\?([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?(#([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?$",
       "description": "A path with optional query string and/or fragment."
+    },
+    "guid": {
```
elliotcm: We can use this definition up in the `content_id` field now, too.

tijmenb: Wonderful! Added commit.

PR #143, opened 2015-11-10
Add email_alert_type to email_alert_signup details hash

mobaig:
This states explicitly the kind of page a given email signup is handling.
This improves clarity when viewing an email-signup content item retrieved
from the content store, and also helps the email-alert-frontend (for example)
resolve how to format its subscription payload for the email-alert-api.

The goal of all this is to handle email subscription/alerting via content
IDs in the links hash and eventually move away from the use of slugs.

### DIFF
--- a/formats/email_alert_signup/publisher/details.json
+++ b/formats/email_alert_signup/publisher/details.json
@@ -7,6 +7,10 @@
     "tags"
   ],
   "properties": {
+    "email_alert_type": {
+      "type": "string",
+      "enum": ["topics", "policies"]
+    },
     "summary": {
       "type": "string"
     },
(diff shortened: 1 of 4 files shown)


formats/email_alert_signup/publisher/details.json
```diff
@@ -4,9 +4,14 @@
   "additionalProperties": false,
   "required": [
     "summary",
-    "tags"
+    "tags",
+    "email_alert_type"
```
tijmenb: if you sort the keys you'll have a cleaner diff.

tijmenb:
I think the usual workflow is to add the attribute first, get it merged, update the consumer app and then make it required in a new PR. That way you won't have a failing test here.

mobaig:
@tijmenb dropped the 'required' for now.

PR #146, opened 2015-11-11
Remove update_type from redirects schema

AndrewVos:
update_type is no longer passed through in the initial payload. It is
now the second parameter of the #publish method.

_I'm not 100% sure this is the correct thing to do, so I've opened a PR to discuss._

### DIFF
--- a/dist/formats/redirect/publisher/schema.json
+++ b/dist/formats/redirect/publisher/schema.json
@@ -5,7 +5,6 @@
   "required": [
     "format",
     "publishing_app",
-    "update_type",
     "redirects",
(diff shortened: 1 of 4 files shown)


elliotcm:
The `update_type` is still needed for apps using the v1 endpoints, so we can't remove this until everything's moved onto v2.

elliotcm:
That said, we can make it optional.  I'll do that in another PR.

elliotcm:
Making `update_type` optional is here: https://github.com/alphagov/govuk-content-schemas/pull/149

PR #147, opened 2015-11-11
Add `can_be_replaced_by` to gone items.

elliotcm:
Part of https://trello.com/c/pULByuHs/393-fix-invalid-content-store-data

Related to https://github.com/alphagov/whitehall/pull/2373

### DIFF
--- a/dist/formats/gone/publisher/schema.json
+++ b/dist/formats/gone/publisher/schema.json
@@ -43,12 +43,33 @@
           }
         }
       }
+    },
+    "links": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "can_be_replaced_by": {
+          "description": "Optional.  The content_item that this redirect has replaced.",
+          "$ref": "#/definitions/guid_list",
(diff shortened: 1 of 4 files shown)


dist/formats/gone/publisher/schema.json
```diff
@@ -43,12 +43,33 @@
           }
         }
       }
+    },
+    "links": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "can_be_replaced_by": {
+          "description": "The optional content_item that this gone item has replaced.",
```
jamiecobbett: Would this read better as: `Optional. The content_item that this gone item has replaced.`?

elliotcm: Sure.  Updated.

danielroseman: Looks like this needs updating in two other places.

elliotcm: @danielroseman Done, and rebased.

PR #148, opened 2015-11-11
Make content ID required for redirects.

elliotcm:
Part of https://trello.com/c/pULByuHs/393-fix-invalid-content-store-data

- [ ] Update short-url-manager to send content IDs first


jamiecobbett:
The short-url-manager app doesn't do any contract testing. It publishes redirects but doesn't send content IDs, so I don't think this should be merged until that has been updated.

elliotcm:
@jamiecobbett Added as a task.

PR #150, opened 2015-11-11
Rename publisher to content_owner within Service Manual Guide

tadast:
We have realised that using publisher to indicate the group responsible for a
piece of content can be ambiguous because within GOV.UK we use the term
"publisher" to name applications that publish content.

I've used this opportunity to rename publisher.name to content_owner.title too
to make it consistent with code in other related repos and bring it in line
with "related_discussion" fields' naming conventions.

Other repos that need this renaming:
      service-manual-publisher (branch rename_publisher)
      government-frontend (branch service-manual-frontent-real-data)

We will not need to sync deploys or migrate data because none of those are used
in production.

Related: https://github.com/alphagov/government-frontend/pull/63

/cc @KushalP

### DIFF
--- a/dist/formats/service_manual_guide/frontend/schema.json
+++ b/dist/formats/service_manual_guide/frontend/schema.json
@@ -58,16 +58,16 @@
       "additionalProperties": false,
       "required": [
         "body",
-        "publisher"
+        "content_owner"
       ],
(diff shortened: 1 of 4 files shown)


AndrewVos:
lgtm

PR #151, opened 2015-11-12
Update email_signup_alert example

mobaig:
Change the tag in the example to 'policies' as this is what is required
from the email alert attributes exported by the policy-publisher. The
goal of this change is to ultimately ensure that the
email-alert-frontend correctly subscribes users using tags named
'policies'.

### DIFF
--- a/dist/formats/email_alert_signup/frontend/schema.json
+++ b/dist/formats/email_alert_signup/frontend/schema.json
@@ -57,8 +57,7 @@
       "type": "object",
       "additionalProperties": false,
       "required": [
-        "summary",
-        "tags"
+        "summary"
       ],
       "properties": {
         "email_alert_type": {
@@ -68,11 +67,24 @@
             "policies"
           ]
         },
-        "summary": {
-          "type": "string"
+        "signup_tags": {
+          "type": "object",
+          "description": "The content item for which this is handling email signups",
+          "additionalProperties": false,
+          "properties": {
+            "policies": {
+              "type": "array"
+            },
+            "topics": {
+              "type": "array"
+            }
+          }
         },
         "tags": {
           "type": "object"
+        },
+        "summary": {
+          "type": "string"
         },
         "breadcrumbs": {
           "type": "array",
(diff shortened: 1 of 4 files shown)


bishboria:
:+1:

rboulton:
I don't think this is the right change to make. Because of how email-alert-service interprets the "tags" field, it would mean that whenever a policy page is published as a major update (which I think is done by policy publisher by default), anyone subscribed to that policy would get an email.  This isn't desired behaviour - they should only get emails about content tagged to that policy.  It's only "luck" that this hasn't been happening already.

To be clear - I think the change from `policy` to `policies` probably _is_ correct.  However, I think we need to change the top level key in the details hash for this from "tags" to something else ("subscriber_tags" perhaps?) before we make the change.  Essentially, having "details" -> "tags" -> "policies" -> "foo" means that the content is tagged to the "foo" policy, which isn't the semantics of what we're trying to express here.

mobaig:
@rboulton changing that top-level tag sounds like the right approach. Re-working, will push updates shortly.

bishboria:
:+1:

dist/formats/email_alert_signup/frontend/schema.json
```diff
@@ -55,10 +55,8 @@
     },
     "details": {
       "type": "object",
-      "additionalProperties": false,
```
rboulton: I believe that removing additionalProperties means that any fields at all are allowed in the details hash (the schema is checked for those that are defined, but others are allowed) - I don't think this is what we want.

dist/formats/email_alert_signup/frontend/schema.json
```diff
@@ -68,12 +66,22 @@
...
+            },
+            "topics": {
+              "type": "array"
+            }
+          }
+        },
         "summary": {
           "type": "string"
         },
-        "tags": {
```
rboulton: Should we continue to allow "tags" for now, until we've migrated all items to use "signup_tags"?

dist/formats/email_alert_signup/frontend/schema.json
mobaig: Had intended this as a temporary workaround while we're dropping `tags` but on second glance your suggestion below re: leaving tags in for now is obviously much more sensible. Updating...

PR #154, opened 2015-11-19
Add `links.parent` to all formats

tijmenb:
All formats can have a parent link. This can be used to add breadcrumbs to a page.

The `email_alert_signup` and `topic` formats already had this link type, so those have been removed (this causes the hashes in `/dist` to be reordered).

For more context, see [RFC 36](https://gov-uk.atlassian.net/wiki/display/FS/RFC+36%3A+Stop+preserving+order+for+links).


rboulton:
:+1: when tests pass (unrelated fail)

tijmenb:
@rboulton tests pass now.

PR #155, opened 2015-11-19
Make content_id not required temporarily

tijmenb:
hmrc-manuals-api doesn't send the content_id yet, so making the `content_id` required makes all tests on this repo fail.

This can be reverted after hmrc-manuals-api is updated.

The test failure wasn't picked up in https://github.com/alphagov/govuk-content-schemas/pull/141, because hmrc-manuals-api schema tests weren't being run in.


tijmenb:
Closing because https://github.com/alphagov/hmrc-manuals-api/pull/121 might solve this better.

PR #156, opened 2015-11-23
Add "internal_name" to topics and browse pages

tijmenb:
The `internal_name` field will be used in admin interfaces to provide context. 

In the case of topics and browse pages, it will contain "#{parent_title} / #{child_title)", so that we can populate the selectboxes without having to expand the link set (which the publishing-api doesn't do).

screen shot 2015-11-23 at 12 24 42

I'm assuming that a generic `internal_name` field can also be used in other formats to disambiguate items for use in admin interfaces. For example, the `internal_name` of organisations could be used to contain the acronym or alternative spellings, which would make them more easily findable.

@rboulton

### DIFF
--- a/dist/formats/mainstream_browse_page/frontend/schema.json
+++ b/dist/formats/mainstream_browse_page/frontend/schema.json
@@ -57,6 +57,10 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "internal_name": {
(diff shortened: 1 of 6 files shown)


mobaig:
lgtm - rebase and i'll merge

tijmenb:
Rebased.

PR #157, opened 2015-11-23
Enforce presence of signup_tags in details, remove tags in email_alert_signups

stephen-richards:
The contents of an email-alert-signup has changed: there is no longer a 'tags' key in the details hash - it is replaced by a new 'signup-tags' key (see https://github.com/alphagov/govuk-content-schemas/commit/1bc4952cfb021698340dbb6c29df2d25570c517e for reason why).

This change amends the JSON schema to enforce those changes.

Trello: https://trello.com/c/t42ZUP2j/434-make-signup-tags-required-in-govuk-content-schemas-email-alert-signup-and-remove-tags

### DIFF
--- a/formats/email_alert_signup/publisher/details.json
+++ b/formats/email_alert_signup/publisher/details.json
@@ -3,6 +3,7 @@
   "type": "object",
   "additionalProperties": false,
   "required": [
+    "signup_tags",
     "summary"
   ],
   "properties": {
@@ -18,9 +19,6 @@
         "policies": { "type": "array" },
         "topics": { "type": "array" }
       }
-    },
-    "tags": {
-      "type": "object"
     },
     "summary": {
       "type": "string"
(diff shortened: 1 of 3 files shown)


tijmenb:
Can you reference https://github.com/alphagov/govuk-content-schemas/commit/1bc4952cfb021698340dbb6c29df2d25570c517e in the commit message? That contains the reason for this change.

formats/email_alert_signup/publisher/details.json
```diff
@@ -3,7 +3,8 @@
   "type": "object",
   "additionalProperties": false,
   "required": [
-    "summary"
+    "summary",
```
tijmenb: Not a blocker, but try to add things alphabetically sorted.

mobaig:
Maybe something like this: 

```
This is a followup to 1bc4952, which provides detail around why this change was introduced in the first place. Also note that this rename fixes a subtle bug - when this field was still called 'tags', published changes to the email_alert_signup content item would result in an email alert. This was not intended behaviour.
```

PR #160, opened 2015-11-26
Accept destination in redirect items for all formats

tijmenb:
Currently the formats use the same validation for the items in `routes` and `redirects`. 

This is probably a mistake, as an element in the `redirects` array can also have a `destination` property.

Key commit is https://github.com/alphagov/govuk-content-schemas/commit/01fb06547807ff7d187a16ed59440c17cf94e224.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -268,6 +268,29 @@
         }
       }
     },
+    "redirect_route": {
+      "type": "object",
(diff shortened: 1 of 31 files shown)


jamiecobbett:
This doesn't cover the `segments_mode` behaviour which @boffbowsh and @benlovell added. I've opened a PR here: https://github.com/alphagov/govuk-content-schemas/pull/162

PR #162, opened 2015-11-27
Add segments_mode to the redirect_route type

jamiecobbett:


tijmenb:
Cool - should this also be added to redirects in the redirect schema?

https://github.com/alphagov/govuk-content-schemas/blob/master/formats/redirect/publisher/schema.json#L28

jamiecobbett:
@tijmenb yes - added and rebased.

tijmenb:
:+1:

PR #163, opened 2015-11-27
Add publisher examples

steventux:
https://trello.com/c/R46p9TMH/366-docs-publisher-examples-for-publishing-api

Publisher examples for different content formats for both content and links where applicable.


danielroseman:
I don't think the links ones are right: they should literally just have the `links` hash.

steventux:
As discussed, the links hash isn't valid on its own, I'll look for a way to validate the /v2 links payload separate from the content payload.

elliotcm:
@danielroseman are you happy with this?

danielroseman:
Steve and I have been working on this today, more changes coming.

PR #164, opened 2015-12-01
Add publisher examples

steventux:
https://trello.com/c/R46p9TMH/366-docs-publisher-examples-for-publishing-api

- Adds publisher v2 examples
- Generates separate v2 content and links schemas in default task

/cc @jamiecobbett @danielroseman @tuzz


jamiecobbett:
I wonder if https://github.com/alphagov/govuk-content-schema-test-helpers should default to checking against v2 publisher schemas?

danielroseman:
@jamiecobbett Not sure about changing the default - it'll be more a question of changing the test configuration in publishing apps to use "publisher_v2" instead of "publisher" when calling `GovukContentSchemaTestHelpers.configure`.

steventux:
This PR still requires v1 metadata in v2 schemas which isn't right (content_id and update_type being inaccurate) so closing until we resolve this omission.

steventux:
@danielroseman updated to omit `content_id` and `update_type` from v2 schemas.

jamiecobbett:
@steventux is practical to make `base_path` required in v2 in this PR?

steventux:
@jamiecobbett thanks for spotting this, will add it to this PR :+1:

tijmenb:
@steventux I think the attributes of the links schema should be nested inside a `links` attribute:

https://github.com/alphagov/content-tagger/blob/91b7e03f10b3e1694c5546cb1efdddf9f6f962f3/app/forms/tagging_update_form.rb#L28-L34

If we want to validate the entire payload `previous_version` should probably be a valid attribute as well.

steventux:
@tijmenb thank you for spotting this, will update.

lib/govuk_content_schemas/schema_combiner.rb
```diff
@@ -60,6 +74,20 @@ def add_format_field(schema)
     end
   end
 
+  # v2 schemas do not require these properties.
+  def update_required_properties(schema)
+    return if schemas.key?(:metadata) && schemas.key?(:base_links) # v1 schema
+
+    if schema['required']
```
danielroseman: I don't think this logic really belongs in the combiner script. Is there any way we can move it into the schemas themselves?

steventux: Fair point, given what @tijmenb has commented about `previous_version` I think we need a to define v1 and v2 properties separately, preserving common properties in `formats/metadata.json` and explicitly combine what's needed...?

steventux:
Updated.
/cc @tijmenb @danielroseman @jamiecobbett

dist/formats/case_study/publisher_v2/links.json
```diff
@@ -0,0 +1,138 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "previous_version": {
+      "type": "string"
+    },
+    "base_path": {
```
tijmenb: I don't think `base_path` belongs here.

steventux: You're right, I've mistakenly combined `v2_metadata.json` when building the links schema. Will fix :+1:

steventux:
@tijmenb see https://github.com/alphagov/govuk-content-schemas/commit/01e7a6c2206992cb7e654f05b755cfb147cda571 for update

tijmenb:
:+1: looks good to me now.

PR #165, opened 2015-12-02
basic schema and example for lloyds press release content page

christophwong:
Port hard coded html pages from email-campaign-frontend to content store


formats/financial_release/publisher/details.json
```diff
@@ -0,0 +1,26 @@
...
+     "title",
+     "sub_title",
+     "content",
+     "published_date"
+   ],
+  "properties": {
+    "title": {
+      "type": "string"
+    },
+    "sub_title": {
```
boffbowsh: In other areas of GOV.UK we refer to this as a `summary`, worth keeping to the same naming scheme where possible. This is what would render a press release in Whitehall: https://github.com/alphagov/whitehall/blob/master/app/views/news_articles/show.html.erb
And the model for it if it's useful: https://github.com/alphagov/whitehall/blob/master/app/models/news_article.rb.

My bad, I should have linked you these before.

formats/financial_release/publisher/links.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  // "required": [
```
boffbowsh: I'm assuming these comments are here just for reference purposes. I don't believe you're going to need any links for financial releases themselves, so I believe if you simply don't include the links file everything will still build and validate correctly. See https://github.com/alphagov/govuk-content-schemas/blob/master/bin/combine_publisher_schema#L39 for details.

boffbowsh:
Couple of comments, otherwise looks good so far. You'll need to speak to Dai or Tom about how we make this mergeable from a fork.

christophwong:
Closing pull request, as we have decided to first integrate our frontend app with the examples before merging to master.

formats/financial_release/publisher/details.json
jamiecobbett: Does this actually fit into the `description` field that lives at the top level of the document? For context, if I recall correctly, that is also displayed underneath search results.

christophwong:
Reopening as frontend is now using the content schema for its' press release pages

PR #167, opened 2015-12-04
Add contacts format

davidbasalla:
We create the content schema for `contacts` (for how they are currently represented in the `content-store`), so that we can implement contract testing in `contacts-frontend` and `contacts-admin`. We make the change to add `organisations` to the links JSON, so that in future we can reference organisations from the content-store (currently we are passing `organisation` data in the details hash as well). 

The reason why we cannot not just switch to using the `organisations` from the `content-store` right now is that in `contacts-frontend` we require the `organisation.abbreviation` attribute, which is currently not stored for `content-store` organisations.

Improvements can be made in future PR's, removing `organisation` data from the details hash.

### DIFF
--- a/formats/contact/publisher/details.json
+++ b/formats/contact/publisher/details.json
@@ -1,0 +1,308 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "title",
+    "description"
+  ],
+  "properties": {
+    "slug": {
+      "type": "string"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "organisation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "contact_index_content_id"
+      ],
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "title": {
+          "type": "string"
+        },
+        "format": {
+          "type": "string"
+        },
+        "slug": {
+          "type": "string"
+        },
+        "abbreviation": {
+          "type": "string"
+        },
+        "govuk_status": {
+          "type": "string"
+        },
+        "created_at": {
+          "type": "string",
+          "format": "date-time"
+        },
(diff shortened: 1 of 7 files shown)


jamiecobbett:
I did write a rake task to validate existing documents against the (frontend) schema, which I think would be a useful exercise: https://github.com/alphagov/content-store/blob/master/lib/tasks/check_content_items_against_schema.rake

Hopefully it still works.

formats/contact/publisher/details.json
```diff
@@ -0,0 +1,330 @@
...
+    "slug": {
+      "type": "string"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "organisation": {
```
boffbowsh: Should this be in the links hash? Or is it not that sort of organisation?

issyl0: Good point for the future. "The future" here is because I've just regained a decent thought process for this after talking it through and having been away from it for a few days. The Contacts app is already on V1, but it is schemaless, so we should be documenting how it currently does things on V1 (as evidenced by this schema with an expanded "organisation" part), not how we want it to look in the next story which we've not got to yet - actually doing the work to make it V2. How can we invent schemas for things we haven't made yet and make everyone happy? Other apps don't have this problem as much 'cause they're going from **nothing `publishing-api`-related** to **V2**, AFAICT.

davidbasalla:
This is ready to be re-reviewed.

@jamiecobbett I ran the script with the format and the only error that came up was about `analytics_identifier` being `null` and not a `string`. This is a problem with content items that do not have an `analytics_identifier` set (the same error is thrown for `case_studies`). The ContentItemPresenter appears to set it to nil when that value is not set. Maybe it would be worth allowing `null` values for `analytics_identifier`?

formats/contact/publisher/details.json
issyl0: We have, however, added `organisation` to the `links` hash in the example and schema, but kept this way around as well so we have a current view on the state of the app.

PR #168, opened 2015-12-07
Financial releases content schemas

christophwong:
First attempt at exporting content out of 4 main lloyds press release pages in order to move content out of frontend app.


formats/financial_releases_index/publisher/details.json
```diff
@@ -0,0 +1,56 @@
...
+        "url": {
+          "type": "string",
+          "format": "uri"
+        },
+        "alt_text": {
+          "type": "string"
+        }
+      }
+    },
+    "press_releases": {
```
boffbowsh: These should be in the links hash rather than embedded within details, linking to the `financial_press_release` content items.

danielroseman:
Do we need separate schemas for the campaign page and the geoblocker? They look much the same.

mathildathompson:
@danielroseman thanks very much for taking the time to review our pull request. I have looked over campaign and geoblocker schemas - header_legal_disclaimer and email_caption are unique to campaign page. Do you not think it is worth keeping the schemas separate? Makes it easier to ascertain validity of example json against their own schema?

christophwong:
There has been a forced update to change some sample information.

formats/financial_releases_index/publisher/details.json
mathildathompson: @boffbowsh we have moved press releases into a links hash. Please could you review when you have time. We also have a couple of questions regarding referencing links inside html strings -would be great if we could chat to you about this briefly this morning if you have time? Thank you

danielroseman:
:+1: - the separate schemas are fine.

This can't currently be merged because it requires the CI tests to run, and they can't work on an external fork. You should have write access to this repo: please can you push to a branch and re-open the PR from there?

christophwong:
Sounds good, thanks @danielroseman

PR #169, opened 2015-12-09
Port combined_publisher_schemas to Rake

boffbowsh:
This PR ports the combination of Publisher V1 schemas to a Rake task from Make. For every make task, it shells out to `bundle exec bin/some_script`, which involves a lot of overhead in invoking the Ruby/Gems/Bundler environment for each file. This will get very unwieldy as the number of schema grows. 

For benchmarking purposes, processing `dist/formats/*/publisher/schema.json` takes ~8.5 seconds using Make, and around half a second using Rake. Most of that is the initial boot of the Ruby environment, I imagine even once we've got all the formats ported across an entire Rake run with all tasks will take around a couple of seconds.

If people prefer this approach, I can work on porting the other tasks.

For reference, total runtime of the Make suite at the moment from clean is 1.5 minutes


tijmenb:
:heart: for intent.

lib/tasks/combined_publisher_schemas.rake
```diff
@@ -0,0 +1,21 @@
...
+require 'json'
+
+READER = JSON::Schema::Reader.new(accept_file: true, accept_uri: false)
+
+rule %r{^dist/formats/.*/publisher/schema.json} => ->(f) {
+    Rake::FileList.new(
+      "formats/{definitions,metadata,v1_metadata,base_links}.json",
+      f.pathmap("%{^dist/,}p").pathmap("%d/{details,links}.json")
+    )
+  } do |t|
```
tijmenb: I think this could do with some explaining or variable extraction, I'm having a little bit of trouble grasping what is going on here :)

boffbowsh: Sure thing. I was going to refactor it if I got broad approval on approach.

tijmenb: :+1:

boffbowsh:
This is ready for review and merge. The second commit here is a possible refactoring. I am struggling to work out which form is more readable - opinions very welcome.

This PR now uses the Rake task in place of the main Make schema combination task, so make now takes 25 seconds less than before.

Part 1 of a few PRs to gradually port this Make suite.

tijmenb:
This is really cool. 

Do you think we could also take another approach? As far as I understand the build process, we loop through all formats, combine them if necessary and copy them to `/dist`. Could we avoid the `rule` matching by using `Dir.glob()` or something like that?

In pseudo-ruby:

```ruby
all_formats = Dir.glob("formats/*")

all_formats.each do |format_name|
  schema = generate_dist_schema(format_name)
  file.write(schema)
end

def generate_dist_schema(format_name)
  if File.exist?(format_name + "/schema.json")
    File.read(format_name + "/schema.json") # hand-made
  else
    SchemaCombiner.new({
      definitions: File.read("definitions.json"),
      metadata: File.read("metadata.json"),
      details: File.read("#{format_name}/details.json")
    }, format_name).combined
  end
end
```

boffbowsh:
I don't believe `Dir.glob` is compatible with Rake FileLists. We want to use Rake to get the same dependency resolution / partial build functionality as Make.

tijmenb:
> use dependency resolution / partial build functionality

I'm fairly ignorant when it comes to build systems. What is the reason we want to have this for this project?

boffbowsh:
So that if one of the dependency files changes, the related schema get rebuilt. If only e.g. `formats/case_study/publisher/details.json` changes, then we only need to rebuild `dist/formats/case_study/publisher/schema.json`.

tijmenb:
Ok, so that's what `rule` does somehow? 

I think that there's a trade-off between build speed and maintainability of the tools. I find the flow of the rake DSL really hard to follow. 

I suspect that the tasks performed here aren't that complex that we couldn't rebuild the entire thing in a few seconds, so maybe we can use pure Ruby?

benilovj:
While I think that this declarative style is fairly unusual within ruby frameworks, Rake is the appropriate tool for this task and this is using Rake in the correct way, too.

So :+1: from me to merge this, especially after @boffbowsh's change to make the method names less magic.

Makefile
```diff
@@ -28,7 +28,7 @@ publisher_v2_validation_records := $(publisher_v2_examples:formats/%.json=dist/f
 publisher_v2_links_validation_records := $(publisher_v2_links_examples:formats/%.json=dist/formats/%.json.publisher_v2_links.valid)
 
 # The various scripts used in the build process
-combiner_bin := bundle exec ./bin/combine_publisher_schema
```
jamiecobbett: Should this PR delete `bin/combine_publisher_schema`? The only references to it I can find now are in its spec and the README.

boffbowsh: Yeah, I'll need to repurpose the spec to test the Rake task instead I think.

boffbowsh:
Marking as DNM until I've added some sort of tests.

boffbowsh:
Removed old `bin` script and ported its spec to cover the new Rake task. Also added spec coverage for the v2 schema

PR #170, opened 2015-12-09
Financial releases content schemas

christophwong:
follow up on previously [forked pull request 168](https://github.com/alphagov/govuk-content-schemas/pull/168)

This is to port content out of email-campaign-frontend and to use content-schemas instead.


jamiecobbett:
I don't think the `.valid` files are supposed to be committed, see the [.gitignore](https://github.com/alphagov/govuk-content-schemas/blob/master/.gitignore)

formats/financial_releases_campaign/frontend/examples/financial_releases_campaign.json
```diff
@@ -0,0 +1,20 @@
...
+    "content": "<p>Vis nostro intellegebat at. Impedit pericula duo ne, pri nonumes ullamcorper in. Oratio invidunt et his, ele, at principes similique nam. No eam eleifend euripidis intellegat, sensibus theophrastus consectetuer.</p><p>Sumo solum quaestio has cu. Eu oblique euismod deleniti eum, congue necessitatibus ne sit.</p><p>No eum mutat soleat definiebas, ut argumentum efficiantur delicatissimi ius. At feugait adversarium vel. At cum mutat numquam graecis. Sit appareat adolescens constituto eu, te eleifend postulant theophrastus eam</p><p>Nulla regione quaeque ei pro, pro vulputate temporibus efficiantur id. Sit eu solum mediocrem, est te pertinax suavitate temporibus, et est incorrupte comprehensam.</p><p>Ne patrioque democritum mea, per natum doctus eu. Vel eu doming iudicabit. Al probo erant euripidis nec, blandit singulis euripidis an vel.</p><p>If you would like to receive email updates for daily ice cream flavours, send us a hand written note.</p><p>Vel ut iudico dolore. Vix ne diam iuvaret referrentur, eam nisl accusamus temporibus ne. If you like yoga <a id='lloyds-factsheet' href='/lloydsfactsheet' target='_blank'>click here</a></p>",
```
jamiecobbett: Looking at the other schemas, I think `body` would be a more consistent name for this, unless I've misunderstood?

formats/financial_releases_campaign/frontend/examples/financial_releases_campaign.json
```diff
@@ -0,0 +1,20 @@
...
+  "locale": "en",
+  "public_updated_at": "2015-12-07T11:22:21.000Z",
+  "format": "financial_releases_campaign",
+  "details": {
+    "image": {
+      "alt_text": "HM Government",
+      "url": "campaign.large.jpg"
+    },
+    "header_legal_disclaimer": "Header legal disclaimer Vel ut iudico dolore. Vix ne diam iuvaret referrentur, eam nisl accusamus temporibus ne. Nec quot deserunt dissentiet ut, ad  be available from <a href='/lloydsshares'>www.gov.uk/lloydsshares</a>.",
+    "title": "An Campaign Title",
```
jamiecobbett: Does this title really differ from `title` at the document root?

christophwong:
Sorry, that was a mistake, I will take those out now.

formats/financial_releases_campaign/frontend/examples/financial_releases_campaign.json
christophwong: Is it ok to use data outside of details and links (the meta data at document root) for frontend display? If yes then we should change it.

formats/financial_releases_campaign/frontend/examples/financial_releases_campaign.json
christophwong: Agree, I will change that.

formats/financial_releases_campaign/frontend/examples/financial_releases_campaign.json
jamiecobbett: Yes absolutely. The idea is that the root of the document is for properties that are common to more or less everything, `details` is for things which vary by format.

If you want an example of an app that renders from content-store, [government-frontend](https://github.com/alphagov/government-frontend) is (hopefully!) relatively easy to follow.

christophwong: Thank you for the pointer.

christophwong:
added a new commit which is a force-update to remove .valid files and to clean up the two commits regarding adding generated files.

formats/financial_releases_campaign/publisher/details.json
```diff
@@ -0,0 +1,39 @@
...
+         },
+         "alt_text": {
+           "type": "string"
+         }
+       }
+     },
+    "header_legal_disclaimer": {
+      "type": "string"
+    },
+    "title": {
```
jamiecobbett: I think this title should also be removed and the one at the document root used instead.

formats/financial_release/publisher/links.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "campaign_page"
+  ],
+  "properties": {
+    "campaign_page": {
```
jamiecobbett: Could this use the existing [`parent`](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/base_links.json#L17-L21) links entry?

formats/financial_releases_geoblocker/publisher/details.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "body"
+  ],
+  "properties": {
+    "title": {
```
jamiecobbett: I think this should also use the `title` at the root.

formats/financial_release/publisher/links.json
christophwong: @jamiecobbett the campaign page is not a parent of financial release page. We are not actually using the link, but hardcoding it into the page body; we thought to put it in a links hash for documentation. 
Do you suggest that we remove it if it is only needed in the body content?

formats/financial_releases_campaign/publisher/details.json
christophwong: Sorry, I must have missed it accidentally. There's now a new commit removing this and the other title that you've pointed out.

formats/financial_release/publisher/links.json
christophwong: we will go ahead with removing the campaign_page link for now, as it is not necessary at the moment.

formats/financial_release/frontend/examples/financial_release.json
```diff
@@ -0,0 +1,14 @@
+{
+  "base_path": "/lloyds-share-offer",
+  "content_id": "74314796-eee4-4a66-bc1e-3a2039f680e6",
+  "title": "Financial Release",
+  "format": "financial_release",
+  "locale": "en",
+  "public_updated_at": "2015-12-01T00:00:00.000Z",
+  "details": {
+    "summary": "Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,quis nostrud exercitation",
```
jamiecobbett: These `summary` fields seem like they should perhaps be in `description at the document root?

formats/financial_release/frontend/examples/financial_release.json
```diff
@@ -0,0 +1,14 @@
...
+    "published_date": "2015-10-05T00:00:00.000Z",
```
jamiecobbett: I wonder if this should be in `public_updated_at` at the document root?

christophwong: is public_updated_at more synonymous to "last_modified_time" or "first_published_time"?

formats/financial_release/frontend/examples/financial_release.json
christophwong: if we put summary at the document root, is there a way to validate examples with summary as a required field?
Also, why is it preferred to have it in the document root as opposed to inside of details?

formats/financial_release/frontend/examples/financial_release.json
tijmenb: I think `first_published_at` would be better in that case, as there's prior art for that:

https://github.com/alphagov/govuk-content-schemas/blob/4e1a2abc22afc65f3f1b887222103448294bdd79/dist/formats/specialist_document/publisher_v2/schema.json#L614-L617

christophwong: @tijmenb @jamiecobbett Thanks for pointing it out! It is now changed to `first_published_at`.

formats/financial_release/frontend/examples/financial_release.json
tijmenb: I think there's no way to make `description` required, but that's not intentional I think. Seems like a shortcoming in the build system. I'll take a look if we can make that possible.

I understand `description` was added to the root instead of the `details` hash because every content item should have one, and they are included in expanded links from the content-store.

formats/financial_releases_campaign/publisher/details.json
```diff
@@ -0,0 +1,36 @@
...
+         "url": {
+           "type": "string",
+           "format": "uri"
+         },
+         "alt_text": {
+           "type": "string"
+         }
+       }
+     },
+    "header_legal_disclaimer": {
```
tijmenb: It would be great if we could add small descriptions for these fields if they're not immediately obvious.

formats/financial_releases_index/frontend/examples/financial_releases_index.json
```diff
@@ -0,0 +1,35 @@
...
+      "alt_text": "lloyds-logo"
+    }
+  },
+  "links": {
+    "press_releases": [
+      {
+        "content_id": "57add221-0376-4f3e-a025-849954df97ac",
+        "title": "Press Release 1",
+        "base_path": "/lloyds-share-offer",
+        "api_url": "",
```
tijmenb: The `api_url` is always `https://www.gov.uk/api/content/base_path`. They also always have a description.

christophwong:
@tijmenb just actioned the last suggestion, if it all seems ok, I will rebase the commits to get the build to pass

formats/financial_releases_campaign/publisher/details.json
christophwong: it has been added by this [new commit](https://github.com/alphagov/govuk-content-schemas/commit/88f856a81eef711e28cce581c43a2cce8e853b98)

dist/formats/financial_release/frontend/schema.json
```diff
@@ -0,0 +1,214 @@
...
+        "body",
+        "first_published_at"
+      ],
+      "properties": {
+        "body": {
+          "type": "string"
+        },
+        "first_published_at": {
+          "type": "string",
+          "pattern": "^[1-9][0-9]{3}-(0[1-9]|1[0-2])-([012][0-9]|3[0-1])$"
```
jamiecobbett: Is there a reason this doesn't use the `date-time` format?

jamiecobbett:
This branch needs a rebase on top of master to be mergeable.

dist/formats/financial_release/frontend/schema.json
christophwong: In this instance, we actually only display the date and not the time. So I thought to follow with just the date format.

christophwong:
waiting on #177 to be merged so the Specialist frontend contract tests will pass

PR #171, opened 2015-12-09
Gitignore *.json.publisher_v2.valid files

issyl0:
- These seem to be a new addition. We couldn't work out why they were
  polluting our `git status` when we were generating new schemas and why
  we'd never seen anything like them before, then we found this file and
  determined they'd been missed when they were added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.json.valid
 *.json.frontend.valid
 *.json.publisher.valid
+*.json.publisher_v2.valid
+*.json.publisher_v2_links.valid


jamiecobbett:
I suspect its a result of https://github.com/alphagov/govuk-content-schemas/pull/164

jamiecobbett:
Seems good to merge to me, @danielroseman?

danielroseman:
:+1: but needs rebasing.

davidbasalla:
Rebased!

issyl0:
Thanks @davidbasalla for the rebase! :star:

PR #172, opened 2015-12-09
Add new taxon schema

mobaig:
We are introducing taxon content items to start building a new taxonomy
for GOVUK. This will help us to start tagging content items to specific
parts of the taxonomy.

### DIFF
--- a/formats/taxon/publisher/details.json
+++ b/formats/taxon/publisher/details.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "internal_name": {
+      "type": "string",
+      "description": "An internal name for admin interfaces. Includes parent."
+    }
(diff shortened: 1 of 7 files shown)


formats/taxon/publisher/links.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "children": {
```
tijmenb: We won't be using `children` probably.

formats/taxon/publisher/details.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false
```
tijmenb: Perhaps `internal_name` would be useful here too https://github.com/alphagov/govuk-content-schemas/pull/156

PR #175, opened 2015-12-11
Copy handmade schemas to publisher_v2 dist...

steventux:
Handmade schemas for redirects, gones and special_routes are 'handmade'
and not combined from component schemas so make an exceptional copy of
these to publisher v2 build directories.


tijmenb:
Awesome. Test failures here are unrelated btw (https://github.com/alphagov/contacts-frontend/pull/38). I'll merge as soon as that's fixed.

issyl0:
Sorry for the test failures!

PR #136, opened 2015-10-16
Add employment tribunal decisions schema and examples

robmckinnon:
The Employment Tribunal publishes decisions that describe the outcome of a tribunal case.

Tribunal decision finders are used by public, professionals and the tribunals themselves to search for past decisions.

Published decisions form a part of case law and are used for research purposes and to prepare for current or future cases.

PR #177, opened 2015-12-14
Change tribunal_decision_categories to match schema in examples

robmckinnon:
Fixes error caused by non-matching `tribunal_decision_categories` values in employment tribunal decisions examples.


robmckinnon:
This is a fix for error in tests reported here: https://github.com/alphagov/govuk-content-schemas/pull/136#issuecomment-164439288

tijmenb:
:+1: thanks for the quick fix.

PR #180, opened 2016-01-05
Allow additional properties in ‘placeholder’ formats…

tuzz:
This eases the migration by allowing publishing
apps to send all required data to the publishing
api, without actually making the switch to the
new format.


tuzz:
We're going to do this differently.

PR #181, opened 2016-01-06
Replace example validation make logic with rake

benilovj:
This change makes validation quicker and simpler:

- having to spin up an instance of ruby for each example file was the most expensive
part of the example validation process. Having this whole process in rake means
validation is an order of magnitude faster (< 10 sec with the current number of
examples)
- because validation is so much faster, we can drop the somewhat
involved dependency resolution logic and just always validate everything.

/cc @boffbowsh


benilovj:
I believe that the test failures in the Whitehall contract tests are to do with the broken Whitehall build, not this change.

PR #182, opened 2016-01-07
Remove stray links.json file in /dist

benilovj:
This `links.json` file for `financial_release` doesn't have a source in `/formats` and is missing
when `/dist` is regenerated. Maybe it was inadvertently checked in?
It also means that currently the `financial_release` schemas are inconsistent.

/cc @jamiecobbett @boffbowsh @christophwong @mathildathompson 

If this file is supposed to be there, an alternative change would be to check in the source
and to regenerate the schemas.


mathildathompson:
Thanks for spotting this @benilovj - it must have been  inadvertently checked in. Will merge now.

PR #184, opened 2016-01-07
More rigorous `clean` step

benilovj:
The idea is that everything in `dist/formats` should be generated.
To make sure that this is indeed the case, the `clean` step should
delete everything in that folder and the build should successfully
regenerate it.

/cc @boffbowsh


benilovj:
This build now passes and should be ready for a review/merge.

PR #187, opened 2016-01-07
Build v2 links.json for all schemas

boffbowsh:
If a schema doesn’t have a customised links.json, it should still use
the base, so we should generate it anyway.


jamiecobbett:
Seems right to me, but I suggested it. Anyone else have an opinion?

jamiecobbett:
Do we know when this bug was introduced? Have we always had it?

steventux:
The bug was likely introduced [when the source json files were carved up to accommodate v2 publisher examples](https://github.com/alphagov/govuk-content-schemas/pull/164/files). I refactored the links into separate files as v2 publishing api endpoints are split between content and links. I evidently didn't catch this case so thank you for catching this and fixing :+1:

PR #188, opened 2016-01-07
Travel advice govspeak

tuzz:
It's easier to just look at this commit: https://github.com/alphagov/govuk-content-schemas/commit/bd238599c60a77e4f0bc9dd5b4b3ef8cab8c991d (the other just rebuilds the schemas)

This change is made more complicated by the fact
that the schema differs between backend and
frontend.

By the time the content reaches the front-end, the
‘text/html’ content type is inlined. Therefore, we
need to generate slightly different schemas to
cater for this.

https://trello.com/c/L6gBHNFN/471-make-travel-advice-publisher-send-both-govspeak-and-html-to-the-publishing-api

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -312,6 +312,25 @@
         }
       }
     },
+    "multiple_content_types": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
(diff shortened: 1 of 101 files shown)


boffbowsh:
Looks good to me, I assume the build failure is known seeing as it's about travel-advice-publisher anyway :+1:

tuzz:
Yes, the build is dependent on the application PR merge.

PR #190, opened 2016-01-07
Ensure output files are newline terminated

benlovell:
Ensures outputted files are newline terminated as per the POSIX
standard.


jamiecobbett:
:leftwards_arrow_with_hook:

PR #195, opened 2016-01-08
Remove can_be_replaced_by

jamiecobbett:
I believe this was never actually used - @elliotcm?

In any case, the solution to this problem is that now certain formats
(coming_soon, redirect, gone, unpublishing) are allowed to overwrite other
formats using a different content-id. These formats are also overwritable by
any other format.

Aside: you might think that the Whitehall tests which check that eg coming_soon or unpublishings are valid against these schemas should have failed. They don't, because although this field was required in their links hash, the links hash itself isn't required and they don't supply it.


tuzz:
:+1:

PR #196, opened 2016-01-13
Add updated_at to travel_advice_index countries

tuzz:
Depends on https://github.com/alphagov/travel-advice-publisher/pull/96 being merged.


danielroseman:
:+1: but needs rebase.

PR #198, opened 2016-01-13
Add analytics specific data

benlovell:
Adds analytics specific data to enable integration style testing for the
`govuk_component` project's `analytics_meta_tags` component.


gpeng:
Looks good. Will you just test against one document or do they all need an `analytics_identifier`?

benlovell:
For the time being I'm localising this here as I have my doubts about the general approach. I'll take it up with the team tomorrow, but for now this needs to be merging before I can configure triggered builds of alphagov/static.

PR #200, opened 2016-01-14
Split campaign_page body html fragment into smaller fields

mathildathompson:
Complex html is difficult for the content team to edit.
  Break down the body field into multiple smaller fields to avoid this.


boffbowsh:
Currently, other publishing apps are responsible for rendering their "nice" parts of bodies into one body field, which the frontend then renders. Perhaps a better pattern here would be for the publishing script to take the various fields and concatenate them or otherwise formulate them into one larger body field?

mathildathompson:
Makes sense thanks for the feedback @boffbowsh

jamiecobbett:
>Currently, other publishing apps are responsible for rendering their "nice" parts of bodies into one body field, which the frontend then renders. Perhaps a better pattern here would be for the publishing script to take the various fields and concatenate them or otherwise formulate them into one larger body field? 

If you're building a new app, though, you won't be storing content items in the app, so won't have the ability to separate and combine fields like that.

PR #203, opened 2016-01-18
Topic enhancements to support Service Manual Topics

tadast:
After several discussions with the Finding Things team we've decided to use the existing Topics format for Service Manual Topics ([see example](http://service-manual-staging.herokuapp.com/agile-delivery/)). To make it work we need the following:

1. topic groups need an optional `description` field to describe each group ("sub-section"?)
2. reference linked items within groups using `content_id`s
3. be able to "expand" `content_id`s so that a URL and a title for the referenced content items could be displayed on a frontend page.

To implement 3 we have decided to *temporarily* use the `links -> linked_items` hash. Content ids in the links hash are automatically expanded by the content-store. This can later be used to match the expanded content items to match with their references in `groups -> content_ids`. This is expected to be replaced by a dependency resolution service that is currently being spiked.

@rboulton could you please have a look and let me know if you have any questions?

There are related WIP branches: 
1) on service-manual-publisher: to generate payload in this format, [see here](https://github.com/alphagov/service-manual-publisher/pull/130)
2) on government-frontend: to display Service Manual Topics that is using this format: [see here](https://github.com/alphagov/government-frontend/compare/service_manual_topics?expand=1).

### DIFF
--- a/formats/topic/publisher/links.json
+++ b/formats/topic/publisher/links.json
@@ -6,6 +6,10 @@
     "children": {
       "description": "Any child topics",
       "$ref": "#/definitions/guid_list"
+    },
+    "linked_items": {
+      "description": "Includes all content ids referenced in 'details'. This is a temporary measure to expand content ids for frontends which is planned to be replaced by a dependency resolution service.",
+      "$ref": "#/definitions/guid_list"
     }
   }
 }
--- a/dist/formats/topic/frontend/schema.json
+++ b/dist/formats/topic/frontend/schema.json
@@ -115,10 +115,10 @@
         "children": {
           "$ref": "#/definitions/frontend_links"
         },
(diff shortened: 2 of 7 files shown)


formats/topic/publisher/links.json
```diff
@@ -6,6 +6,10 @@
     "children": {
       "description": "Any child topics",
       "$ref": "#/definitions/guid_list"
+    },
+    "linked_items": {
+      "description": "Includes all content ids referenced in 'details'. This is a temporary measure to expand content ids for frontends which is planned to be replaced by a dependency resolution service.",
```
rboulton: Thank you for adding this description.

rboulton:
This seems like a reasonable approach, for the reasons you state.

Thanks for a well written PR.

PR #206, opened 2016-01-19
Statistics announcement schema and examples

benlovell:
Adds the schemata and front end examples for Statistics Announcements.

Trello: https://trello.com/c/grRZvVD9

### DIFF
--- a/formats/statistics_announcement/publisher/details.json
+++ b/formats/statistics_announcement/publisher/details.json
@@ -5,7 +5,7 @@
   "required": [
     "display_date",
     "state",
-    "sub_type"
+    "format_sub_type"
   ],
   "properties": {
     "cancellation_reason": {
@@ -32,7 +32,7 @@
         "provisional"
       ]
     },
-    "sub_type": {
+    "format_sub_type": {
       "type": "string",
       "enum": [
         "national",
--- a/formats/statistics_announcement/publisher/details.json
+++ b/formats/statistics_announcement/publisher/details.json
@@ -1,0 +1,43 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "display_date",
+    "state",
+    "sub_type"
+  ],
+  "properties": {
+    "cancellation_reason": {
+      "type": "string"
+    },
+    "cancelled_at": {
+      "type": "string",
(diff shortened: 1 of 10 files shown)


boffbowsh:
This is missing the ability to distinguish between Official and National statistics, which is needed for the display type thing.

benlovell:
@boffbowsh Good catch.

boffbowsh:
:+1: when rebased

formats/statistics_announcement/publisher/details.json
```diff
@@ -0,0 +1,38 @@
...
+    "latest_change_note": {
+      "type": "string"
+    },
+    "logo": {
+      "$ref": "#/definitions/image"
+    },
+    "previous_display_date": {
+      "type": "string"
+    },
+    "release_date": {
```
dsingleton: Whats the reason for this (and `previous_display_date`) being a string, rather than a `datetime` type?

benlovell: The `*_at` type dates by convention seem to be actual dates, the rest are strings.

dsingleton: Just had a chat offline and agree'd we might want to keep the convention used in WH Publisher of `display_date`, which gives some inference of it not being a proper datefield. 

Though we're both a bit ¯\_(ツ)_/¯ on the semantics. Any other opinions?

dsingleton:
LGTM after the changes.

The only thing i'm not 100% on is `sub_type` as a field name, it's a pattern we're going to be using in quite a few places, and this'll set some precedent, so it might be worth the nitpicking (sorry).

dsingleton:
Adding something more constructive.

Formats we know have sub types of:
- [Publication](https://github.com/alphagov/whitehall/blob/master/app/models/publication_type.rb#L94-L111) (Policy Paper, Guidance, Forms, Research, Maps, etc)
- [News articles](https://github.com/alphagov/whitehall/blob/master/app/models/news_article_type.rb#L54-L56) (News Story, Press Release, Response, etc)
- Consultation (Open Consultation, Closed Consultation, Consultation Outcome)

Whitehall tends to use `type` and `sub_type`, but we trying to model formats rather than WH types, eg we have `Statistic`s as their own format, but in WH they're both sub-types of `Publication`. Maybe something like `sub_format`?

PR #208, opened 2016-01-20
Add update_type to V2 metadata schema

tommyp:
As Applications move to V2 they will start to the use the Publishing API
as a data store. When they do this, in order to persist `update_type`
between draftings, it will need to be stored as part of the metadata.
This commit adds it to the V2 metadata but doesn't make it required, as
the `update_type` can also be provided when calling the `publish`
command on the content item.

### DIFF
--- a/dist/formats/case_study/publisher_v2/schema.json
+++ b/dist/formats/case_study/publisher_v2/schema.json
@@ -96,6 +96,13 @@
     "previous_version": {
       "type": "string"
     },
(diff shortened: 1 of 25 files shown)


boffbowsh:
This needs adding to the handcrafted schemas too https://github.com/alphagov/govuk-content-schemas/commit/2bf12e9c3721021c5bf8c6a232539834be68521e

PR #209, opened 2016-01-20
Contacts: Put `organisation` in the links hash, not the details hash

issyl0:
- This was being sent by contacts-admin in both places prior to
  https://trello.com/c/HEB3ukuP/263-remove-organisations-from-details-hash-for-contacts.
  We only need it in the one place - the links hash - as that's the
  convention now. It works as it is - contacts currently uses
  organisations from the links hash, but we should tidy up here.

The build is failing on contacts-admin because of https://github.com/alphagov/contacts-admin/pull/224 not being on master yet. Once that's merged and the build here is rerun, this _should_ go green.

Edit: That PR is merged, so this has indeed gone green. :green_heart:

### DIFF
--- a/dist/formats/contact/frontend/schema.json
+++ b/dist/formats/contact/frontend/schema.json
@@ -77,50 +77,6 @@
         "description": {
           "type": "string"
         },
-        "organisation": {
-          "type": "object",
-          "additionalProperties": false,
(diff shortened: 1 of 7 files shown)


davidbasalla:
Looks good :+1:

PR #210, opened 2016-01-20
Remove organisations from details hash in hmrc_manual and hmrc_manual_sections

mobaig:
hmrc-manuals-api is being updated to publish organisation data in the
links hash rather than the details hash, as per current convention for
documents across GOVUK. This change drops details>organisations as a
required attribute and updates the hmrc_manual_section frontend example.


mobaig:
Closing while I sort out hmrc-manuals-api.

PR #211, opened 2016-01-21
Remove contact groups from contacts

davidbasalla:
Remove contact groups from the contacts schema as it is currently not used in `contacts-frontend` so therefore irrelevant in the content item. It stores organisation data from the `contacts-admin` app (imported from `whitehall`) which we want to move away from, using organisations from the `content-store` instead.

The tests will pass once [this PR](https://github.com/alphagov/contacts-admin/pull/226) has been merged.


jamiecobbett:
:+1: merge when green. If this was needed now, we would implement it with the links hash.

PR #213, opened 2016-01-22
Adds lead organisations link type

MatMoore:
Currently users can explicitly order the organisations in
policy-publisher when publishing the policy, but this will change
once policy-puplisher is migrated to use publishing api for storing
link data.

We think that all users need to be able to do is distinguish between
lead and supporting organisations, in a similar manner to whitehall.

The lead_organisations field here is intended to overlap with the
organisations one, so that other uses of organisations are not
impacted by the change.

This relates to https://trello.com/c/BQDQ824W/468-use-publishing-api-v2-for-policy-publisher

### DIFF
--- a/formats/base_links.json
+++ b/formats/base_links.json
@@ -16,6 +16,11 @@
       "$ref": "#/definitions/guid_list"
     },
     "organisations": {
+      "description": "All organisations linked to this content item. This should include lead organisations.",
+      "$ref": "#/definitions/guid_list"
+    },
+    "lead_organisations": {
+      "description": "A subset of organisations that should be emphasised in relation to this content item. All organisations specified here should also be part of the organisations array.",
       "$ref": "#/definitions/guid_list"
     },
     "parent": {
--- a/dist/formats/case_study/publisher/schema.json
+++ b/dist/formats/case_study/publisher/schema.json
@@ -216,6 +216,7 @@
       ],
       "properties": {
         "lead_organisations": {
+          "description": "A subset of organisations that should be emphasised in relation to this content item. All organisations specified here should also be part of the organisations array.",
           "$ref": "#/definitions/guid_list"
         },
         "related_policies": {
@@ -249,6 +250,7 @@
           "$ref": "#/definitions/guid_list"
         },
         "organisations": {
+          "description": "All organisations linked to this content item. This should include lead organisations.",
           "$ref": "#/definitions/guid_list"
         },
         "parent": {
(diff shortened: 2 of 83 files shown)


benlovell:
It may be helpful to add this to a frontend example too.

formats/base_links.json
```diff
@@ -18,6 +18,9 @@
     "organisations": {
       "$ref": "#/definitions/guid_list"
     },
+    "lead_organisations": {
```
rboulton: Could you add a "description" field to this object, and also to the "organisations" one above?  These descriptions are just documentation for us, but are very useful to ensure that link types have a consistent meaning.

In particular, the `organisations` description should mention that this is all organisations associated with the content, duplicating the entries which are put in the `lead_organisations` field.  And the `lead_organisations` description should mention that all entries added to this field must also be added to the "organisations" field.

MatMoore: Yeah that definitely needs to be documented. Will add it in today.

MatMoore:
@benlovell @rboulton Regarding the frontend examples, could you have a look at the example we've just added and see if it makes sense?

I'm a bit unsure of a couple of things because the frontend_links type includes a lot of information that will be duplicated for the lead organisations, and won't be particularly useful to the frontend. Should we rework this so that it is just a content ID?

```
  "links": {
      "type": "object",
      "properties": {
        "related": {
          "$ref": "#/definitions/frontend_links"
        },
        "email_alert_signup": {
          "$ref": "#/definitions/frontend_links"
        },
        "available_translations": {
          "$ref": "#/definitions/frontend_links"
        },
        "alpha_taxons": {
          "$ref": "#/definitions/frontend_links"
        },
        "mainstream_browse_pages": {
          "$ref": "#/definitions/frontend_links"
        },
        "topics": {
          "$ref": "#/definitions/frontend_links"
        },
        "organisations": {
          "$ref": "#/definitions/frontend_links"
        },
        "lead_organisations": {
          "$ref": "#/definitions/frontend_links"
        },
        "parent": {
          "$ref": "#/definitions/frontend_links"
        },
        "policies": {
          "$ref": "#/definitions/frontend_links"
        },
        "policy_areas": {
          "$ref": "#/definitions/frontend_links"
        }
      }
```

To provide a bit more context, I was imagining that the frontend would consume this by primarily referring to the organisations object, and using the presence of an entry in lead_organisations purely for sorting/formatting. In the simple case where there is only one organisation, there are two possible ways of representing it (with/without the lead orgs). Which one the frontend sees is determined by the publishing app.

bradwright:
@MatMoore the intent is for the hash to provide everything the app needs to render it. If the lead orgs entry just points at the orgs entry, the presenters will need to unpack that at display time. So while your imagined use case might be more efficient, it's a trade off we haven't previously made in our data formats.

rboulton:
I think this is all good.  Merging.

PR #214, opened 2016-01-22
Rename `StatisticAnnouncement#policy_areas` to `topics`

gpeng:
This preserves consistency as statistic announcements relate to topics
rather than policy areas.


benlovell:
It seems this may well be incorrect given the renaming that occurred was actually the inverse i.e. topic to policy area. According to alphagov/whitehall#2240 (although this didn't get merged but referencing here for intent), alphagov/whitehall#2249 and alphagov/whitehall#2349.

If not, then colour me confused. :see_no_evil:

PR #219, opened 2016-01-26
Service Manual content owners

tadast:
Now that we have community pages to reference, we can replace the "hardcoded" `content_owner` attribute in Service manual guides with a standard content_id type link that will get expanded and will carry an up-to-date value of title and base path in it.

We'll need to re-PUT all existing SM guide drafts to the publishing-api (there are no published items at the moment) and remove content owner field from the details hash in this repo once it's done.

I've made the details content owner an optional field. Once migrated we can make the content_owners field in links mandatory if it makes sense.

Also, the frontend application will need to be updated to prefer the links field and fall back
to the one in details (https://github.com/alphagov/government-frontend/pull/78)

Do the same for the topics: for Service Manual topics we need to render a sidebar linking to the community pages that are responsible for the guides listed in the topic. Add a similar `content_owners` link to topics as well (/cc @rboulton)

### DIFF
--- a/formats/service_manual_guide/frontend/examples/basic_with_related_discussions.json
+++ b/formats/service_manual_guide/frontend/examples/basic_with_related_discussions.json
@@ -17,6 +17,28 @@
       { "title" : "What is it, why it works and how to do it", "href" : "#what-is-it-why-it-works-and-how-to-do-it" }
     ]
   },
+  "links": {
+    "topics": [{
+      "content_id": "f08d0fbb-c14e-4c08-8260-a24c9eacceb4",
+      "title": "Agile delivery",
+      "base_path": "/service-manual/agile-delivery",
+      "description": "Agile delivery topic description",
+      "api_url": "http://content-store.dev.gov.uk/content/service-manual/agile-delivery",
+      "web_url": "http://www.dev.gov.uk/service-manual/agile-delivery",
+      "locale": "en",
+      "analytics_identifier": null
+    }],
+    "content_owners": [{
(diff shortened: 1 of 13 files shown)


formats/service_manual_guide/frontend/examples/basic_with_related_discussions.json
```diff
@@ -17,6 +17,28 @@
       { "title" : "What is it, why it works and how to do it", "href" : "#what-is-it-why-it-works-and-how-to-do-it" }
     ]
   },
+  "links": {
+    "topics": [{
+      "content_id": "f08d0fbb-c14e-4c08-8260-a24c9eacceb4",
+      "title": "Agile delivery",
+      "base_path": "/service-manual/agile-delivery",
```
rboulton: This should probably be `"/topic/service-manual/agile-delivery"`, I think (we planned that the service-manual topic would live under /topic like all the other topics).  Similarly elsewhere in this PR.

tadast: I've checked with Tom and he says that you've agreed with publishing topics under `/service-manual` initially (at the bottom of [Tom's notes](https://docs.google.com/document/d/1-O_Zt_k7cjXm-ZTzMA12xDPQu2G5qzFCT8YIPaYzmAk/edit)). I think, among other reasons, we can't currently publish under `/topics` because the IA team has not allocated any topic to Service manuals. 

Tom also mentioned that you spoke about creating redirects when we migrate to collections publisher. 

Come over and talk to Tom if my answer is not very clear, I'm sure he can explain it better :)

PR #220, opened 2016-01-28
Add topical event about page schema and examples

binaryberry:
Adds the schema and front end examples for Topical Event About pages.

[Trello](https://trello.com/c/C5dHT8Zr/265-3-topical-event-about-page-content-schema-and-example-medium)


formats/topical_event_about_page/publisher/links.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "parent"
+  ],
+  "properties": {
+    "parent": {
+      "$ref": "#/definitions/guid_list"
```
boffbowsh: You can add a `maxItems` property here to enforce that there is only one parent http://spacetelescope.github.io/understanding-json-schema/reference/array.html#length

formats/topical_event_about_page/frontend/examples/topical-event-about-page-archived.json
```diff
@@ -0,0 +1,27 @@
+{
+  "content_id": "6f2cb109-5117-4e6d-9c82-df30c9fd3791",
+  "base_path": "/government/topical-events/london-conference-on-afghanistan-2014/about",
+  "description": "The London Conference on Afghanistan will take place on 4 December 2014, co-hosted by the governments of the UK and Afghanistan. Associated events will take place on the afternoon of Wednesday 3 December. The conference will provide a platform for the government of Afghanistan to set out its vision for reform and for the international community to demonstrate enduring solidarity and support for Afghanistan.",
+  "details": {
+    "name": "About the London Conference on Afghanistan",
```
jamiecobbett: I think this should be the `title` in the root of the document.

formats/topical_event_about_page/publisher/links.json
jamiecobbett: In fact, `parent` is already defined in `base_links.json` with `"maxItems": 1`, so we don't need to include it here. I'll rebase shortly.

jamiecobbett:
Rebased to address existing comments.

benlovell:
Rebased again.

fofr:
Can `formats/topical_event_about_page/frontend/examples/topical-event-about-page.json` be renamed to `topical_event_about_page.json` for consistency with other examples.

benlovell:
@fofr Yup, will address that (after fixing the date-dependent test failures in static.) :see_no_evil:

fofr:
The example body contains many double spaces where there should be single spaces, eg `The virus  is only`.

When rendering the example these spaces get removed which makes integration tests like the following fail: `assert_has_component_govspeak(@content_item["details"]["body"])`

I think the spaces should be removed.

fofr:
There should also be a "short text" example, so that a frontend without any contents can be tested.
eg https://www.gov.uk/government/topical-events/battle-of-the-somme-centenary-commemorations/about
vs https://www.gov.uk/government/topical-events/daesh/about

fofr:
:+1:

PR #221, opened 2016-01-29
Validate schemas explicitly

jamiecobbett:
I made a mistake whilst modifying a schema. It was combined successfully and
all the tests passed. but it was only when I added an example that the schema
was validated and I realised the problem. The error was not descriptive and it took time for me to realise the problem was actually in the schema, not the example. Let's shorten the feedback loop and
make it an explicit step.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,7 @@ begin
 rescue LoadError
 end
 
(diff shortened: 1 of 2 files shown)


gpeng:
Looks good to me... Just needs a rebase.

benlovell:
Look ma' I did a rebase!

PR #223, opened 2016-01-29
Update policy links schema to reflect live behaviour

MatMoore:
The related links array is not always returned by content store if there are no links of this type.

See https://github.com/alphagov/finder-frontend/pull/244

On that PR, @tommyp suggested adding an example to the schemas so we can have a regression test for this bug.

This is not actually possible without changing the schema itself, but I think it makes sense to have this reflect reality. Also, rather than add a new one I thought it would be easier to change the "minimal" example.

Once merged I'll need to raise another PR against finder frontend to extend the specs for this example.

### DIFF
--- a/dist/formats/policy/frontend/schema.json
+++ b/dist/formats/policy/frontend/schema.json
@@ -184,8 +184,7 @@
       "type": "object",
       "additionalProperties": false,
       "required": [
-        "organisations",
-        "related"
+        "organisations"
       ],
       "properties": {
         "people": {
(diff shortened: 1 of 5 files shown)


rboulton:
I think that the content store should never be returning empty arrays for the links hash.  In particular, as currently modelled, in the publishing-api there is no difference between an empty list of links of a particular type and an absent list of links for a particular link type - there simply isn't a way in the database to represent an empty array (because all we store is the relations).

I think these empty arrays are therefore never going to come from the content store for anything which has been updated in content-store since we changed the modelling of links to be like this in publishing-api.

MatMoore:
@rboulton Should this work be extended then? Might be good to discuss in person when you have a moment.

Davidslv:
@MatMoore this branch is not mergeable.

PR #227, opened 2016-02-02
Add expanded frontend links

steventux:
https://trello.com/c/YUaMfA4y/524-add-breadcrumbs-and-related-links-components-to-multipage-frontend-travel-advice-page renders 'expanded' links suchas breadcrumbs and related links for the travel advice format.
This PR makes the frontend schemas handle nested parent links in the links hash and provides an additional travel advice example.

/cc @fofr 

This PR is necessary for integration tests here https://github.com/alphagov/multipage-frontend/pull/17 to pass.


fofr:
Thanks for adding these @steventux

PR #228, opened 2016-02-04
Add publisher v2 examples for email-campaign

steventux:
Adds publisher v2 examples for email-campaign formats.
Part of [this story](https://trello.com/c/QOevhBRK/114-114-refactor-content-store-specs)

### DIFF
--- a/formats/financial_release/publisher_v2/examples/financial_release.json
+++ b/formats/financial_release/publisher_v2/examples/financial_release.json
@@ -1,0 +1,26 @@
+{
+  "base_path": "/lloyds-share-offer",
+  "title": "Financial Release",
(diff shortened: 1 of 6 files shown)


jamiecobbett:
Looks good. I can understand why you've used dummy content, but is the dummy text of a realistic length? I ask this because one of the use cases for the frontend examples is to enable design work to be done against the [dummy-content-store](https://github.com/alphagov/govuk-content-schemas/blob/master/docs/running-frontend-against-examples.md).

steventux:
@jamiecobbett, thanks for looking at this. These are publisher v2 examples, am I right in thinking these don't get returned by the dummy-content-store? If so perhaps not a big issue with regards to rendering. I agree the frontend examples will need modifying as a couple of the legal disclaimers are short compared to real content.

jamiecobbett:
Yes - you're right, merged.

PR #230, opened 2016-02-05
Expose ruby version via Gemfile

dsingleton:
Not all environments will pick up a `.ruby-version` file when deciding
which version of ruby to use. Heroku uses a Gemfile's `ruby x.y.z`

This repo is deployed to Heroku as an instance of `dummy_content_store`
serving it's examples [1].

In order for this app to be deployed to Heroku at the right ruby version
the version is exposed via the Gemfile as well, but not duplicated, so
there isn't a risk of the versions being set independently and falling
out of sync.

When a version isn't set, Heroku defaults to `2.0.0`, which it is sun
setting [2], so a Heroku-recognised version needs to be exposed for
the instance of the app to continue working.

This also makes the ruby version specific to the patch level, not just
minor. The difference in version is important when running as an
application, eg, the dummy content store.

Although this repo isn't an application intended to be deployed, it's
Heroku instance is a supported workflow [3] and it's often run locally
during development, when not working against a live content store, and
in both these cases we should be pinned to a specific version of ruby.

Exposing the `.ruby-version` via the `Gemfile` _is_ a little ugly, but
it's a pattern we've used before on apps we want to run in heroku [4]
and as our apps become simpler, and easier to run independantly of the
rest of our stack, this is a pattern we may want to follow on more of
our apps.

[1]: https://govuk-content-store-examples.herokuapp.com
[2]: https://devcenter.heroku.com/articles/ruby-support#ruby-versions
[3]: https://github.com/alphagov/govuk-content-schemas/blob/master/docs/running-frontend-against-examples.md
[4]: https://github.com/alphagov/government-frontend/commit/98c05c01198133c8afd88a3ad4bb0a9cd06d9ba0#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R3

**cc*/ @benlovell and @boffbowsh, who I discussed this with on Slack.


boffbowsh:
:+1: but not merging to allow for discussion

benlovell:
I'm :+1: too.

erikse:
I'd be :+1: for removing .ruby-version or .gitignoring them, and just having the value the Gemfile. Heroku doesn't read it, and recent RVM can read the version out of the Gemfile anyway
```
Eriks-MacBook-Pro:alphagov erik$ ruby -v
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]

Eriks-MacBook-Pro:alphagov erik$ cd smart-answers/
RVM used your Gemfile for selecting Ruby, it is all fine - Heroku does that too,
you can ignore these warnings with 'rvm rvmrc warning ignore /Users/erik/Coding/alphagov/smart-answers/Gemfile'.
To ignore the warning for all files run 'rvm rvmrc warning ignore allGemfiles'.

Eriks-MacBook-Pro:smart-answers erik$ ruby -v
ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin15]

Eriks-MacBook-Pro:smart-answers erik$ ls .ruby-version
ls: .ruby-version: No such file or directory
```

boffbowsh:
Yes but rbenv doesn't support that without a plugin.

benlovell:
@erikse Some of us use other ruby version managers.

dsingleton:
> I'd be  for removing .ruby-version or .gitignoring them, and just having the value the Gemfile. Heroku doesn't read it, and recent RVM can read the version out of the Gemfile anyway

As I understand it we use `.ruby-version` for compatibility with `rbenv`, which doesn't seem to support the `Gemfile` approach [without a plugin](https://github.com/rbenv/rbenv/issues/727).

erikse:
Just seems backwards to me, making our code depend on some tooling config

bradwright:
@erikse it's a standard pattern in 12factor apps to include all necessary dependencies for the target platform in the code repository: http://12factor.net/dependencies. Our target platform uses rbenv to manage Ruby versions.

boffbowsh:
But that tooling is codified within our deployment processes and the puppet configuration of both the dev VMs and (non-contractor) developer laptops.

tuzz:
:+1: from me

dsingleton:
OK. Sounds like `.ruby-version` isn't going to be going away any time soon, and that this is a valid workaround for apps which need to be deployed to Heroku :+1: 

Any reason I shouldn't open a PR to add this workaround to https://github.com/alphagov/govuk-rails-app-template as a default? That repo has the advantage that it can be a somewhat canonical example of good, and the commit there can mirror the information here.

boffbowsh:
Sounds sensible to me :+1:

benlovell:
Makes sense and helps with the Appeal to a Higher Power :tm: situations in further related PRs.

PR #231, opened 2016-02-05
Allow placeholder start/end dates to be null

jamiecobbett:
It turns out that Topical Events don't have to have dates.

### DIFF
--- a/dist/formats/placeholder/frontend/schema.json
+++ b/dist/formats/placeholder/frontend/schema.json
@@ -74,14 +74,28 @@
           ]
         },
         "start_date": {
-          "type": "string",
-          "format": "date-time",
-          "description": "Used for topical events, so that related documents can get the date. Remove when topical events are migrated."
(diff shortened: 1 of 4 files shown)


boffbowsh:
Rather than this, could we not just make the presenter not send the key at all if it's null?

jamiecobbett:
@boffbowsh it could. It changes the presenter logic from:
```ruby
      details = {
        start_date: item.start_date.try(:to_datetime),
        end_date: item.end_date.try(:to_datetime),
      }
      super.merge(details: details)
```
to:
```ruby
      details = {}
      details[:start_date] = item.start_date.to_datetime if item.end_date
      details[:end_date] = item.end_date.to_datetime if item.start_date
      super.merge(details: details)
```

Opinions?

boffbowsh:
Or this https://github.com/alphagov/whitehall/pull/2456#discussion_r52010767

PR #232, opened 2016-02-05
Allow additional properties in frontend links

fofr:
To render topical event about pages the frontend needs  some extra information from the topical event - its start and end date.

Rather than including the specific details and start/end dates in the properties, we are allowing additonalProperties. Per format link definitions are hard to define at the moment, for instance start/end date may have different meaning based on format.

This is a short term solution in lieu of more programatically generated frontend links schemas. When the new pub api dependency work is completed the schemas will be used to drive what fields are included so it'll be able to be embedded correctly in the frontend schema.

Unblocks https://trello.com/c/P4KCL1UJ/

Additional changes need to be made to content-store to do the expansion so that the content-items returned look like the included examples in this PR.

cc @boffbowsh @jamiecobbett


boffbowsh:
This approach was discussed and agreed with @danielroseman

PR #233, opened 2016-02-10
Include alert_status in travel advice example

fofr:
This is a feature in the travel advice format that FCO publishers don’t appear to be using at the moment. However they could begin using it and it would render a block on the front-end, so an appropriate example should be included.

cc @dsingleton


dsingleton:
:+1: to having examples covering possible cases.

It's a bit weird that the field is an array, rather than a single value, and the value is free text, not an enum, to avoid issues like typos, as we do on some [other formats](https://github.com/alphagov/govuk-content-schemas/blob/a73006a08e6a05e20e06e014ba641555bdba7988/formats/statistics_announcement/publisher/details.json). But thats all well out of scope of this change, so merging.

PR #236, opened 2016-02-11
Add parent to the service manual topic example.

thehenster:
We are going to use the links->parent array to generate the breadcrumb for a topic in government-frontend for the service manual.

The content_id is taken from the special_route set up for https://www.gov.uk/api/content/service-manual.

### DIFF
--- a/formats/topic/frontend/examples/service_manual_topic.json
+++ b/formats/topic/frontend/examples/service_manual_topic.json
@@ -52,6 +52,14 @@
         "web_url":"http://www.dev.gov.uk/service-manual/test-topic",
         "locale":"en"
       }
+    ],
+    "parent": [
+      {
+        "content_id": "51dd8775-cd2a-4fb3-b6df-8ed03591122d",
+        "title": "Service manual",
+        "base_path": "/service-manual",
+        "locale": "en"
+      }
     ]
   },
   "description":"Service Manual Topic description",


danielroseman:
I'd say that the right thing to do is to publish a service-manual page. There is a "special route publisher" method in gds-api-adapters which you can use to register a route with a content_id at deploy time - see for example how it's done in [publisher](https://github.com/alphagov/static/blob/678947e8df38970de824434f07b55b1530635c24/lib/tasks/publishing_api.rake).

AndrewVos:
@danielroseman Thanks.

The thing is, we're going to actually avoid publishing the root page until later. For now we're going live (as in publishing, not production (we're already in production)) with topic pages.

Maybe we could use the old /service-manual/ page, but that doesn't seem consistent to me?

thehenster:
I wonder if perhaps we should just remove the `content_id` from the example. We won't be sending one from the publisher for quite some time.

boffbowsh:
That wouldn't be valid - the [content_id is required](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/topic/frontend/schema.json#L11)

thehenster:
This is a `content_id` in links->parent. Which [looks like it's also required](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/topic/frontend/schema.json#L288) :) Blast.

thehenster:
I've updated the content_id from the special_route as per https://www.gov.uk/api/content/service-manual.

Thank you all for the tips :bow:

thehenster:
Change of plan. We're going to create a service manual specific topic instead so this file will get moved/deleted.

PR #238, opened 2016-02-15
Service manual topic format

thehenster:
The new service manual is going to get its own topic format. Technically it is similar to 'topic'. The goal is to help keep it conceptually grouped together with the other service manual formats and separate from the non-service manual formats in government-frontend. Once the formats have settled and found their rightful frontend we hope to be able to delete this service_manual_topic format.

This schema was based on topic but with the following notable changes:

 * details->internal_name removed from the schema
 * details->beta removed from the schema


dist/formats/service_manual_topic/frontend/schema.json
```diff
@@ -0,0 +1,323 @@
...
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/guid"
+      }
+    },
+    "image": {
```
binaryberry: Here you could use the [image definition](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json) as we did [here](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/take_part/publisher/details.json#L13)

dist/formats/service_manual_topic/publisher_v2/links.json
```diff
@@ -0,0 +1,180 @@
...
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/guid"
+      }
+    },
+    "image": {
```
binaryberry: Same comment as above here - the image definition should be used.

dist/formats/service_manual_topic/publisher_v2/schema.json
```diff
@@ -0,0 +1,269 @@
...
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/guid"
+      }
+    },
+    "image": {
```
binaryberry: And same thing here - using the image definition would be best

dist/formats/service_manual_topic/publisher/schema.json
```diff
@@ -0,0 +1,313 @@
...
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+    },
+    "guid_list": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/guid"
+      }
+    },
+    "image": {
```
binaryberry: And here.

dist/formats/service_manual_topic/frontend/schema.json
thehenster: I'm very likely to be wrong (I'm new around here - hi!)... But the definition above is being pulled in during the compilation process? We don't actually use "image".

binaryberry: Hello! And welcome, btw! :sunny:  I hadn't noticed this was in the dist, apologies; it's normal it is expanded here. Not obvious to me why they get compiled though - it is not part of metadata.json, nor your links nor details.

AndrewVos:
lgtm

PR #242, opened 2016-02-17
HTML publications schema and examples

benlovell:
Trello: https://trello.com/c/Wcl1hWJ2

Adds both the schema and related examples to highlight a first published and an updated HTML publication.

### DIFF
--- a/dist/formats/html_publication/frontend/schema.json
+++ b/dist/formats/html_publication/frontend/schema.json
@@ -1,0 +1,311 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "title",
+    "format",
+    "locale",
+    "public_updated_at",
+    "content_id"
+  ],
+  "properties": {
+    "base_path": {
+      "$ref": "#/definitions/absolute_path"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "anyOf": [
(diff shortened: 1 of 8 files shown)


fofr:
Summarising an offline discussing with @benlovell.

* HTML attachments have bespoke logic around manually numbered headings, we should include an example of that.
* The manually numbered flag might need to be included in the content item if the frontend is generating the contents list (which it's not at the moment) …

__Headers and contents lists__
* The `headers` part of the content item, used for displaying a contents list, is a new way of handling contents lists.
  * Topical event about pages generate them from the body HTML in government-frontend
  * Specialist content also generates them, but instead uses a headers hierarchy, see: https://www.gov.uk/api/content/cma-cases/ladbrokes-coral-group-merger-inquiry
  * HTML attachments as-is will provide a rendered contents list
* The logic to generate the headers HTML from the govspeak markdown in Whitehall was originally moved to a worker because the generation process was too slow. This might be a warning against doing the contents list generation in the frontend (ie the topical event about page approach)

We're going to begin the frontend development working against this branch and the examples, and see what problems we come across.

fofr:
I didn't think this was ready to be merged? Still some important outstanding questions.

benlovell:
@fofr Yeah. It was a little premature. We can pair up on the front end bits tomorrow and see what falls out of it.

gpeng:
Yes... Sorry about that. I had it open and Paul's comments weren't on the page until after I merged it. I'll revert if you'd prefer?

PR #244, opened 2016-02-17
Image from asset manager

ssatish:
Image field on campaign should be editable. We are using asset manager to store the images uploaded. And the asset url must be captured in the content store. This PR is to change govuk-content-schemas to reflect this change.


formats/financial_releases_campaign/publisher/details.json
```diff
@@ -25,5 +25,28 @@
     "footer_legal_disclaimer": {
       "$ref" : "#/definitions/multiple_content_types"
     }
+  },
+  "definitions": {
+    "asset_link": {
```
boffbowsh: These should go in the base `definitions.json` file

jmeter.log
```diff
@@ -0,0 +1,36 @@
+2015/12/23 14:20:35 INFO  - jmeter.util.JMeterUtils: Setting Locale to en_US
```
boffbowsh: You've left a log file in here

dist/formats/financial_releases_campaign/publisher_v2/schema.json
```diff
@@ -258,6 +258,27 @@
...
+        "url",
+        "content_type",
+        "alt_text"
+      ],
+      "properties": {
+        "url": {
+          "type": "string",
+          "format": "uri"
+        },
+        "content_type": {
```
boffbowsh: Why is the content type needed?

formats/financial_releases_campaign/publisher/details.json
ssatish: @boffbowsh What is the file path for this base definitions.json file ?

dist/formats/financial_releases_campaign/publisher_v2/schema.json
ssatish: @boffbowsh This format will have a pdf document uploaded to it too. Very soon. Does that mean we will need to have content type then ?  
If not, I can remove it. I am not really using the content type in the app.

ssatish: I have removed content type

formats/financial_releases_campaign/publisher/details.json
boffbowsh: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json

boffbowsh:
This needs a rebase

ssatish:
@boffbowsh Rebased.

boffbowsh:
You still need to address my comment from previously. Can you also please add some more context to this? Either to the PR or the commit messages?

ssatish:
@boffbowsh I don't really understand your comment regarding asset_link to go into definitions file.

boffbowsh:
You've created a new definition in that file. You'll see that the `image` type is defined in the base `definitions.json` file, which is also where the `asset_link` type should be defined so that other formats can share it.

boffbowsh:
Based on the PR description, it seems like this will only store images, is this correct? If so, what was wrong with the original `image` type?

ssatish:
Maybe I don't need this PR. Because the whole upload functionality does seem to work without this change merged.

ssatish:
I will close the PR.

PR #245, opened 2016-02-19
Add document_type and schema_name to top level

danielroseman:
These are alternatives to format; it is an error to supply both sets. To
enable this, the top-level properties has been replaced with a 'oneOf'
matcher with two choices, each of which includes all the existing fields plus
either format or the two new fields. This is because JSON Schema does not
support extending a schema with alternative options, only replacing it.

To reduce duplication in the generated schemas, the links and details that
would have been included in the top level fields have been moved to the
definitions hash with only a reference included in the two options.

It is now also possible to specify the allowed values for document_type, by adding a separate document_types.json schema in the publisher directory. If this is not present, the allowed value is not constrained.


boffbowsh:
Can you explain the rationale behind this approach rather than using the existing `format` field for the proposed `schema_name`, and a first class `document_type` field? What advantages does this have? I'm concerned about unknown code that might rely on `format` being present, and complexity for developers switching between different publishing and frontend apps needing to remember which style is used where.

boffbowsh:
For completeness, just noticed this is related to https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+41%3A+Separate+document+type+from+format. 

mysterysolved

danielroseman:
@elliotcm restored newlines and rebased.

PR #246, opened 2016-02-19
Rename policy_group to working_group

jamiecobbett:
Whitehall sends these to publishing-api as placeholder_working_group, and
policies link to them in an entry called "working_groups". The Whitehall commit
which sets the format used for publishing-api has more detail:
https://github.com/alphagov/whitehall/commit/0bf9ce56fa4a9f82552e7eb551a1cc3c95b7ec39


boffbowsh:
Good spot. Sorry for missing this in previous review.

jamiecobbett:
@boffbowsh no worries, it is very confusing!

PR #253, opened 2016-02-26
    Add public_updated_at as an optional property

AndrewVos:
Adds to:

- redirect
- gone
- unpublishing
- coming_soon

This allows us to publish content, when that content has not been
published before, but has an "update_type" of "major".

### DIFF
--- a/dist/formats/coming_soon/frontend/schema.json
+++ b/dist/formats/coming_soon/frontend/schema.json
@@ -181,6 +181,10 @@
         "publish_time": {
           "type": "string",
           "format": "date-time"
+        },
+        "public_updated_at": {
+          "type": "string",
+          "format": "date-time"
         }
       }
     },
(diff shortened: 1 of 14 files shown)


tuzz:
I think this may only be the case if there hasn't previously been a public_updated_at set via the put_content action. You can provide it either at update time or publish time, but one of those must be provided.

Either way, publishing api shouldn't be responding with a 500. We'll discuss this in the team next week.

AndrewVos:
Thanks @tuzz. For now we've just used "major" (https://github.com/alphagov/service-manual-publisher/pull/170) temporarily. Though I'm not sure about why we original decided to even go with "minor". If publishing-api should accept the first put_content with "minor" then this pull request should be closed?

boffbowsh:
This needs rebasing

tijmenb:
:+1:

PR #254, opened 2016-03-01
Fix HTML for working_group example

jamiecobbett:
There was some malformed HTML, so I grabbed a fresh copy. The old HTML included
markup from a Pinterest browser extension (data-pin attributes), which might
partly explain why the HTML was broken.

/cc @dsingleton


fofr:
Why does the new one have lots of line breaks, whereas the old one didn't? Lots of `\r\n` throughout.

jamiecobbett:
I don't know. I think the old one might have been manually JSON-encoded (eg swapping `"` for `'`). Maybe @binaryberry remembers?

binaryberry:
Hi! I did have to do some single and double quotes swapping when copy/pasting the HTML, which I did from the console. Worth noting I had also shortened the content compared to the original page so that it wouldn't be quite as long, hence the length difference.

jamiecobbett:
This is blocking me finishing the frontend work, is someone happy to merge it?

binaryberry:
:tada:

jamiecobbett:
Thanks! :fireworks:

PR #257, opened 2016-03-02
Add an optional subscriber_list_document_type…

tuzz:
See https://trello.com/c/gPjk5P2A/611-add-document-type-only-query-to-email-alert-api


elliotcm:
:+1: when rebased and green

PR #259, opened 2016-03-08
[WIP] Remove `update_type`

tijmenb:
The `update_type` field is sent to the publishing-api on publish, not in the main payload.


tijmenb:
Closing this, because it is still valid to send a `update_type` in the content item. This value will be used if you `publish` a doc without an `update_type`.

PR #261, opened 2016-03-08
Update schemas to use the new subscriber_list details format

tuzz:
**We will merge the pull requests ourselves in the Publishing Platform team when everyone's happy.**

https://trello.com/c/oqGCU2XP/525-add-email-signup-links-to-travel-advice-on-frontend

We've been migrating travel advice over to use the new email-alert-frontend workflow rather than the old pagewatch workflow.

Whilst doing this work, we noticed that policy publisher passes through subscriber list information as implicit links on the signup content item (which should really be reserved for more general associations with other content items). After talking with @tijmenb and @mobaig in Finding Things we concluded that the details hash is the best place for that information.

Therefore, this work consists of four parts:

1) An update to travel advice publisher to create the new email alert signup forms
https://github.com/alphagov/travel-advice-publisher/pull/123

2) An update to policy publisher to change the structure of the subscriber list data
https://github.com/alphagov/policy-publisher/pull/129

3) An update the email-alert-frontend to support searching by document_type (required for travel index)
https://github.com/alphagov/email-alert-frontend/pull/25

4) An update to the content schemas to reflect the new subscriber list data structure
https://github.com/alphagov/govuk-content-schemas/pull/261

We'd like to merge and deploy all four pull requests at (roughly) the same time. Please could comment as usual and state whether you are happy for the pull requests to be merged.

### DIFF
--- a/dist/formats/email_alert_signup/frontend/schema.json
+++ b/dist/formats/email_alert_signup/frontend/schema.json
@@ -175,13 +175,43 @@
       "type": "object",
       "additionalProperties": false,
       "required": [
-        "signup_tags",
+        "subscriber_list",
         "summary"
       ],
       "properties": {
-        "subscriber_list_document_type": {
-          "type": "string",
-          "description": "The document_type to pass through to the email alert api when creating a subscriber list"
+        "subscriber_list": {
+          "type": "object",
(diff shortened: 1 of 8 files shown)


steventux:
:+1:

PR #262, opened 2016-03-09
Change placeholder schema for organisations

tvararu:
Modify the placeholder schema to accommodate the branding / logo information on the organisation model. Also adds an example.

https://trello.com/c/cJGrtnc1

### DIFF
--- a/formats/placeholder/frontend/examples/organisation.json
+++ b/formats/placeholder/frontend/examples/organisation.json
@@ -1,0 +1,36 @@
+{
+  "base_path": "/government/organisations/bona-vacantia",
+  "content_id": "c9109663-0128-4db4-b513-a9cdfb691034",
+  "title": "Bona Vacantia",
+  "format": "placeholder",
+  "document_type": "organisation",
+  "schema_name": "placeholder",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "2016-03-09T15:42:28.867Z",
+  "public_updated_at": "2015-03-31T18:33:12.000+00:00",
+  "phase": "live",
+  "analytics_identifier": "OT564",
+  "links": {
+    "available_translations": [
+      {
+        "content_id": "c9109663-0128-4db4-b513-a9cdfb691034",
+        "title": "Bona Vacantia",
+        "base_path": "/government/organisations/bona-vacantia",
+        "description": null,
+        "api_url": "https://www.gov.uk/api/content/government/organisations/bona-vacantia",
+        "web_url": "https://www.gov.uk/government/organisations/bona-vacantia",
+        "locale": "en",
+        "analytics_identifier": "OT564",
+        "links": {}
+      }
+    ]
+  },
+  "details": {
+    "brand": "attorney-generals-office",
+    "logo": {
+      "formatted_title": "Bona Vacantia",
+      "crest": "single-identity"
+    }
+  }
+}
--- a/dist/formats/placeholder/frontend/schema.json
+++ b/dist/formats/placeholder/frontend/schema.json
@@ -192,6 +192,37 @@
           "format": "date-time",
           "description": "Used for topical events, so that related documents can get the date. Remove when topical events are migrated."
         },
+        "brand": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "description": "Used for organisations, to allow us to publish branding / logo information. Remove when organisations are migrated."
+        },
+        "logo": {
+          "formatted_title": {
+            "type": "string",
+            "description": "Used for organisations, to allow us to publish branding / logo information. Remove when organisations are migrated."
+          },
+          "crest": {
+            "type": "string",
+            "enum": [
+              "bis",
+              "custom",
+              "eo",
+              "hmrc",
+              "ho",
+              "mod",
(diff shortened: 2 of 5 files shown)


formats/organisation/publisher/links.json
```diff
@@ -0,0 +1,7 @@
+{
```
boffbowsh: This file isn't required. If it's not provided, the schema will be generated with an empty links hash.

formats/organisation/frontend/examples/bona_vacantia.json
```diff
@@ -0,0 +1,36 @@
...
+  "phase": "live",
+  "analytics_identifier": "OT564",
+  "links": {
+    "available_translations": [
+      {
+        "content_id": "c9109663-0128-4db4-b513-a9cdfb691034",
+        "title": "Bona Vacantia",
+        "base_path": "/government/organisations/bona-vacantia",
+        "description": null,
+        "api_url": "https://www-origin.integration.publishing.service.gov.uk/api/content/government/organisations/bona-vacantia",
```
boffbowsh: These URLs should be for the public site, so `www.gov.uk`

formats/organisation/frontend/examples/bona_vacantia.json
```diff
@@ -0,0 +1,36 @@
+{
+  "base_path": "/government/organisations/bona-vacantia",
+  "content_id": "c9109663-0128-4db4-b513-a9cdfb691034",
+  "title": "Bona Vacantia",
+  "format": "organisation",
+  "document_type": "organisation",
```
boffbowsh: For reference, as discussed in person, this should be `placeholder_organisation`

formats/organisation/frontend/examples/bona_vacantia.json
tvararu: :+1:

tvararu:
@boffbowsh Updated!

tijmenb:
Sorry if I jump in here, but would having a content-item with `document_type: "organisation"` and `schema_name: "placeholder"` make sense? What's the way forward with these names?

tvararu:
@tijmenb valid concern! After discussing with @boffbowsh we've decided to not go in this direction anymore. We'll be using the `placeholder` format and just adding an `organisation.json` example.

formats/placeholder/frontend/examples/organisation.json
```diff
@@ -0,0 +1,36 @@
+{
+  "base_path": "/government/organisations/bona-vacantia",
+  "content_id": "c9109663-0128-4db4-b513-a9cdfb691034",
+  "title": "Bona Vacantia",
+  "format": "placeholder_organisation",
+  "document_type": "organisation",
+  "schema_name": "placeholder",
```
tvararu: @tijmenb @danielroseman: Any thoughts about using `"format": "placeholder_organisation"`?

boffbowsh: To clarify this, I think the `format` is ok, but I don't have a huge amount of context yet on the `document_type`/`schema_name` split. Does this look ok to you @danielroseman?

tijmenb: When you send something to publishing-api you can specify `format` OR `document_type` & `schema_name`.

In the last case, `format` will be set to `schema_name`. So the example could read:

```
+  "format": "placeholder_organisation",
+  "document_type": "organisation",
+  "schema_name": "placeholder_organisation",
```

or

```
+  "format": "placeholder",
+  "document_type": "organisation",
+  "schema_name": "placeholder",
```

tijmenb:
@tvararu @boffbowsh 

If organisations will get extra fields ([here](https://github.com/alphagov/whitehall/blob/organisation-logos/app/presenters/publishing_api_presenters/organisation.rb#L6-L14)), shouldn't it get its own schema? I believe `"additionalProperties": true` will make `placeholder` into a kind of wildcard schema that allows everything and we lose the benefit of having content-schemas.

boffbowsh:
@tijmenb this still needs to be a `placeholder` as it doesn't contain enough information to render the page, but other formats will need it. Placeholders will gradually disappear, and this pattern will need to happen with more and more of them as we work through migration. Creating many smaller placeholder formats will end up quite unwieldy.

I think this situation might get addressed anyway as part of the dependency resolution work - it'll be necessary to refactor some of the content schemas, maybe allowing some form of inheritance.

boffbowsh:
Had a chat with @tijmenb offline, what we've decided is that for now we'll carry on using the placeholder format for this, but add the fields explicitly to the schema and make them not required. We can then set `additionalProperties` back to false. There's prior art for this, see the `start_date` and `end_date` in the placeholder which is used for Topical Events.

tijmenb:
:+1:

formats/placeholder/frontend/examples/organisation.json
boffbowsh: This sounds sensible to me.

tvararu:
I think this should be sorted out now. :raised_hands:

boffbowsh:
This needs rebasing

tvararu:
:green_apple:

PR #263, opened 2016-03-10
Add "pdf" field in campaign publishing model

christophwong:
We are working to allow editors to attach a pdf file (just need the url)
to a Campaign. It uses the definition of an image as it is an object
with just a required field of "url".


formats/financial_releases_campaign/publisher/details.json
```diff
@@ -24,6 +24,8 @@
     },
     "footer_legal_disclaimer": {
       "$ref" : "#/definitions/multiple_content_types"
-    }
-  }
+    },
+    "pdf": {
+      "$ref": "#/definitions/image"
+    }}
```
tuzz: Please could we move the second closing brace onto a new line?

dist/formats/financial_releases_campaign/frontend/schema.json
```diff
@@ -196,6 +196,9 @@
         },
         "footer_legal_disclaimer": {
           "type": "string"
+        },
+        "pdf": {
+          "$ref": "#/definitions/image"
```
steventux: Not sure this is the correct definition.
https://github.com/alphagov/govuk-content-schemas/blob/03e412bbbc5cdd22a2a6c235b51df2518bbcfc56/formats/travel_advice/publisher/details.json#L80 defines an `asset_link`, it might make more sense to move this definition into `formats/definitions.json` and re-use it.

PR #268, opened 2016-03-15
Remove available translation block from HTML pub example

benilovj:
Whitehall Publisher doesn't track the relationships between an HTML
publication and its translation, and there isn't any navigation
for users to travel between these translations.

/cc @benlovell


gpeng:
:+1:

PR #273, opened 2016-03-29
Add optional max_cache_time to travel_advice(_index)…

tuzz:
We’d like to specify a smaller cache time for
Travel Advice content. This change introduces a
max_cache_time field that can optionally be set.
There will be some subsequent work in Content
Store to respect this field.

We can extend this out to other formats if needed
in the future.


danielroseman:
Needs a rebase.

tuzz:
@danielroseman rebased

PR #274, opened 2016-03-29
Add change history to service manual guide

AndrewVos:
https://trello.com/c/qzHbv9S9/181-show-change-notes-and-page-history-at-the-bottom-of-guidance-page

Dependent on:

https://github.com/alphagov/government-frontend/pull/129
https://github.com/alphagov/service-manual-publisher/pull/209
https://github.com/alphagov/govuk-content-schemas/pull/274
Add change history to service manual guide
    
Extract `change_history` into definitions, and use this reference
inside `case_study/publisher/schema.json` and
`special_document/publisher/schema.json`

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -227,24 +227,7 @@
           ]
         },
         "change_history": {
-          "type": "array",
-          "items": {
-            "type": "object",
-            "additionalProperties": false,
-            "properties": {
-              "public_timestamp": {
-                "type": "string",
-                "format": "date-time"
-              },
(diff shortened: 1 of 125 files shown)


danielroseman:
There's already a pattern for this, used by specialist documents and case studies: see https://github.com/alphagov/govuk-content-schemas/blob/12ac2580368a0279a98abaa972e006d177d9f31e/formats/specialist_document/publisher/details.json#L28

AndrewVos:
@danielroseman Thanks. I also noticed they omit the "update_type" field, which in retrospect isn't needed because all updates listed in there will be "major".

PR #277, opened 2016-04-01
Add content_owners to service-manual-topic format

thehenster:
The goal is to list the communities related to the guides included in a topic in the sidebar of a service manual topic page. We are calling them content_owners because that is the pattern set out by service-manual-guide.

https://trello.com/c/rZlsL4Am/152-add-communities-sidebar-to-the-topic-pages


AndrewVos:
lgtm

PR #278, opened 2016-04-04
Add govspeak content_type to body in manual, manual_section & specialist-doc

mathildathompson:
- Currently only storing html in content-store, need to store both html and govspeak;
- Body attribute for manual, manual_section and specialist_document now stores both gov_speak and html;
- I think this is the only attribute that requires this functionality?

[Trello Card](https://trello.com/c/3TVLuEPC/14-import-all-the-old-specialist-publisher-documents-into-the-new-version-of-specialist-publisher)


benilovj:
:+1:

benilovj:
LGTM. Merging.

PR #279, opened 2016-04-05
Add `detailed_guide` definition and example.

tvararu:
This defines the detailed guide links and details.

Only one basic example is defined so far. Suggestions (or better yet, `slug`s) are welcome with examples of other guides which feature particular edge cases that we have not covered.

Relevant Trello ticket:
https://trello.com/c/HyUyYqDu/314-6-detailed-guides-migration-content-schema-and-example-large

cc @binaryberry @tijmenb @mgrassotti

### DIFF
--- a/formats/detailed_guide/frontend/examples/detailed_guide.json
+++ b/formats/detailed_guide/frontend/examples/detailed_guide.json
@@ -1,0 +1,96 @@
+{
+  "content_id": "c9e77115-22aa-45a2-8c0d-827d92462758",
+  "base_path": "/guidance/salary-sacrifice-and-the-effects-on-paye",
+  "description": "Find out how salary sacrifice arrangements work and how they might affect an employee's current and future income.",
+  "public_updated_at": "2016-02-18T15:45:44.000+00:00",
+  "title": "Salary sacrifice",
+  "updated_at": "2016-02-18T10:37:00Z",
+  "schema_name": "detailed_guide",
+  "document_type": "detailed_guide",
+  "format": "detailed_guide",
+  "locale": "en",
+  "details": {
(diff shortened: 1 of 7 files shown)


formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,35 @@
+{
```
tijmenb: I think you can remove `primary_specialist_sectors` and `additional_specialist_sectors` here, and remove a couple that are provided by `base_links.json`.

tvararu:
@tijmenb: Updated, removed all the properties that aren't being used by the example.

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,56 @@
...
+      "$ref": "#/definitions/image"
+    },
+    "scheduled_publication": {
+      "type": "string"
+    },
+    "alternative_attachment_email": {
+      "type": "string"
+    },
+    "related_mainstream_url": {
+      "type": "string"
```
boffbowsh: This should be in the `links` hash as a link to another content item. Same for the additional related item.

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,14 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "topics",
+    "lead_organisations"
+  ],
+  "properties": {
+    "worldwide_organisations": {
```
binaryberry: We still need the related_guides in here

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,56 @@
...
+    },
+    "image": {
+      "$ref": "#/definitions/image"
+    },
+    "scheduled_publication": {
+      "type": "string"
+    },
+    "alternative_attachment_email": {
+      "type": "string"
```
boffbowsh: This should be in the links hash as a link to an organisation, referred to as an `alternative_format_provider_organisation`. We will then expand the `alternative_format_email` from that organisation by default.

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,56 @@
...
+    },
+    "excluded_nations": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "alternative_scotland_url": {
+      "type": "string"
```
boffbowsh: These should be links to related content items in the links hash.

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,56 @@
...
+  "required": [
+    "body",
+    "first_published_at"
+  ],
+  "properties": {
+    "body": {
+      "type": "string"
+    },
+    "first_published_at": {
+      "type": "string"
```
boffbowsh: This should be a `date-time`

binaryberry: The tests fail when we try to set it to date-time - it expects a string.

tijmenb: Did you `.iso_8601` the date?

formats/detailed_guide/publisher/details.json
tvararu: I don't know how that looks like. What does this need instead of `"type": "string"`?

```
    "related_mainstream_url": {
      "type": "string"
    },
    "additional_related_mainstream_url": {
      "type": "string"
    },
    "alternative_scotland_url": {
      "type": "string"
    },
    "alternative_wales_url": {
      "type": "string"
    },
    "alternative_nothern_ireland_url": {
      "type": "string"
    },
```

formats/detailed_guide/publisher/details.json
tvararu: Sorted:

```
    "first_published_at": {
      "type": "string",
      "format": "date-time"
    },
```

tvararu:
All comments addressed, to the best of my knowledge. :smiley:

formats/detailed_guide/frontend/examples/detailed_guide.json
```diff
@@ -0,0 +1,71 @@
...
+    "change_note": "The second entry in the table Examples of salary sacrifice has been amended to correct the explanation of how much of the salary is subject to tax and National Insurance contributions.",
```
fofr: The live example, https://www.gov.uk/guidance/salary-sacrifice-and-the-effects-on-paye shows a complete history of changes. We should represent that here.

formats/detailed_guide/frontend/examples/detailed_guide.json
```diff
@@ -0,0 +1,71 @@
...
+        "base_path": "/government/topics/tax-and-revenue",
+        "api_url": "https://www.gov.uk/api/content/government/world/topics/tax-and-revenue",
+        "web_url": "https://www.gov.uk/government/world/topics/tax-and-revenue",
+        "locale": "en"
+      }
+    ],
+    "parent": [
+      {
+        "content_id": "6e1d83bc-d37a-447c-bdac-31be4b441417",
+        "title": "PAYE",
```
fofr: I expected the parent to be "Business tax", which is what the title context currently links to on the live page.
screen shot 2016-04-05 at 17 34 28

formats/detailed_guide/frontend/examples/detailed_guide.json
```diff
@@ -0,0 +1,71 @@
...
+        "base_path": "/topic/business-tax/paye",
+        "api_url": "https://www.gov.uk/api/content/topic/business-tax/paye",
+        "web_url": "https://www.gov.uk/topic/business-tax/paye",
+        "locale": "en"
+      }
+    ],
+    "topics": [
+      {
+        "content_id": "6e1d83bc-d37a-447c-bdac-31be4b441417",
```
fofr: Should this include both PAYE and Business tax as topics?

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: Me too! Here's the spelunking story:

```ruby
# whitehall rails c
> dg = Document.find_by_slug("salary-sacrifice-and-the-effects-on-paye").latest_edition
> dg.primary_specialist_sectors.first # only one
<SpecialistSector id: 47465, edition_id: 595175, tag: "business-tax/paye", created_at: "2016-02-18 09:02:19", updated_at: "2016-02-18 09:02:19", primary: true>
```

@tijmenb pointed out that `primary_specialist_sectors` map to `parent`:

img_0561

And he also demonstrated how to find the above object in the `content-store`:

```ruby
# content-store rails c
> ci = ContentItem.find_by(base_path: "/topic/business-tax/paye"
> ci.title
"PAYE"
# etc
```

formats/detailed_guide/frontend/examples/detailed_guide.json
```diff
@@ -0,0 +1,71 @@
...
+        "api_url": "https://www.gov.uk/api/organisations/hm-revenue-customs",
+        "web_url": "https://www.gov.uk/government/organisations/hm-revenue-customs",
+        "locale": "en",
+        "analytics_identifier": "D25"
+      }
+    ],
+    "worldwide_organisations": [
+      {
+        "content_id": "6667cce2-e809-4e21-ae09-cb0bdc1ddda3",
+        "title": "HM Revenue & Customs",
```
fofr: This isn't an accurate worldwide organisation. It leads to the example rendering in the frontend like this:
screen shot 2016-04-05 at 17 46 38
screen shot 2016-04-05 at 17 46 35

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: The model in the Whitehall database only links to this one, so that's why there's only one here. I agree it doesn't look correct, but I don't know what the correct solution is.

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: Awesome, thanks for that, I'll update it.

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: Me and @binaryberry can't figure out why we actually put this in, as we don't need it. :flushed:

tvararu:
Comments addressed, again.

formats/detailed_guide/frontend/examples/detailed_guide.json
tijmenb: Currently things can only be tagged to "second level" topics ("subtopics").

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,32 @@
...
+    "lead_organisations"
+  ],
+  "properties": {
+    "related_guides": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "alternative_format_provider_organisation": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "related_mainstream_url": {
```
boffbowsh: These shouldn't be named as "url", these are pointers to pieces of content, whose urls can change.

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,32 @@
...
+      "$ref": "#/definitions/guid_list"
+    },
+    "additional_related_mainstream_url": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "alternative_scotland_url": {
```
boffbowsh: After discussion with Tatiana earlier, these might be external URLs which won't be in the content store. Has this been confirmed one way or the other?

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "first_published_at": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "image": {
+      "$ref": "#/definitions/image"
+    },
+    "scheduled_publication": {
```
boffbowsh: What does this do?

formats/detailed_guide/frontend/examples/detailed_guide.json
tijmenb: This is a good point - it's showing the title of top-level-topic there.

I believe Whitehall can do this because content-api exposes the parent for any tag, so it displays that title of the `primary_specialist_sector`'s parent ("Business Tax") of the title of the actual `primary_specialist_sector` ("PAYE").

The eventual idea is to expand `links.parent` recursively with the dependency resolution, so that content-store has the full data to render this page.

cc @mobaig, who is converting the frontend of Whitehall to use the links hash (and thus parent).

Perhaps a chat tomorrow would be good?

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,39 @@
...
+      "type": "string"
+    },
+    "excluded_nations": {
+      "type": "array",
+      "items": {
+        "type": "string"
```
boffbowsh: This should be an enum:
```json
"enum": ["england", "scotland", "wales", "northern_ireland"]
```

formats/detailed_guide/frontend/examples/detailed_guide.json
boffbowsh: Business tax is the parent of PAYE. If PAYE is added, dependency resolution will (eventually) render the parent as well for the frontend. I'm not entirely sure how this will look in the content store, @tuzz will hopefully be able to point you at an example.

tuzz: Something like:

```javascript
{
  "expanded_links": {
    "parent": [
      {
        "content_id": "111-111",
        "title": "VAT Rates",
        "base_path": "/tax/vat-rates",
        "parent": [
          {
            "content_id": "222-222",
            "title": "Taxation",
            "base_path": "/tax",
            "parent": [
              {
                "content_id": "333-333",
                "title": "Home",
                "base_path": "/"
              }
            ]
          }
        ]
      }
    ]
  }
}
```

More information here:
https://github.com/alphagov/publishing-api/blob/master/doc/dependency-resolution.md

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,32 @@
...
+    "alternative_format_provider_organisation": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "related_mainstream": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "additional_related_mainstream": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "alternative_scotland_url": {
```
tvararu: Re-adding comment from @boffbowsh that GitHub marked as outdated but isn't:

> After discussion with Tatiana earlier, these might be external URLs which won't be in the content store. Has this been confirmed one way or the other?

tvararu:
Marked as WIP until parallel work on backend and frontend for initial example is implemented.

formats/detailed_guide/publisher/links.json
tvararu: Spelunking (yes, I do like this word and I will use it a lot) reveals they are external indeed URLs, and as such should be part of `details` and not `links`.

E.g., https://www.gov.uk/guidance/storing-oil-at-a-home-or-business has excluded nations:

```ruby
> d = Document.find_by_slug("storing-oil-at-a-home-or-business")
> dg = d.latest_edition
> dg.nation_inapplicabilities.last
<NationInapplicability id: 122149, nation_id: 4, edition_id: 574326, created_at: "2015-02-25 17:22:38", updated_at: "2015-12-16 17:13:00", alternative_url: "https://www.doeni.gov.uk/publications/control-poll...">
```

The `alternative_url` is fully qualified, so it's external.

cc @boffbowsh

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "topics",
```
binaryberry: These should be required in details.json instead, as that's where topics now live.

formats/detailed_guide/publisher/links.json
```diff
@@ -0,0 +1,22 @@
...
+    "related_guides": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "alternative_format_provider_organisation": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "related_mainstream": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "additional_related_mainstream": {
```
boffbowsh: This would be better handled as a single links for both related and additional related, then a `related_mainstream_content` field in `details` which specifies the order.

formats/detailed_guide/publisher/details.json
```diff
@@ -0,0 +1,73 @@
...
+      "type": "string"
+    },
+    "excluded_nations": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": ["england", "scotland", "wales", "northern_ireland"]
+      }
+    },
+    "change_note": {
```
boffbowsh: Is this needed as well as `change_history`?

formats/detailed_guide/publisher/details.json
tvararu: Probably `DetailedGuide.first.scheduled_publication` but it shouldn't be public, so removing.

Postponed doing this until I did the backend work just to be sure of it.

formats/detailed_guide/publisher/details.json
tvararu: Case Studies have it, but I agree we don't need it.

If we do decide to remove it, worth noting: `first_public_at` is also called as such to maintain consistency with Case Studies, but I feel it should be `first_published_at`.

formats/detailed_guide/publisher/links.json
tvararu: So something like this?

Before:

```json
// links.json:
"related_mainstream": {
},
"additional_related_mainstream": {
}

// And accompanying example:
"related_mainstream": [{
  "content_id": "aaa"
}],
"additional_related_mainstream": [{
  "content_id": "bbb"
}]
```

After:

```json
// links.json:
"related_mainstream": {
}

// details.json
"related_mainstream_content": {
  "type": "string",
  "enum": ["main_first", "additional_first"]
}

// And accompanying example:
"related_mainstream": [{
  "content_id": "aaa"
}, {
  "content_id": "bbb"
}],
"details": {
  "related_mainstream_content": "main_first"
}
```

Could we just get rid of the details parameter altogether and establish that `main_first` is the only valid behaviour?

If we sometimes only need `additional_related_mainstream` content with no main `related_mainstream` content, we could potentially set `null` as the first item of the array, but that would require overriding the `links` type from `"$ref": "#/definitions/guid_list"` and potentially creating a new type of list, I think.

formats/detailed_guide/publisher/details.json
tvararu: Removed FWIW. :+1:

formats/detailed_guide/publisher/links.json
boffbowsh: The links aren't ordered, so the details field needs to contain an array of the content ids in order. The frontend will then map that ordering to the actual expanded links content. There's no concept of "first" from the links hash itself.

tvararu: Fair enough, I forgot about the ordering.

I think I prefer the current implementation, however. It doesn't require the collating of two disparate properties to figure out which is the `main` and what's the `additional` content. And then you have to search an array for a matching ID instead of just directly calling a property; it's clunkier to use. Explicit property names document slightly better as well.

But I'll defer to you; if you're sure, then let's do the `details. related_mainstream_content ` version.

boffbowsh: This is a pattern that will be repeated across a lot of formats, so it makes sense to start it here, even if there are only 2 entries allowed currently

tvararu: Right; it doesn't make sense for just 2 entries but it would for `n > 3`. I'll update it. :+1:

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: This comment thread has been sorted: https://github.com/alphagov/content-store/pull/203 will enable the content store to expand the parent tree recursively, thus sending the necessary info to the client.

formats/detailed_guide/frontend/examples/detailed_guide.json
tvararu: This comment thread has been sorted: https://github.com/alphagov/content-store/pull/203 will enable the content store to expand the parent tree recursively, thus sending the necessary info to the client.

tvararu:
I've added two commits (which I plan to fixup later; details/links https://github.com/alphagov/govuk-content-schemas/pull/279/commits/3f5301220e08a9d2962842ab6260e4c44728482a and compiled output https://github.com/alphagov/govuk-content-schemas/pull/279/commits/84ddd38ccd81595d653a95d49682fed915b791e6) which go hand in hand with https://github.com/alphagov/whitehall/pull/2553/commits/20b5ea4db7b60d7197a947c201eee12c162d6765 to align the `related_mainstream_content` with what we're actually working with.

Whitehall admin offers an interface to edit a few fields on the detailed guides model:

screen shot 2016-04-19 at 16 04 12

These were implemented in `govuk-content-schemas` as part of the `links` hash.

While trying to export them as links from within `whitehall`, me and @mgrassotti realised that it's not actually feasible. The only information that we have about these related mainstream content items is their assigned title and url. Crucially, we don't have any content IDs to work with, because it looks like they're just arbitrary URLs.

As such, we've updated the schema for this field (again). It's now part of the `details` hash and follows this structure:

```ruby
related_mainstream_content: {
  main: {
    title: "",
    url: "",
  },
  additional: {
    title: "",
    url: "",
  },
}
```

And, again, Whitehall can [export to this just fine](https://github.com/alphagov/whitehall/pull/2553/commits/20b5ea4db7b60d7197a947c201eee12c162d6765).

binaryberry:
@tvararu could you get the base path from the url provided and query the content-store to get the title?

tvararu:
I mean you're allowed to type whatever you wish into those text fields, AFAIK. Is there additional validation preventing you from that?

binaryberry:
Possibly not but for new entries, but we could show the user an error message if the content-item can't be found in the content-store, which would indicate an invalid url. There might be existing urls that don't work, but we could potentially leave the existing content title for these?

boffbowsh:
This is something we should try and tidy up. We have the option here of linking to either a piece of mainstream content via content id, OR an external URL. If we can use content ids we should, as it means we don't get broken links if content gets renamed.

boffbowsh:
There's 99 instances of incorrect data, 14 of which are genuine links to external sites, most of which are just badly formed links:

```sql
SELECT `related_mainstream_content_url`, `additional_related_mainstream_content_url` FROM editions 
WHERE 
	`related_mainstream_content_url` IS NOT NULL
AND
	`related_mainstream_content_url` <> ''
AND 
	`related_mainstream_content_url` NOT LIKE '%www.gov.uk%' 
OR 
	`additional_related_mainstream_content_url` IS NOT NULL
AND
	`additional_related_mainstream_content_url` <> ''
AND
	`additional_related_mainstream_content_url` NOT LIKE '%www.gov.uk%'
```

boffbowsh:
Actually only 5 when you limit it to published editions

boffbowsh:
I've made the changes to the content in production and am waiting for them to be reviewed. Then it's just a case of ensuring we parse the current urls correctly.

tvararu:
Right. Then I'm going to stash away the two commits in a branch, and revert this one so that it refers to `related_mainstream_content` via links, as before.

formats/detailed_guide/publisher/details.json
tvararu: I'm removing it.

tvararu:
Merging this, as it contains a good enough DetailedGuide example and schema which to the best of my knowledge addresses comments. Other changes will come with subsequent PRs.

fofr:
:fireworks:

PR #280, opened 2016-04-06
Add an optional note_description:string field to change_history

AndrewVos:
https://trello.com/c/qzHbv9S9/181-show-change-notes-and-page-history-at-the-bottom-of-guidance-page

In Service Manual, we're going to have two fields for change notes.

The current form of change history only stores "note" and no second value.

Because other teams use this type, we have made the newer field
optional.

In Service Manual we store change notes like this:

image

### DIFF
--- a/formats/definitions.json
+++ b/formats/definitions.json
@@ -151,7 +151,12 @@
             "format": "date-time"
           },
           "note": {
-            "type": "string"
+            "type": "string",
+            "description": "A summary of the change"
+          },
+          "reason_for_change": {
+            "type": "string",
+            "description": "Why the change is being made. This is multiline text, and will be rendered inside <p> tags. Govspeak is not yet supported here."
           }
         },
         "required": [
--- a/formats/service_manual_guide/publisher_v2/examples/with_change_history.json
+++ b/formats/service_manual_guide/publisher_v2/examples/with_change_history.json
@@ -1,0 +1,33 @@
+{
+  "publishing_app": "service-manual-publisher",
+  "rendering_app": "government-frontend",
+  "public_updated_at" : "2015-10-09T08:17:10+00:00",
+  "format": "service_manual_guide",
+  "locale": "en",
+  "details" : {
+    "body" : "<h2 id=\"what\">What it is, why it works and how to do it</h2>\n<p>Agile methodologies will help you and your team to build world-class, user-centred services quickly and affordably.</p>\n",
+    "related_discussion" : {
+      "title" : "Design community on hackpad",
+      "href" : "https://designpatterns.hackpad.com/"
+    },
+    "header_links" : [
+      { "title" : "What is it, why it works and how to do it", "href" : "#what-is-it-why-it-works-and-how-to-do-it" }
+    ],
+    "change_history": [
+      {
+        "public_timestamp": "2015-10-09T08:17:10+00:00",
+        "note": "This is a change",
+        "reason_for_change": "This is why we made this change\nand it has a second line of text"
+      }
+    ]
+  },
+  "base_path" : "/service-manual/agile",
+  "description" : "What agile is, why it works and how to do it",
+  "title" : "Agile",
+  "routes": [
+    {
+      "path": "/service-manual/agile",
+      "type": "exact"
+    }
+  ]
+}
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -444,7 +444,12 @@
             "format": "date-time"
           },
           "note": {
-            "type": "string"
+            "type": "string",
+            "description": "A summary of the change"
+          },
+          "reason_for_change": {
+            "type": "string",
+            "description": "Why the change is being made. This is multiline text, and will be rendered inside <p> tags. Govspeak is not yet supported here."
           }
         },
         "required": [
--- a/dist/formats/case_study/publisher/schema.json
+++ b/dist/formats/case_study/publisher/schema.json
@@ -315,7 +315,12 @@
             "format": "date-time"
           },
           "note": {
-            "type": "string"
+            "type": "string",
+            "description": "A summary of the change"
+          },
+          "reason_for_change": {
+            "type": "string",
+            "description": "Why the change is being made. This is multiline text, and will be rendered inside <p> tags. Govspeak is not yet supported here."
           }
         },
         "required": [
--- a/dist/formats/case_study/publisher_v2/links.json
+++ b/dist/formats/case_study/publisher_v2/links.json
@@ -233,7 +233,12 @@
             "format": "date-time"
           },
           "note": {
-            "type": "string"
+            "type": "string",
+            "description": "A summary of the change"
+          },
+          "reason_for_change": {
+            "type": "string",
+            "description": "Why the change is being made. This is multiline text, and will be rendered inside <p> tags. Govspeak is not yet supported here."
           }
         },
         "required": [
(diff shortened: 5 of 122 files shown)


dsingleton:
Could you give a bit more detail on the semantic difference between the two fields? From the change along it's not super clear _why_ there are two fields, and what different purposes they support.

At a guess, `note` is intended to be a short (one line?) summary of the change and `note_description` is for longer form (multi-line? supporting markup?) description of the change.

Apologies for the semantic nit-picking, but we want content items to be as self-documenting as possible, and avoid confusion down the line - so it's worth investing a bit up front :)

thehenster:
Exactly that - the `note_description` is the longer form.

@elliotcm mentioned that existing uses of `note` tend to be short / summaries, so we've tried to make `note_description` sound like the longer form of `note` without renaming it. In a perfect world we'd probably rename them into `summary` and `description` but that would require other apps to conform.

How about adding descriptions to each field? eg.

```
"note": {
  "type": "string",
  "description: "A summary of the change"
},
"note_description": {
  "type": "string",
  "description": "A more detailed description of the change"
}
```

dsingleton:
Using the `description` attributes for the fields is a huge help, thanks. I'm still not sure on the name, and it doesn't help clarify if the longer field has formatting, or is multiline.

(CC @boffbowsh who had an opinion on this from another discussion.)

boffbowsh:
Can you please explain more about what value this adds? What restriction is in place on the `note` field that you can't just use that?

boffbowsh:
An example somewhere might be useful

AndrewVos:
@boffbowsh Added a screenshot of what we do in service manual

boffbowsh:
Sorry, I mean an example in one of the format examples. I see why it's necessary now, and think it's fine. But I'm not sure about the name of it. We're in the context of an item of `change_history`, so maybe something like `full_description`?

AndrewVos:
@boffbowsh Added an example https://github.com/alphagov/govuk-content-schemas/pull/280/commits/e71a3110e813499559af578b48a43b8a3fd9d74d

AndrewVos:
The reason we went with `note_description` is because we already have `note`, and `_description` felt like the natural progression.

boffbowsh:
I think `note_description` is a bit non-sensical. It's not a description of a note, it's a description of why a change was made.

thehenster:
Naming! :sweat_smile: 

`full_description` might be good @boffbowsh. It ticks the box of sounding longer than `note` to me. However it does make `note`, like a notes field, sound optional to me. `note` is the required field.

I agree `note_description` doesn't make much sense by itself but it does sound secondary to `note`.

My guess is that without renaming `note` there won't be a really good answer. Y'all have seen more schemas than me so will have a better feel for what is "right".

Amending the descriptions to mention formatting / multiline is good suggestion @dsingleton :+1:

AndrewVos:
@thehenster @boffbowsh Done.

thehenster:
`full_description` wouldn't actually be my choice. But I'm wedded to none of them.

boffbowsh:
I get we're in danger of bikeshedding here, but also this is the first time this has been brought up, and this PR will essentially codify a standard that will be used across all government publishing. It's not an easy thing to change later, either. The name of it needs to make sense to developers in GDS, but also eventually developers in wider government with less readily-available context. I'd like to get the thoughts from someone who's been working full time on the pub API and has done naming of things in this context so far, so maybe @elliotcm, @tuzz, @danielroseman, @steventux or even @heathd would like to comment?

elliotcm:
Hey @boffbowsh.  I had a chat with this team when they brought this issue up with us.  The proposed solution here is what we settled on precisely because it doesn't codify a standard for the whole of gov.uk, because it's an optional field that we don't do anything special with.

One thing that could be done to reduce the scope of this is just to add it to the service manual formats rather than all of them, but in any case we have a ticket to address change notes across gov.uk and bring them all together.  Service Manual are aware that this might mean the removal of their extra field if that's the way the research and discussions come out.

https://trello.com/c/we9wSH0S/622-spike-investigate-if-anything-we-want-to-migrate-is-blocked-by-the-way-we-currently-model-change-notes-and-update-types

steventux:
I agree with @ecm; it might be worthwhile making this change bespoke to the service manual format in the short term, when we address change notes ftw this and other definitions will inform the global definition.

heathd:
Hello!

**Regarding the naming of the field**

given that the user input field is labelled **"Why the change is being made"**, perhaps `rationale` would convey purpose of the field better (or `reason_for_change` or `justification_for_change`)?

**Regarding the approach of adding an optional field**

that seems fine to me, provided that the `note` field retains the same semantic purpose in both contexts. If not, perhaps consider whether the `change_history` in a `service_manual_guide` should actually be a different type such as `detailed_change_history`. I don't have enough context to say which is more appropriate. I guess it's preferable to avoid diverging if you can, however.

Hope that helps!

tuzz:
I think it's ok to introduce this optional field across all formats as it seems like a sensible addition. It isn't particularly unique to one format and I'd rather we have one way of doing this rather than each publishing app independently arriving at different solutions for this problem.

That being said, I don't think the `full_description` name is representative of what this field is for and I think @heathd's suggestions are more appropriate. I'd personally prefer `reason_for_change`, but I'm not opposed to the others.

I will also leave a couple of comments on the file changes that I think should be addressed.

formats/definitions.json
```diff
@@ -151,7 +151,12 @@
             "format": "date-time"
           },
           "note": {
-            "type": "string"
+            "type": "string",
+            "description": "A summary of the change"
+          },
+          "full_description": {
+            "type": "string",
+            "description": "A more detailed description of the change. This is multiline text, and will be rendered inside <p> tags."
```
tuzz: What representation does multiline text use when this field is used?

Does it support Govspeak or is it just newline characters (`\n`) ?

formats/service_manual_guide/publisher_v2/examples/with_change_history.json
```diff
@@ -0,0 +1,33 @@
...
+      "href" : "https://designpatterns.hackpad.com/"
+    },
+    "header_links" : [
+      { "title" : "What is it, why it works and how to do it", "href" : "#what-is-it-why-it-works-and-how-to-do-it" }
+    ],
+    "change_history": [
+      {
+        "public_timestamp": "2015-10-09T08:17:10+00:00",
+        "note": "This is a change",
+        "full_description": "This is why we made this change"
```
tuzz: It would be useful if the example used multiline text to demonstrate how this works.

thehenster:
I like `reason_for_change` too.

I still believe the semantics of `note` change in the context of the extra field, optional or not. We are changing it's meaning without changing the field. Maybe that's just life though and will be addressed when the publishing team look at their change history story.

Thanks to all for helping out btw :+1:

AndrewVos:
Can we have a vote on `reason_for_change` please? I've rebased this thing more times than there are lines of code....

AndrewVos:
Please add reactions to that comment ^^^ :)

formats/definitions.json
AndrewVos: It does not support govspeak at this time, though we may change this later.

AndrewVos:
@tuzz Did you want me to mention that the field doesn't support govspeak, in the description?

boffbowsh:
`reason_for_change` sounds good to me. Thanks for chiming in, @heathd :)

AndrewVos:
@elliotcm @tuzz @boffbowsh @heathd Could someone click that merge button please :)

tuzz:
> @AndrewVos: @tuzz Did you want me to mention that the field doesn't support govspeak, in the description?

No, that's ok. The example has a newline character so I think that's fine.

:tada: :star:

PR #281, opened 2016-04-06
Add administrative appeals tribunal decisions schema and examples

robmckinnon:
The Upper Tribunal Administrative Appeals Chamber (UTAAC) publishes decisions that describe the outcome of a tribunal case.

Tribunal decision finders are used by public, professionals and the tribunals themselves to search for past decisions.

Published decisions form a part of case law and are used for research purposes and to prepare for current or future cases.

### DIFF
--- a/formats/specialist_document/frontend/examples/utaac-decision.json
+++ b/formats/specialist_document/frontend/examples/utaac-decision.json
@@ -1,0 +1,53 @@
+{
+  "content_id": "3917cb0e-0928-45b7-bf63-20f9bc113a7b",
+  "base_path": "/administrative-appeals-tribunal-decisions/example-utaac-decision",
+  "title": "Example UTAAC Decision",
+  "description": "Example text.",
+  "format": "specialist_document",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "2016-04-05T14:41:08.625Z",
+  "public_updated_at": "2016-04-05T14:41:08.000+00:00",
+  "details": {
+    "metadata": {
+      "hidden_indexable_content": "Example text.",
+      "tribunal_decision_categories": [
+        "benefits-for-children",
+        "bereavement-and-death-benefits"
+      ],
+      "tribunal_decision_decision_date": "2015-08-30",
+      "tribunal_decision_judges": [
+        "angus-r"
+      ],
+      "tribunal_decision_sub_categories": [
+        "benefits-for-children-benefit-increases-for-children",
+        "bereavement-and-death-benefits-bereaved-parents-allowance"
+      ],
+      "bulk_published": false,
+      "document_type": "utaac_decision"
+    },
+    "change_history": [
+      {
+        "public_timestamp": "2016-04-05T14:41:08+00:00",
+        "note": "First published."
+      }
+    ],
+    "body": "<p>Example text.</p>\n"
+  },
+  "phase": "beta",
+  "links": {
+    "available_translations": [
+      {
+        "content_id": "3917cb0e-0928-45b7-bf63-20f9bc113a7c",
+        "title": "Example UTAAC Decision",
+        "base_path": "/administrative-appeals-tribunal-decisions/example-utaac-decision",
+        "description": "Example text.",
+        "api_url": "http://content-store.dev.gov.uk/content/administrative-appeals-tribunal-decisions/example-utaac-decision",
+        "web_url": "http://www.dev.gov.uk/administrative-appeals-tribunal-decisions/example-utaac-decision",
+        "locale": "en"
+      }
+    ]
+  },
+  "schema_name": "specialist_document",
+  "document_type": "utaac_decision"
+}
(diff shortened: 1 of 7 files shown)


formats/specialist_document/publisher/details.json
```diff
@@ -880,6 +881,64 @@
...
+        },
+        "tribunal_decision_sub_categories_name": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "tribunal_decision_decision_date": {
+          "type": "string",
+          "pattern": "^[1-9][0-9]{3}-(0[1-9]|1[0-2])-([012][0-9]|3[0-1])$"
```
h-lame: Could we use `"type": "date"` here?  Particularly as this pattern looks like it allows `00` as the dd part of `yyyy-mm-dd`?

formats/specialist_document/frontend/examples/utaac-decision.json
```diff
@@ -0,0 +1,53 @@
...
+  "base_path": "/administrative-appeals-tribunal-decisions/example-utaac-decision",
+  "title": "Example UTAAC Decision",
+  "description": "Example text.",
+  "format": "specialist_document",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "2016-04-05T14:41:08.625Z",
+  "public_updated_at": "2016-04-05T14:41:08.000+00:00",
+  "details": {
+    "metadata": {
```
h-lame: This metadata seems to be missing entries for `tribunal_decision_categories_name`, `tribunal_decision_judges_name`, and `tribunal_decision_sub_categories_name`.

formats/specialist_document/publisher/details.json
robmckinnon: We followed this repo's instructions for [converting finder schemas](https://github.com/alphagov/govuk-content-schemas/blob/master/docs/converting-finder-schemas.md#converting-finder-schemas).

So the [FinderSchemaConverter](https://github.com/alphagov/govuk-content-schemas/blob/master/lib/govuk_content_schemas/finder_schema_converter.rb) was run on the [UTAAC schema](https://github.com/ministryofjustice/specialist-publisher/blob/master/finders/schemas/administrative-appeals-tribunal-decisions.json).

The [converter outputs that regexp pattern](https://github.com/alphagov/govuk-content-schemas/blob/master/lib/govuk_content_schemas/finder_schema_converter.rb#L147) in place of [date fields from the schema](https://github.com/ministryofjustice/specialist-publisher/blob/master/finders/schemas/administrative-appeals-tribunal-decisions.json#L2379).

Do GDS want to change the converter to output `"type": "date"` instead? If yes, then all other dates in the `details.json` probably should be changed to `"type": date` then. Maybe that could be done in a separate pull request, changing the UTAAC entry at that time?

h-lame: This is probably just my unfamiliarity with the schemas and finders!  Probably `"type":"date"` is overkill, but that regexp seems suspect to me.  If anything does need to change you're right that it would be across all schemas, not just UTAAC, so let's not hold up this PR for that.

formats/specialist_document/frontend/examples/utaac-decision.json
robmckinnon: I took this JSON from a locally running content-store instance, e.g. http://content-store.dev.gov.uk/content/administrative-appeals-tribunal-decisions/example-utaac-decision . My understanding is that this is the correct approach to get a realistic example file.

You're right, there are no values for `tribunal_decision_categories_name`, `tribunal_decision_judges_name`, and `tribunal_decision_sub_categories_name`. Not sure why that is.

Though in my locally running finder frontend, those values are populated correctly:
screen shot 2016-04-12 at 15 35 58

Did you want to check on the integration content store to see what the JSON output is? I'm not sure of the url for the integration content-store.

h-lame: I've checked the integration content-store for the same document and it doesn't have the missing fields in the JSON rep either.  I'll see if anyone else can help out.

h-lame: They're not in content-store, or the examples because these three fields aren't actually sent by specialist-publisher to the publishing-api, they're only sent to rummager for search indexing.  What this means is they shouldn't be in the schema; however as there's prior art where these fields are in the schema we won't let it hold up this PR and we'll fix it later.

h-lame:
@robmckinnon - now I've clarified the points I raised this looks good to merge, but it needs a rebase before we can.

robmckinnon:
@h-lame Thanks for the review. I've rebased in preparation for a merge. A suggested feature improvement for this process would be to somehow automate altering the schema details and examples. Perhaps triggered by changes to the specialist-publisher repo. At present it's a manually intensive process, but mostly following mechanical steps. There's no automated mechanism to ensure these files stay in sync with the schema in specialist-publisher either.

PR #284, opened 2016-04-07
Specialist documents has Govspeak and html in body

christophwong:
In V2 of specialist publisher, we want to save Govspeak content ([PR here](https://github.com/alphagov/specialist-publisher/pull/697)), so we
can use it for editing and previews.


benilovj:
This duplicates https://github.com/alphagov/govuk-content-schemas/pull/278 which has been merged, so closing.

PR #286, opened 2016-04-13
Tidy up schemas

h-lame:
Best reviewed as single commits, but the tl;dr is:

1. Remove `xxx_name` fields from specialist documents as these are only ever sent to rummager, never to publishing-api.
2. Fix the pattern used to match "date" fields to disallow a value of `00` in the `dd` part of `yyyy-mm-dd`


elliotcm:
👍 You'll probably want another commit after running `rake`.

elliotcm:
Correction, you've already run that.  Could you pull the `dist/*` changes into separate commits?

h-lame:
Sure.  Incoming in 5... 4...

h-lame:
@elliotcm done all the dist changes as one commit, that ok?

elliotcm:
Marvellous.

PR #287, opened 2016-04-18
Add request_id to travel advice formats

steventux:
https://trello.com/c/Io2JYTFv/27-ensure-request-id-is-available-to-multipage-frontend-and-frontend-apps

In order to trace the `request_id` thoroughly for travel advice content updates we will need to persist it and render it. The [request id will come from the load balancer](https://github.com/alphagov/govuk-puppet/pull/4320/commits/fdfbc827bf6d39e32dc4227c95fd01633b1ed582) and be [added to worker jobs](https://github.com/alphagov/travel-advice-publisher/pull/139/commits/e31ed8a3f56920fe973346547f08698d5cf300af), it will also flow through the publishing pipeline and be present in the page source of rendered content.

### DIFF
--- a/dist/formats/travel_advice/frontend/schema.json
+++ b/dist/formats/travel_advice/frontend/schema.json
@@ -260,6 +260,10 @@
         "max_cache_time": {
           "description": "The maximum length of time the content should be cached, in seconds.",
           "type": "integer"
+        },
+        "publishing_request_id": {
+          "description": "A unique identifier used to track publishing requests to rendered content",
+          "type": "string"
         }
       }
     },
--- a/dist/formats/travel_advice/publisher/schema.json
+++ b/dist/formats/travel_advice/publisher/schema.json
@@ -89,6 +89,10 @@
         "max_cache_time": {
           "description": "The maximum length of time the content should be cached, in seconds.",
           "type": "integer"
+        },
+        "publishing_request_id": {
+          "description": "A unique identifier used to track publishing requests to rendered content",
+          "type": "string"
         }
       }
     },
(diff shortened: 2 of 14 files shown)


tijmenb:
So this `request_id` is the ID of the request that originally published the content item?

steventux:
@tijmenb yes, I've added a better description of where it comes from and where it goes.

tijmenb:
Would naming it something like `original_publishing_request_id` make sense? That probably gives more context to someone only looking at the output from content-store.

gpeng:
I agree with @tijmenb that `request_id` is a bit vague. Would `publishing_request_id` be sufficient. ie. the id of the request that did the publishing. That said, I could go with `original_publishing_request_id` if you feel that's clearer.

tijmenb:
I'm also good with `publishing_request_id` 👍

steventux:
thanks @tijmenb + @gpeng amended

boffbowsh:
Won't the request_id already be in publishing-api as part of https://github.com/alphagov/publishing-api/pull/273? Therefore it won't need adding to the publisher schemas. Then, it'd be better to change content store to always present the request_id for all formats, and change the frontend schemas for all formats. Where possible with this TAP work, we shouldn't make solutions for just TAP when the wider publishing platform will benefit, else TAP ends up becoming more of a special snowflake than it already is.

gpeng:
My concern with us using the `govuk_request_id` as a published version unique id is that it gets out of sync if we inadvertently persist it somewhere along the stack for a request that isn't 'the publishing request'. We already see 'side effects' in the stack such as the live content store content item having its `updated_at` changed when a draft item is created. I'm not sure whether this is best raised/solved here or in https://github.com/alphagov/publishing-api/pull/273 (or elsewhere) but I think it's worth discussing.

boffbowsh:
Ok, in that case we could store an array of event ids that affected the content item. Either as a normal postgres array or as a separate join table. This would help validate the above side affects comment.

steventux:
I think recording events should be the responsibility of the event log, I'd be wary of duplicating this in the content item or an associated model. I agree that the content item -> event association is weak and needs some thinking.

PR #288, opened 2016-04-22
Add withdrawn notices to detailed guides

fofr:
* Base on case studies withdrawn pattern
* Add withdrawn detailed guide example
* Extract withdrawn notice into a definition so it is consistently used
* Update detailed guides and case studies to use this definition

Part of:
https://trello.com/c/P2CLsRtp/383-6-detailed-guides-migration-front-end-work-withdrawing-documents-small


dsingleton:
👍 for the de-duping, LGTM.

PR #290, opened 2016-04-25
Add government and political fields to detailed guides

fofr:
The plan is to allow government to be a link, so that `current`, names and slugs can be updated automatically without having to republish. However governments represent URL-less content, which the content store doesn’t currently handle.

As an intermediate to unblock detailed guides, put government into the details hash instead.

* From `government` and `political`, logic that determines if the history banner should show can be derived.
* Government slug needed for analytics
* Include example of a political detailed guide

Political example based on:
https://www.gov.uk/guidance/onshore-wind-part-of-the-uks-energy-mix

Part of:
https://trello.com/c/74ymVfIw/384-6-detailed-guides-migration-front-end-work-for-history-mode-political-status-and-government-small


formats/definitions.json
```diff
@@ -52,6 +73,10 @@
         }
       }
     },
+    "political": {
+      "type": "boolean",
+      "description": "If the content is considered political in nature, relfecting views of the government it was published under."
```
boffbowsh: Typo, "reflecting".

boffbowsh:
Minor typo, otherwise 👍

fofr:
@boffbowsh Typo fixed.

dsingleton:
👍

tvararu:
Looks good to me. 👍

PR #291, opened 2016-04-26
Remove unused tags from detailed guide details

fofr:
These mainstream browse pages, topics and policies have equivalents in links/base_links. They aren’t being used by the frontend, they are empty in the examples.

See:
https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/detailed_guide/publisher_v2/links.json
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/base_links.json

cc @tvararu @tijmenb


tijmenb:
👍 excellent

tvararu:
:+1:

PR #296, opened 2016-05-03
Remove updated_at from detailed_guide examples.

tvararu:
Whitehall won't publish these, and they're not being used.
`public_updated_at` is used instead in `government-frontend`.


boffbowsh:
This field is provided by content store directly, so it should be in the examples.

PR #292, opened 2016-04-26
Revert "Remove unused tags from detailed guide details"

tijmenb:
Reverts alphagov/govuk-content-schemas#291.

The `details.tags` is used for email-alerts:

https://github.com/alphagov/whitehall/blob/39f8a6342f449f9a0f01701afce61212cac97e47/app/presenters/publishing_api_presenters/edition.rb#L63-L74

PR #298, opened 2016-05-04
Create minimum viable publications format

fofr:
https://trello.com/c/v85kPiKN/366-9-publications-migration-mvp-content-schema-examples-and-front-end-work-medium

See https://github.com/alphagov/government-frontend/pull/139 for an example rendering.

Details:

* Purposefully omit:
  * __Excluded nations__ - pattern used by detailed guides may not be correct
  * __Format subtype__
* Include government, political and first_public_at as required fields, this also matches detailed guides
* Publications can be withdrawn, include withdrawal notice
* Case studies and detailed guides both need a tags block in the details, `details.tags` are used for email alerts: #292
* Documents is an array of attachments, each attachment contains the markup necessary to render it within a govspeak block. This is an intentionally dumb approach, and matches the approach taken
for embedded attachments. See https://docs.google.com/document/d/1Flx_eCUzixmkTT6WgjCN2bJzQ8HuVCwMga5RxUsv07M/edit

Links:

* Add ministers
* Policies and policy areas are included in base_links
* Add related statistical data sets
* Add topical events
* Add world_locations, matching Case Studies
* Make leading organisations a required field, provided by base_links
* Include supporting organisations to match Case Studies – could be provided by base_links "organisations" instead, whichever it is, the two should be consistent

Include simple example based on:
https://www.gov.uk/government/publications/d17-9ly-veolia-es-uk-limited-environmental-permit-issued


formats/publication/publisher/details.json
```diff
@@ -0,0 +1,63 @@
...
+    "political"
+  ],
+  "properties": {
+    "body": {
+      "type": "string"
+    },
+    "documents": {
+      "type": "array",
+      "items": {
+        "type": "string"
```
boffbowsh: Why are these being rendered by the publishing app? These don't need to be embedded in Govspeak, so I think they should be made available as plain data and rendered by the frontend. Or was there a conversation I missed?

fofr: I don't understand your comment.

They are responsible for this bit in the frontend:
screen shot 2016-05-06 at 13 41 08

Details about this in this commit message, with links to decisions: 08d08f29b34275049b096e0239603308a30ff5bc

boffbowsh: Ok, so that's a decision I missed. Fine as a short term solution 👍

formats/publication/publisher/links.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "lead_organisations"
+  ],
+  "properties": {
+    "document_collections": {
```
boffbowsh: These are supplied by content store / dependency resolution, not by the publishing app.

formats/publication/frontend/examples/withdrawn_publication.json
```diff
@@ -0,0 +1,58 @@
...
+        "base_path": "/government/organisations/department-for-environment-food-rural-affairs",
+        "api_url": "https://www.gov.uk/api/content/government/organisations/department-for-environment-food-rural-affairs",
+        "web_url": "https://www.gov.uk/government/organisations/department-for-environment-food-rural-affairs",
+        "locale": "en",
+        "analytics_identifier": "D7"
+      }
+    ]
+  },
+  "schema_name": "publication",
+  "document_type": "publication"
```
boffbowsh: I know this is deliberately left for later, but if you specify the correct key here (`guidance`) and for the Veolia publication (`notice`), we get 50% of that work done.

formats/publication/publisher/links.json
fofr: I'm continuing to match other formats, eg Case Studies: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/case_study/publisher/links.json#L24-L26

Presumably the document collection work needs a PR to remove that block from the detailed guides and case studies links hash?

boffbowsh: Yes, it will do.

fofr: Removing `document_collections` gives:
```The property '#/links' contains additional properties ["document_collections"] outside of the schema when none are allowed in schema file:///var/govuk-sites/govuk-content-schemas/dist/formats/publication/frontend/schema.json#
The following examples don't validate against their schemas:
formats/publication/frontend/examples/political_publication.json```

boffbowsh: Ah, that's because there's no way of generating separate frontend_links at the moment, sorry. Ok fair enough 👍

PR #299, opened 2016-05-05
Allow rendering_app & publishing_app in frontend schemas

tijmenb:
The current frontend schemas do not allow `rendering_app` & `publishing_app` to be set.

Since https://github.com/alphagov/content-store/pull/209 we expose those attributes, making the schemas invalid.

To prevent the current examples and API response from being invalid, we make the attributes optional. Once all is deployed we should update the examples and make the attributes required.


tijmenb:
(Test failure is unrelated)

PR #300, opened 2016-05-05
Add national_applicability definition and example.

tvararu:
This replaces `excluded_nations` with something which is more
frontend-friendly. Using this structure means that frontends don't need to
invert the list of all nations to determine which nations a publication
`Applies to:`.

Review commit by commit for simplicity.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -490,6 +490,44 @@
           "format": "date-time"
         }
       }
+    },
+    "nation_applicability": {
+      "description": "An object specifying the applicability of a particular nation.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "label": {
+          "description": "The pretty-printed, translated label for this nation.",
+          "type": "string"
+        },
+        "alternative_url": {
+          "description": "An optional alternative URL to link to for more information on this content item pertaining to this nation.",
+          "type": "string"
+        },
+        "applicable": {
+          "description": "Whether the content applies to this nation or not.",
+          "type": "boolean"
+        }
+      }
+    },
+    "national_applicability": {
+      "description": "An object specifying the applicable nations for this content item. If it applies to all nations, it should be omitted.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "england": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "northern_ireland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "scotland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "wales": {
+          "$ref": "#/definitions/nation_applicability"
+        }
+      }
     }
   }
 }
--- a/dist/formats/case_study/publisher/schema.json
+++ b/dist/formats/case_study/publisher/schema.json
@@ -355,6 +355,44 @@
           "format": "date-time"
         }
       }
+    },
+    "nation_applicability": {
+      "description": "An object specifying the applicability of a particular nation.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "label": {
+          "description": "The pretty-printed, translated label for this nation.",
+          "type": "string"
+        },
+        "alternative_url": {
+          "description": "An optional alternative URL to link to for more information on this content item pertaining to this nation.",
+          "type": "string"
+        },
+        "applicable": {
+          "description": "Whether the content applies to this nation or not.",
+          "type": "boolean"
+        }
+      }
+    },
+    "national_applicability": {
+      "description": "An object specifying the applicable nations for this content item. If it applies to all nations, it should be omitted.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "england": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "northern_ireland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "scotland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "wales": {
+          "$ref": "#/definitions/nation_applicability"
+        }
+      }
     }
   },
   "oneOf": [
--- a/dist/formats/case_study/publisher_v2/links.json
+++ b/dist/formats/case_study/publisher_v2/links.json
@@ -282,6 +282,44 @@
           "format": "date-time"
         }
       }
+    },
+    "nation_applicability": {
+      "description": "An object specifying the applicability of a particular nation.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "label": {
+          "description": "The pretty-printed, translated label for this nation.",
+          "type": "string"
+        },
+        "alternative_url": {
+          "description": "An optional alternative URL to link to for more information on this content item pertaining to this nation.",
+          "type": "string"
+        },
+        "applicable": {
+          "description": "Whether the content applies to this nation or not.",
+          "type": "boolean"
+        }
+      }
+    },
+    "national_applicability": {
+      "description": "An object specifying the applicable nations for this content item. If it applies to all nations, it should be omitted.",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "england": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "northern_ireland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "scotland": {
+          "$ref": "#/definitions/nation_applicability"
+        },
+        "wales": {
+          "$ref": "#/definitions/nation_applicability"
+        }
+      }
     }
   }
 }
--- a/dist/formats/case_study/publisher_v2/schema.json
+++ b/dist/formats/case_study/publisher_v2/schema.json
@@ -298,6 +298,44 @@
           "format": "date-time"
         }
       }
+    },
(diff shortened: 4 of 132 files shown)


boffbowsh:
This is only valid for `detailed_guide`s so shouldn't go in the main `definitions` file so as not to pollute other formats.

tvararu:
@fofr pointed out that he would need it for Publications as well, which are coming soon, which is why I placed it in the main definitions. I'll oblige either way. 👌

boffbowsh:
Oh, ok. Fair enough.

boffbowsh:
Needs a rebase anyway.

tvararu:
I will rebase it after I complete the Whitehall publisher implementation, and then merge them back to back.

tvararu:
Rebased, PR on Whitehall for the dependant contract tests that are failing: https://github.com/alphagov/whitehall/pull/2580

fofr:
Nice separation of commits.

fofr:
What about the alternative content links?

tvararu:
@fofr I remember you mentioning that they should be part of this object rather than separate. Let's try to find an example on Whitehall that has them, and go from there, updating this PR with another example and updated schema that scopes the feature in.

tvararu:
Two additional commits added which can be squashed into the previous two (or at the least, reordered slightly and two `update dist` ones squashed together).

tvararu:
So I've found an example with alternative URLs at https://www.gov.uk/guidance/report-lost-or-stolen-cattle.

@fofr I think we may have to revert to the `excluded_nations` version for the detailed guide schema. Now that we need to consider `alternative_url`s, it seems like we’d have to overcomplicate our current approach to include them.

Given this fictitious example:

```
"applicable_nations": {
  "all_nations": false,
  "nations": ["England"]
}
```

Let’s say that `Scotland` and `Northern Ireland` both have alternative urls. You can’t add them to this key, because they’re not `applicable_nations`, so they logically don't belong here. You'd need another key.

If we replace it with `excluded_nations`, which was the initial thought:

```
"excluded_nations": [
  {
    "label": "Scotland",
    "alternative_url": "http://www.scotland.org/"
  },
  {
    "label": "Northern Ireland",
    "alternative_url": "http://www.northernireland.gov.uk/"
  },
  {
    "label": "Wales",
    "alternative_url": ""
  }
]
```

Which is terse, but doesn't explicitly state all the nations involved. This is not a huge concern given that there's only 4 of them and they (hopefully) won't change significantly. However, it does require the frontend to handle this concern.

Here's an idea for a middle ground that is explicit about both:

```
"nations": [
  {
    "label": "Scotland",
    "alternative_url": "http://www.scotland.org/",
    "applicable": false,
  },
  {
    "label": "Northern Ireland",
    "alternative_url": "http://www.northernireland.gov.uk/",
    "applicable": false,
  },
  {
    "label": "Wales",
    "alternative_url": "",
    "applicable": false,
  }
  {
    "label": "England",
    "alternative_url": "",
    "applicable": true
  }
]
```

This way no information is inferred, everything is laid out plainly. It's verbose, but I like it best I think. It's also easy to work with on the frontend; you can filter the array to get `applicable`/`excluded`.

Wide open to other suggestions.

dsingleton:
I like that approach, 👍  to being explicit. 

Nitpicking, but for the root `national_applicability` (or similar) over `nations`. Presumably we'd omit that field entirely for formats without national applicability? What about for a format that can have national applicability but applies to everything, I don't think we'd still want the field with all nations `applicable: true`, right?

tvararu:
> Presumably we'd omit that field entirely for formats without national applicability?

That's technically not possible; [`whitehall/app/models/nation.rb`](https://github.com/alphagov/whitehall/blob/master/app/models/nation.rb#L16-L18) defines:

```ruby
def self.potentially_inapplicable
  [Scotland, Wales, NorthernIreland]
end
```

But even if it were…

> What about for a format that can have national applicability but applies to everything, I don't think we'd still want the field with all nations applicable: true, right?

I can see why we would not, I agree that it's redundant. But because the data set is n=4 (very small) anyway, I think it's fine not to denormalize this information as a separate value.

As such, I believe the frontend should do:

```ruby
def applies_to_all_nations
  content_item["details"]["nations"].find({ n } !n.applicable ).nil?
end
```

Or something similar, but that's fine.

If we decide to derive data on the frontend's behalf, there's a lot more we could send down potentially:

```json
"national_applicability": {
  "applies_to_all_nations": false,
  "applies_to_at_least_one_nation": true,
  "some_nations_have_alternative_urls": true,
  "nations": [...]
}
```

I think it's best to provide the frontend with all the necessary information, and allow it to derive whichever data it needs. We shouldn't cater to usecases that are too specific, because those can and will change over time.

> Nitpicking, but for the root `national_applicability` (or similar) over nations. 

I was going to name it that but I hate typing `applicability`. I'll rename it. 😛

fofr:
> `national_applicability`
`devolution`?

boffbowsh:
What about:
```json
{
  "national_applicability": {
    "england": {
      "applicable": true
    },
    "scotland": {
      "applicable": false,
      "alternative_url": "..."
    },
    ...
  }
}
```

?

boffbowsh:
I agree with Theo on being explicit here for everything, and providing the data for the frontend to derive the correct thing to do.

tvararu:
@boffbowsh I thought about that, but since the keys would have to serve as labels, it would be:

```json
{
  "national_applicability": {
    "England": {
      "applicable": true
    },
    "Scotland": {
      "applicable": false,
      "alternative_url": "..."
    },
    "Northern Ireland": { ... },
    "Wales": { ... },
  }
}
```

And case sensitive keys, and ones that contain spaces, aren't to everyone's liking. I'm perfectly fine with them (as long as they don't make some parser choke), and I can go with this approach instead if requested.

boffbowsh:
Keys can either be already humanised, or, my preference, frontend converts them as:

```ruby 
'northern_ireland'.humanize.titleize
=> "Northern Ireland"
```

But I have no objection to JSON keys with caps and spaces in them.

fofr:
If we need to pass the name through, I'd prefer:
```json
england: {
  name: "England"
}
```

Capitalised keys, or keys with spaces, looks plain wrong to me. Like it's breaking some imaginary style guide.

dsingleton:
I'm still on the fence about omitting the field for something with no applicability limits - given thats the overwhelmingly common case and is semantically equivalent to a format without applicability limits. 

In the abstract - which is what what we should be thinking about with the schema, IMO - any format/document could have national applicability, and would we want to send the full representation for all documents of that type?

tvararu:
@boffbowsh Hmm, I think that could potentially not work well with future translation support. A few examples:

```json
"locale": "it",
"details": {
  "national_applicability": {
    "inghilterra": { ... }, // Becomes Inghilterra.
    "galles": { ... }, // Becomes Galles.
    "irlanda_del_nord": { ... } // Becomes Irlanda Del Nord, which is incorrect. It's "Irlanda del nord". Lowercase n.
    "scozia": { ... } // Becomes Scozia.
  }
}
```

```json
"locale": "ro",
"details": {
  "national_applicability": {
    "anglia": { ... }, // Becomes Inghilterra.
    "țara_galilor": { ... }, // Becomes Țara Galilor.
    "irlanda_de_nord": { ... } // Becomes Irlanda de Nord, which is incorrect. It's "Irlanda de Nord". Uppercase N.
    "scoția": { ... } // Becomes Scoția.
  }
}
```

And I'm sure stuff like hiragana can potentially throw more wrenches in the system, so no, I don't think the programmatic capitalisation should be used.

It's either english keys, in which case every frontend needs to have a locale mapping for the countries to every possible language it would have to serve, or sending down the exact localised string.

> I'm still on the fence about omitting the field for something with no applicability limits - given thats the overwhelmingly common case and is semantically equivalent to a format without applicability limits.

What are the benefits to omitting it? I'm aware and agree with saving bytes (as you point out, potentially on every single format), but are there others?

fofr:
Why do keys need to be labels? The keys need to be keys for the strings in locale files.

dsingleton:
> What are the benefits to omitting it? I'm aware and agree with saving bytes (as you point out, potentially on every single format), but are there others?

Simplicity of content items. I think the more things they have to represent the more dilute the meaningful content becomes. It shouldn't have stuff that it doesn't need.

It's more of a feel/reckon about how we should be using schemas/content items in general. I wouldn't hold up the PR on it, but it's worth raising.

tvararu:
@fofr `government-frontend` has no locale files, not at the moment. Whitehall doesn't have keys for countries either, but hypothetically speaking, if some codebase were to gain them, it would be the publisher, since the published documents are already translated in most every other field ("title", "details.body", et al). You may as well send the names of the nations translated, I think.

@dsingleton So the contract would be:

- if `national_applicability` is present, use it to determine what nations this content applies to.
- else, it means it applies to everything.

It's the `else` that isn't communicated explicitly by omitting it. But yeah, it's a sensible assumption to make anyway, and if it removes this much boilerplate, we should go your route. You've convinced me.

tvararu:
I've reworked the field, as discussed above; check the temporary commit at https://github.com/alphagov/govuk-content-schemas/pull/300/commits/be0c62c6629e8ce4c5335769650e181a49efa2fe. (it will be squashed before merging)

dsingleton:
👍  to the new structure

fofr:
Squash commits (to avoid to and fro), rebase against master and then we're good to merge?

tvararu:
Yep, seems like it. Right away.

PR #306, opened 2016-05-11
Add `emphasised_organisations` to details hash

jackscotti:
Currently, there are three types of organisations stored in the publishing-api database:

- lead organisations
- organisations
- supporting organisations
Where: organisations = lead organisations + supporting organisations

GOV.UK publishing apps do not send to the publishing-api the same organisation types consistently. We are going to modify all applications to send at least the organisations content ids in the links hash and the "emphasised_organisations" in the details hash. 
Frontend applications will then decide what and how to display the organisations by reading the details and links hashes of a piece of content. 

The `emphasised_organisations` in details hash will allow frontend applications to decide the order in which to display organisations. (emphasised_organisations will go first)

ticket: https://trello.com/c/us3MI1n9/602-fix-organisation-tagging
related to: https://github.com/alphagov/whitehall/pull/2585

Future steps:
- remove lead and supporting organisations from the schemas

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -270,6 +270,13 @@
               "format": "date-time"
             }
           }
+        },
+        "emphasised_organisations": {
+          "description": "The content ids of the organisations that should be displayed first in the list of organisations related to the item, these content ids must be present in the item organisation links hash.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/guid"
+          }
         }
       }
     },
--- a/dist/formats/case_study/publisher/schema.json
+++ b/dist/formats/case_study/publisher/schema.json
@@ -78,6 +78,13 @@
               "format": "date-time"
             }
           }
+        },
+        "emphasised_organisations": {
+          "description": "The content ids of the organisations that should be displayed first in the list of organisations related to the item, these content ids must be present in the item organisation links hash.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/guid"
+          }
         }
       }
     },
@@ -96,6 +103,7 @@
           "$ref": "#/definitions/guid_list"
         },
         "supporting_organisations": {
+          "description": "DEPRECATED: this field will be removed.",
           "$ref": "#/definitions/guid_list"
         },
         "world_locations": {
--- a/dist/formats/case_study/publisher_v2/links.json
+++ b/dist/formats/case_study/publisher_v2/links.json
@@ -21,6 +21,7 @@
           "$ref": "#/definitions/guid_list"
         },
         "supporting_organisations": {
+          "description": "DEPRECATED: this field will be removed.",
           "$ref": "#/definitions/guid_list"
         },
         "world_locations": {
--- a/dist/formats/case_study/publisher_v2/schema.json
+++ b/dist/formats/case_study/publisher_v2/schema.json
@@ -78,6 +78,13 @@
               "format": "date-time"
             }
           }
+        },
+        "emphasised_organisations": {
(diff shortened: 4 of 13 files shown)


boffbowsh:
I think I prefer `primary_organisations`, or maybe `lead_organisations` should be retained as it matches Whitehall's UI.

tijmenb:
Thanks @boffbowsh. I agree that it's important to keep UI and data naming in sync. 

However in this case it may be better to give it an (overly) specific name, because people could interpret "lead_organisations" as having some sort of real meaning in the context of government.

We also want to use the same mechanism in policy-publisher, where the UI is different:

screen shot 2016-05-12 at 10 19 46

In short, `lead_organisation` could be misinterpreted and lead to people using it in different incompatible ways. Something like `organisations_displayed_first` is easier to change to something more broad, while the reverse is not true. Thoughts?

boffbowsh:
Could just do `organisation_display_order` and list all orgs?

fofr:
> In short, lead_organisation could be misinterpreted and lead to people using it in different incompatible ways. Something like organisations_displayed_first is easier to change to something more broad, while the reverse is not true. Thoughts?

Sounds like we need some user research.

MatMoore:
How about ~~"Boaty McBoatface"~~ "emphasised organisations".

It implies the distinction is purely presentational, but isn't overly specific, and allows us to change the things on the frontend without changing the publishers. I don't think anyone would infer anything about the structure of government from this.

I think the key improvement is moving it to the details hash. The links hash should describe semantics of how things relate to each other, and lead organisation doesn't have a single meaning at the moment. 👍

fofr:
Please try out options with users.

boffbowsh:
This is an internal name @fofr, not for displaying in admin UIs

boffbowsh:
`emphasised_organisations` also works well imo

MatMoore:
Just thinking a little bit more about this -

I'm assuming this means content tagger will have to know about `details`, which it doesn't already? Is this correct? Could we do implement this in such a way that works for all link types?

Regarding user research: I'm aware there has been some research already on publishing workflow, so that may be relevant to helping ensure editors understand the effects of tagging.

I think our (my?) assumptions right now are:
* We're not aware of a user need to order things in an arbitrary way at publishing time.
* The lead/supporting distinction arose because the way we present long lists of organisations looks very odd if you show all of them alphabetically, e.g. the 2012 olympics had lots of organisations tagged to it that people are unlikely to associate with the olympics
* We don't know if there is any end user benefit to this
* We don't have a lot of evidence these organisation links are even useful to end users at all...

boffbowsh:
Content Tagger won't need to know about `details`, only the Publishing apps (mainly Whitehall) would need to know about that in order to add the organisations in the correct order.

MatMoore:
Ah, right forgot organisations is not in content tagger.

tijmenb:
`organisations` _is_ in content-tagger, but setting the `emphasised_organisations` is not its responsibility.

To recap (a consensus):

- We won't change the UIs
- We use `emphasised_organisations` in details hash to allow the frontend to display the things in a certain way
- We will use this pattern for policies (policy-publisher) and whitehall content.

Makes sense?

boffbowsh:
👍  to summary

jackscotti:
👍  thank you all. Updating.

fofr:
> This is an internal name @fofr, not for displaying in admin UIs
Internal names have a habit of making themselves visible, usually for the sake of consistency ;)

jackscotti:
Renamed and added deprecation messages.

fofr:
Should `lead_organisations` still be a required field on some of the formats?
eg https://github.com/alphagov/govuk-content-schemas/blob/master/formats/detailed_guide/publisher/links.json#L6

PR #307, opened 2016-05-12
Add `emphasised_organisations` to policy format

tijmenb:
This uses `lead_organisations` in the same way as Whitehall content.

Follow up to https://github.com/alphagov/govuk-content-schemas/pull/306.


jackscotti:
👍

PR #308, opened 2016-05-13
Add `emphasised_organisations` example

tijmenb:
This example is needed to get government-frontend tested with the new way of specifying emphasised organisations.

Related: https://github.com/alphagov/govuk-content-schemas/pull/306

Trello: https://trello.com/c/us3MI1n9/602-fix-organisation-tagging

### DIFF
--- a/dist/formats/case_study/publisher/schema.json
+++ b/dist/formats/case_study/publisher/schema.json
@@ -96,7 +96,7 @@
       ],
       "properties": {
         "lead_organisations": {
-          "description": "A subset of organisations that should be emphasised in relation to this content item. All organisations specified here should also be part of the organisations array.",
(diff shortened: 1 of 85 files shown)


fofr:
Will the other examples that reference `supporting_organisations` and `lead_organisations` be updated too? (Not necessarily in this PR)

eg:
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/case_study/frontend/examples/archived.json#L38-L46
https://github.com/alphagov/govuk-content-schemas/blob/example-for-emphasised/formats/detailed_guide/frontend/examples/detailed_guide.json#L42-L51
etc.

The examples should remain an accurate reflection for the frontends.

tijmenb:
Yup, good point - I was planning on doing that when we remove `lead_organisations` and `supporting_organisations`, but doing this now makes more sense. I'll update this PR.

MatMoore:
👍  LGTM

PR #309, opened 2016-05-16
Only allow unique content_ids in links

tijmenb:
publishing-api now silently discards duplicates (alphagov/publishing-api#303). This should be reflected in the content-schemas.

/cc @boffbowsh


Davidslv:
👍

PR #310, opened 2016-05-17
Fix some schema inconsistencies

tijmenb:
This fixes some inconsistencies in the schemas. See individual commits for details.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -6,6 +6,7 @@
     "base_path",
     "format",
     "title",
+    "details",
     "locale",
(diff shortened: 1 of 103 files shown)


MatMoore:
👍

benhyland:
As a human, there's no way to sensibly review 103 files. tests pass so it must be good.

boffbowsh:
This is partly true, however in this case Tijmen split the PR into 4 commits, each of which are reviewable individually. We also do not need to review the changes to files within `dist` as these are auto generated by rake tasks and committed with the sources for convenience to other developers.

This leaves us with only 5 files to review, which is definitely achievable. 

👍

PR #312, opened 2016-05-18
Add document_collection schema

fofr:
Details:
* Body is optional for collections
* First public at is required like other editioned content
* A document collection must have an ordered list of groups
  * Each group must have a title and an ordered list of documents, the group may have some body copy
* Collections can be political
* Collections can be withdrawn
* Include tags – these are not really used except for emails, see #292 for why they weren't removed from other formats

Links:
* Add documents GUID list, make this a required field
* Whitehall marks "policy areas" as being required in the editor, but content schemas say this is "largely deprecated", so it has been intentionally left out. (Policy areas are included by base links)
* Add topical events
* Lead organisations are also required, we've changed the pattern for organisations to "organisations" in links and emphasised organisations in details – there isn't a pattern yet for which bits must be required
to match current behaviour.

Include examples based on:
https://www.gov.uk/government/collections/national-driving-and-riding-standards
https://www.gov.uk/government/collections/financial-sanctions-regime-specific-consolidated-lists-and-releases

https://trello.com/c/PMeMQb5E/410-11-document-collection-migration-mvp-content-schema-examples-and-front-end-work

### DIFF
--- a/formats/publication/frontend/examples/political_publication.json
+++ b/formats/publication/frontend/examples/political_publication.json
@@ -71,16 +71,6 @@
         "locale": "en"
       }
     ],
-    "document_collections": [
-      {
-        "content_id": "5eb67fbf-7631-11e4-a3cb-005056011aef",
-        "title": "Climate Change Agreements guidance",
-        "base_path": "/government/collections/climate-change-agreements-guidance",
-        "api_url": "https://www.gov.uk/api/content/government/collections/climate-change-agreements-guidance",
-        "web_url": "https://www.gov.uk/government/collections/climate-change-agreements-guidance",
-        "locale": "en"
-      }
-    ],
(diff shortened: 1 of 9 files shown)


fofr:
Government frontend WIP to render this format:
https://github.com/alphagov/government-frontend/compare/mvp-document-collections

fofr:
Closing while adding more examples.

formats/publication/frontend/examples/political_publication.json
```diff
@@ -71,16 +71,6 @@
         "locale": "en"
       }
     ],
-    "document_collections": [
```
boffbowsh: Why is this removed?

fofr:
Test failures are unrelated.

PR #314, opened 2016-05-20
[Do not merge yet] Make withdrawn notice top level

steventux:
Part of moving Whitehall to use unpublish endpoints. https://github.com/alphagov/whitehall/pull/2596

https://trello.com/c/78Q9mrDe/708-make-whitehall-use-unpublishing-endpoint-medium


steventux:
Closing, this will eventually need to be done but we will transition content first.
see https://github.com/alphagov/govuk-content-schemas/pull/315

PR #316, opened 2016-05-23
Add fatality notice format

nickcolley:
Details:
* Body is required
* First public at is required
* Fatality Notices can be Political and uses history mode

Links:
* Fields of operation are required with at least one item,
small dataset so mostly going to be one.
* Format uses policy areas which is provided by base links
* Organisations are not required and are provided by base links

Whitehall has 'Roll call info' on the fatality notice model
but this is used by fields of operations pages
and not in the format itself.

The example used here is fake due to the sensitive nature of the copy.

Note that Withdraw notices feature is omitted intentionally, will update with a separate PR.

Paired with @fofr 🙊 

Live example:
https://www.gov.uk/government/fatalities/flight-lieutenants-alan-scott-and-geraint-roberts-have-died

Associated ticket: https://trello.com/c/0UAMj5gj/407-8-fatality-notice-migration-implement-publishing-of-format-to-publishing-api

### DIFF
--- a/formats/fatality_notice/frontend/examples/fatality_notice.json
+++ b/formats/fatality_notice/frontend/examples/fatality_notice.json
@@ -1,0 +1,67 @@
+{
+  "base_path": "/government/fatalities/sir-george-pomeroy-colley-killed-in-boer-war",
+  "content_id": "8537348f-c0f1-11e4-8234-005056011aef",
+  "title": "Sir George Pomeroy Colley killed in Boer War",
+  "description": "It is with great sadness that the Ministry of Defense must confirm that Sir George Pomeroy Colley, died in battle in Zululand on 27 February 1881.",
+  "format": "fatality_notice",
+  "need_ids": [],
+  "locale": "en",
+  "updated_at": "1881-03-23T14:00:00.023Z",
+  "public_updated_at": "1881-05-17T14:00:00.023Z",
+  "details": {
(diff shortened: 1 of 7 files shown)


formats/fatality_notice/publisher/links.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "fields_of_operation"
+  ],
+  "properties": {
+    "fields_of_operation" : {
+      "minItems": 1,
```
boffbowsh: This should probably also have a `"maxItems": 1`

fofr: @boffbowsh Reckon it should be named `field_of_operation` even though it's an array?

tvararu:
> The example used here is fake due to the sensitive nature of the copy.

I agree this is likely necessary. That being said, all of our other examples are based on real items. I quite like the fact that you can just copy and paste the base path and plop it into integration/staging/production and have it work.

It's a shame that JSON doesn't allow for comments, but it would be nice to communicate in some way that is close to that `base_path` attribute that it won't work on live, to prevent confusions. The only two ways I can see are by naming the base path something quite obviously fake, but better I think would be a JSON `"//comment": "Note that this content item is fictional and the base_path doesn't map to a real content item on production."` key value pair, which you begrudgingly suggested in our offline chat about this, @nickcolley. It's not the cleanest but I can imagine other places in our examples where this might be useful.

nickcolley:
@tvararu did a bit of digging and found that a top level 'description' property is probably what we want?

http://json-schema.org/latest/json-schema-validation.html#anchor98

"A title will preferrably be short, whereas a description will provide explanation about the purpose of the instance described by this schema."

formats/fatality_notice/publisher/links.json
boffbowsh: Yep sounds good. Similar to `parent`

tvararu:
That's for the actual `schema` specifications, and it's [in use](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json#L9). The example AFAIK is just a JSON compliant document and doesn't comply to the `json-schema` patterns.

boffbowsh:
> Whitehall has 'Roll call info' on the fatality notice model
but this is used by fields of operations pages
and not in the format itself.

This is true, however the Field of Operation will eventually use the roll call info from linked Fatality Notices to populate that page. So maybe they should go in now for completeness?

fofr:
@boffbowsh I'd include the roll call when we need it/want to use it. It _will_ be noticed when migrating fields of operation.

boffbowsh:
Aye fair enough

formats/fatality_notice/frontend/examples/fatality_notice.json
```diff
@@ -0,0 +1,67 @@
...
+        "title": "William Ewart Gladstone",
+        "base_path": "/government/people/william-ewart-gladstone",
+        "api_url": "https://www.gov.uk/api/content/government/people/william-ewart-gladstone",
+        "web_url": "https://www.gov.uk/government/people/william-ewart-gladstone",
+        "locale": "en"
+      }
+    ],
+    "field_of_operation": [
+      {
+        "content_id": "8537348f-c0f1-11e4-9312-005056011aef",
```
boffbowsh: This should probably be a different content id to the minister to avoid confusion

fofr: They are different.

PR #317, opened 2016-05-24
Make lead_organisations optional

jackscotti:
We're getting rid of lead and supporting organisations.
This commit is to make sure that they're optional in our schemas.

Another piece of work to completely get rid of them will come as soon as publishing applications will have been updated to stop publishing those types of organisation.

ticket: https://trello.com/c/us3MI1n9/602-fix-organisation-tagging

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -76,9 +76,6 @@
     "links": {
       "type": "object",
       "additionalProperties": false,
-      "required": [
-        "lead_organisations"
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+A TEST
```
boffbowsh: ?

jackscotti: I wanted to restart the Jenkins tests and added a test commit, please ignore.

fofr:
:+1:

PR #319, opened 2016-05-31
Make public_updated_at an optional field for all formats

tuzz:
If the public_updated_at field is not provided when content
is submitted to the publishing api then it will default it
to Time.now for major updates and raise an error for minor
updates. We should never publish something for the first
time with a minor update so this field is always present.

Because this piece of workflow is now a part of the
publishing api it means our publishing apps should no
longer be required to send this field and they can
optionally leave it out and still pass their schema tests.

This field can still be sent (optionally) if the
publishing app wishes to set a custom public_updated_at.
This is a requirement for some of our publishing apps,
e.g. Whitehall.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -8,7 +8,6 @@
     "title",
     "details",
     "locale",
-    "public_updated_at",
     "content_id",
     "document_type",
     "schema_name"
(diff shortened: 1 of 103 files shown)


benilovj:
👍

danielroseman:
Seems sensible.

boffbowsh:
👍

PR #320, opened 2016-05-31
Restrict usage of deprecated `lead_organisations`, remove `supporting_organisations`

tijmenb:
`lead_organisations` is deprecated and should not be used anymore. This PR makes sure that it's only allowed in the Policy format, where were currently removing it.


MatMoore:
👍  looks good to me but needs rebasing.

For context, we now store only `organisations` in the links hash. There is now a `emphasised_organisations` value in the details hash.

tijmenb:
@MatMoore rebased!

MatMoore:
👍

PR #325, opened 2016-06-14
Rename service manual links topics

deborahchua:
'topics' is already a reserved field and is being renamed to 'service_manual_topics'.


tijmenb:
👍

PR #326, opened 2016-06-15
Add top & second level browse pages to details hash

Davidslv:
We are adding top level and second level browse pages into the details hash because we need them to be able to be ordered when the content is curated in collections publisher.

In the past we made a decision of not preserving the order for links, which can be found here: https://gov-uk.atlassian.net/wiki/display/FS/RFC+36%3A+Stop+preserving+order+for+links

One of the key reasons for this decision was to make sure presentation concerns don't get put in the links hash. That's why we are using details instead: neither of these fields are needed to describe how the browse pages link together (we have active_top_level for that).

There will be following commits to remove these two fields from the links hash once we have fixed the frontend apps.

Trello:
https://trello.com/c/KZJigx96/676-ordering-of-browse-pages-no-longer-respected

@davidslv & @MatMoore

### DIFF
--- a/dist/formats/mainstream_browse_page/frontend/schema.json
+++ b/dist/formats/mainstream_browse_page/frontend/schema.json
@@ -204,6 +204,14 @@
           "type": "string",
           "description": "An internal name for admin interfaces. Includes parent."
         },
+        "top_level_browse_pages": {
+          "description": "All top-level browse pages",
+          "$ref": "#/definitions/guid_list"
+        },
+        "second_level_browse_pages": {
+          "description": "All 2nd level browse pages under active_top_level_browse_page",
+          "$ref": "#/definitions/guid_list"
+        },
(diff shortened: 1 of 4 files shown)


tijmenb:
> There will be following commits to remove these two fields from the links hash once we have fixed the frontend apps.

Wouldn't you still need them, because otherwise the frontend would just have a bunch of UUIDs?

MatMoore:
@tijmenb I'm not sure how to solve it, but it seems to be really against the spirit of the RFC to keep these in the links hash, as the only information they give you deals with presentation concerns.

* All browse pages get the same data for top_level_browse_pages
* The second_level_browse_pages of a top level browse page is just the inverse of the active_top_level relationship
* The second_level_browse_pages of a second level browse page tells you its siblings, which are also implied by active_top_level links.

PR #327, opened 2016-06-16
Add finder_email_signup_schema

Seraphiana:
We've retroactively added a schema for finder_email_signup,
which are published via a rake task on specialist-publisher
[Trello Card](https://trello.com/c/HnpJ2PQf/147-create-missing-schema-for-finder-email-signup-medium)

### DIFF
--- a/formats/finder_email_signup/publisher/details.json
+++ b/formats/finder_email_signup/publisher/details.json
@@ -1,0 +1,69 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "email_signup_choice",
+    "email_filter_by",
+    "subscription_list_title_prefix"
+  ],
+  "properties": {
+    "beta": {
+      "type": "boolean"
+    },
+    "email_signup_choice": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "key",
+          "radio_button_name",
+          "topic_name",
+          "prechecked"
+        ],
+        "properties": {
+          "key": {
+            "type": "string"
+          },
+          "radio_button_name": {
+            "type": "string"
+          },
+          "topic_name": {
+            "type": "string"
+          },
+          "prechecked": {
+            "type": "boolean"
+          }
+        }
+      }
+    },
+    "email_filter_by": {
+      "oneOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ]
+    },
+    "subscription_list_title_prefix": {
+      "oneOf": [
+        {
+          "type": "object",
+          "properties": {
+            "plural": {
+              "type": "string"
+            },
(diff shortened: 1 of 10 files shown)


tuzz:
Looks good. I left a comment here that might be worth considering before we merge this.

https://github.com/alphagov/specialist-publisher-rebuild/pull/803#discussion-diff-67505664

Thanks 👍

Rosa-Fox:
Thanks @tuzz. We have made the changes to `specialist-publisher-rebuild`. It was useful as we realised we needed to make a couple of changes in this PR to make the schemas less strict (for example some SPV2 schema fields required strings where as others required object (for example `"email_filter_by"`) so we removed the requirement of a `type`.

formats/finder_email_signup/publisher/details.json
```diff
@@ -0,0 +1,51 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+   "required": [
+     "details"
+   ],
+  "properties": {
+     "details": {
```
danielroseman: The details schema shouldn't itself contain details; it is included as the details key in the parent.

formats/finder_email_signup/publisher/details.json
```diff
@@ -0,0 +1,51 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "beta": {
+          "type": "boolean"
+        },
+        "email_signup_choice": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "$ref": "#/definitions/email_signup_choice"
```
danielroseman: This doesn't need to be a ref to a separate definition, since it's only used in one place. Move the definition here.

formats/finder_email_signup/publisher/links.json
```diff
@@ -0,0 +1,169 @@
+{
```
danielroseman: This schema doesn't seem right at all. We should only be passing content_ids, not details - those are expanded by publishing-api itself. And it looks like the only keys actually needed in the links are oragnisations and related, both of which are just refs to guid_list which is already defined.

tuzz:
Ahh, I should have looked more closely. Let me know if you need any help to fix these issues @Rosa-Fox @Seraphiana

Rosa-Fox:
Thanks @danielroseman & @tuzz, I have made the changes. I am not 100% sure what the `required_fields` here should be, are [these](https://github.com/alphagov/govuk-content-schemas/blob/55210de6495c5db8a9bf42a292fb7a892c15f659/formats/finder_email_signup/publisher/details.json#L5) right?

tuzz:
@danielroseman please could you take a look now that we've updated this ticket. Thanks!

tuzz:
@dougdroper, @elliotcm, @cbaines: I think Daniel is probably busy. Please could one of you take a look now we have updated this pull request. We're trying to clear our board a bit.

Thanks

PR #329, opened 2016-06-17
Adapt contacts to whitehall model

steventux:
https://trello.com/c/JxKRVn6c/731-export-whitehall-contacts-to-publishing-api-medium

Whitehall contacts are pretty similar to those from the contacts app, a couple of fields are needed to accommodate Whitehall contacts data.
The significant change here is to allow formats to provide custom metadata, this feels a bit strange as metadata should be common across all formats, however the publishing API allows pathless content for specific formats and I've attempted to provide a mechanism for this here. Perhaps there's a better solution?
The description field is no a longer required field, [this is enforced by the contacts-admin application](https://github.com/alphagov/contacts-admin/blob/master/app/models/contact.rb#L28) as it's not necessary for Whitehall.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,11 +36,16 @@ In the future, there may be a third:
 
 The 'publisher' [`schema.json`](dist/formats/case_study/publisher/schema.json) is built from three parts:
 
-  - [`metadata.json`](formats/metadata.json): the top level set of fields. These are the **same for every content
-    format**.
+  - [`metadata.json`](formats/metadata.json): the top level set of fields. These are **common to most content
+    formats**.
 
   - [version metadata eg. `v2_metadata.json`](formats/v2_metadata.json): Fields specific to the publisher version.
     Either [`v1_metadata.json`](formats/v1_metadata.json) or [`v2_metadata.json`](formats/v2_metadata.json)
+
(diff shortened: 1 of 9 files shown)


danielroseman:
How does this deal with the fact that Whitehall contacts don't have a base_path, since they never display on an independent page?

steventux:
@danielroseman I'd not considered this as I _thought_ we already allowed pathless content items to be created in the publishing-api. Perhaps that was the wrong assumption. `base_path` is still required for v2 schemas, could/should this be relaxed on a specific format basis?

danielroseman:
We do allow it in publishing-api, by adding formats to `ContentItem::EMPTY_BASE_PATH_FORMATS`, but the schemas still also need to support it otherwise they won't validate.

steventux:
@danielroseman I've added a way to provide format-specific metadata to allow valid pathless content.

danielroseman:
Looks good. It would be useful to have a test that covers this in spec/tasks/combine_schema_spec.

Also if possible can you split the actual changes to the contact schema into a separate commit.

danielroseman:
:+1: but one more thing, sorry - the README needs to be updated to include this.

steventux:
Thanks @danielroseman, this definitely needs documenting :+1: updated.

PR #333, opened 2016-06-23
Add a message queue schema

cbaines:
Trello ticket: https://trello.com/c/teMFb0iD/734-content-schemas-for-content-on-message-queue

The publishing api sends JSON documents out on a "message queue". These JSON
documents include a representation of a content item. This new schema aims to
describe these JSON documents, as they are generated currently by the
publishing api.

The schema should accept any current output of the publishing api, but is
incomplete in many areas (e.g. loose validation for expanded_links), and so is
not very useful for consumers of the message queue in terms of documentation of
what can be expected.

Originally I did set out to make a schema per format as is done for the
publisher and frontend schemas, but ran in to issues with the behavior of the
schema combiner, particularly with regard to the mutual exclusivity of the
format and document_type + schema_name fields as the message queue payload
includes all three. As a result of this, I switched to building a schema to
match the message queue payload, for any format.

### DIFF
--- a/formats/message_queue_base.json
+++ b/formats/message_queue_base.json
@@ -1,0 +1,148 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "definitions": {
+    "formats": {
+      "title": "This is a placeholder, and filled when the schema is built"
+    }
+  },
+  "allOf": [
+    {
+      "additionalProperties": false,
+      "required": [
+        "update_type",
+        "analytics_identifier",
+        "content_id",
+        "description",
+        "details",
+        "document_type",
+        "first_published_at",
+        "need_ids",
+        "phase",
+        "public_updated_at",
+        "publishing_app",
+        "redirects",
+        "rendering_app",
+        "routes",
+        "schema_name",
+        "title",
+        "base_path",
+        "locale",
+        "format"
+      ],
+      "properties": {
+        "update_type": {
+          "enum": [ "major", "minor", "republish", "links" ]
+        },
+        "content_id": {
+          "$ref": "#/definitions/guid"
+        },
+        "analytics_identifier": {
+          "$ref": "#/definitions/analytics_identifier"
+        },
+        "links": {
+          "$ref": "#/definitions/links"
+        },
+        "expanded_links": {
+          "type": "object"
+        },
+        "base_path": {
+          "$ref" : "#/definitions/absolute_path"
+        },
+        "title": {
+          "type": "string"
+        },
+        "description": {
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ]
+        },
+        "public_updated_at": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "first_published_at": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "publishing_app": {
+          "type": "string"
+        },
+        "rendering_app": {
+          "type": "string"
+        },
+        "locale": {
+          "type": "string"
+        },
+        "need_ids": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "title": "This is intended to be used by front end applications to present pages to users to show how effectively users' needs are being met"
+          },
+          "title": "Identifiers of user needs"
+        },
+        "routes": {
+          "type": "array",
+          "minItems": 1,
+          "items": {
+            "$ref": "#/definitions/route"
+          }
+        },
+        "redirects": {
+          "type": "array",
+          "items": {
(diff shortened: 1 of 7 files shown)


formats/message_queue_base.json
```diff
@@ -0,0 +1,148 @@
...
+            }
+          }
+        },
+        "phase": {
+          "description": "The service design phase of this content item - https://www.gov.uk/service-manual/phases",
+          "type": "string",
+          "enum": [ "alpha", "beta", "live" ]
+        },
+        "details": {
+          "type": "object"
```
steventux: Should this be the same as https://github.com/alphagov/govuk-content-schemas/blob/master/formats/metadata.json#L88

cbaines: The details property is validated using the details schemas, but this is not done here.

The 2nd element of the `allOf` which contains this line, references `#/definitions/formats` which is generated through running rake. This schema performs this part of the validation.

It possibly makes more sense to look at the generated file: https://github.com/alphagov/govuk-content-schemas/blob/25563d16686766f2b2f92482e1db0957a07e401c/dist/message_queue.json#L3327

The `formats` schema referenced above constrains the `schema_name` and `format` field values, as well as validating the `details` field accordingly (for some formats, this is skipped if a schema is not available).

This line is still essential here, as its required for `additionalProperties: false` and the `required` section to work.

steventux: Yes, I realised [as I looked at the generator](https://github.com/alphagov/govuk-content-schemas/pull/333/files/df2cba8d98c4dd4234014c7b645f3251edc71414#diff-07d491c55fca3fc137b1fcc036090840R41) that it was academic what the type was, ignore that. :+1:

danielroseman:
👍 , needs a rebase though.
It would be good to have some tests for the new combining functionality.

cbaines:
> :+1: , needs a rebase though.

Great, do you want me to rebase, and force push to this branch?

> It would be good to have some tests for the new combining functionality.

I'm not sure how tests would fit in to the code changes, especially as I think all the code is currently just run by humans? Let me know if you want to discuss this more?

danielroseman:
Yes, generally feel free to rebase and force push.

There are a bunch of tests for the different schema combiner tasks in eg spec/tasks and for the lib classes in spec/lib, I think the new code here could be tested there too.

cbaines:
> There are a bunch of tests for the different schema combiner tasks in eg spec/tasks and for the lib classes in spec/lib, I think the new code here could be tested there too.

I have not thought about this much, but I think it would be more valuable to cover the task code indirectly, by having more example inputs, including invalid inputs. I think this would cover the task code indirectly, and give coverage of the input space, which is more valuable. What do you think?

cbaines:
Discussed offline, I'll look at implementing some tests tomorrow.

cbaines:
I've now added tests, rebased on the current master, and re-run the Jenkins jobs until they passed all passed.

spec/tasks/combine_schema_spec.rb
```diff
@@ -168,4 +169,48 @@
...
   end
+
+  context "message queue schema" do
+    let(:output_filename) { File.join("dist", "message_queue.json") }
+    let(:details) { build_schema('details.json', properties: build_string_properties('detail')) }
+    let(:publisher_schema_dir) { project_root.join("formats", format_name, "publisher") }
+
+    around do |example|
+      format_root = ENV["FORMAT_ROOT"]
+      FileUtils.mv("dist/message_queue.json", "dist/message_queue.json.bkp")
```
danielroseman: This seems a bit fragile: is there a possibility that the spec could crash halfway through, leaving the files in the wrong place?

spec/tasks/combine_schema_spec.rb
```diff
@@ -168,4 +169,48 @@
...
+        File.write("#{test_root}/format_foo/publisher/details.json", "{}")
+
+        example.run
+      end
+
+      FileUtils.mv("dist/message_queue.json.bkp", "dist/message_queue.json")
+      ENV["FORMAT_ROOT"] = format_root
+    end
+
+    specify "the frontend schema file contains a frontend schema" do
```
danielroseman: I'm finding it really hard to tell what this is actually testing. Why frontend schemas? Is the point that we're asserting that this is called for format_foo but not format_bar? It would be useful to be more explicit.

spec/tasks/combine_schema_spec.rb
cbaines: If it fails, including errors, the around still continues to execute. I don't like this either, but I'm not sure how to do it better.

spec/tasks/combine_schema_spec.rb
cbaines: I've fixed the message, it was just wrong.

PR #346, opened 2016-07-12
Remove service manual guide summary

thehenster:
We can use the same text for the meta description and the point text.

Because we need to optionally display the description, either for the normal guide page or the point page, the boolean show_description will be used to turn it off and on.


36degrees:
LGTM :+1:

PR #347, opened 2016-07-13
Update message queue schemas

tijmenb:
See individual commits for details.


cbaines:
Looks great :) I hope this means that you have got your sample data generator working?

tijmenb:
No not yet, but this should make it easier.

PR #348, opened 2016-07-13
[Don't merge] Remove `format`

tijmenb:
We can get rid of a lot of complexity and weird error messages if we complete the transition from `format` to `document_type` / `schema_name`. 

This PR to see which apps are still using `format`.


tijmenb:
Closing until https://github.com/alphagov/govuk-content-schemas/pull/353 is merged.

PR #349, opened 2016-07-13
Allow document_type & schema_name for gone

tijmenb:
There are a couple of apps still sending `format` to the publishing-api. We've created PRs for these to send `document_type` and `schema_name` but these can't be merged until those attributes are allowed everywhere.

After those PRs are merged `format` will be removed here, and `document_type` & `schema_name` will become required.


cbaines:
:+1:

PR #351, opened 2016-07-14
Use document_type & schema_name in examples

tijmenb:
- Add document_type & schema_name to all examples
- Remove `format` from publisher examples. Frontends still rely on `format` being present, so we need to keep that attribute in the schema.

### DIFF
--- a/formats/specialist_document/publisher_v2/examples/specialist_document.json
+++ b/formats/specialist_document/publisher_v2/examples/specialist_document.json
@@ -2,7 +2,6 @@
   "base_path": "/cma-cases/example-cma-case",
   "title": "Example CMA Case",
   "description": "This is the summary of an example CMA case",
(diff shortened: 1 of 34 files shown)


formats/specialist_document/publisher_v2/examples/specialist_document.json
```diff
@@ -36,24 +36,32 @@
...
     {
       "path": "/cma-cases/example-cma-case",
       "type": "exact"
     }
   ],
-  "redirects": []
+  "redirects": [
+
+  ],
+  "document_type": "specialist_document",
```
danielroseman: This should be "cma_case".

tijmenb: Thanks, updated.

PR #353, opened 2016-07-14
Remove `format`, make `document_type` required

tijmenb:
This makes `document_type` & `schema_name` required and removes the last `format` fields from the handmade schemas. This is possible now that no apps send `format` to the publishing-api.

### DIFF
--- a/dist/formats/gone/publisher/schema.json
+++ b/dist/formats/gone/publisher/schema.json
@@ -3,9 +3,11 @@
   "type": "object",
   "additionalProperties": false,
   "required": [
+    "document_type",
     "publishing_app",
-    "update_type",
-    "routes"
+    "routes",
+    "schema_name",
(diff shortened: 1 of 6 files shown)


tijmenb:
cc @danielroseman

tijmenb:
Passes after https://github.com/alphagov/govuk-content-schemas/pull/351 is merged.

danielroseman:
Presumably we'll need to deploy all those apps before we can merge this? Otherwise they'll all start failing validation in publishing-api.

tijmenb:
That raises an interesting question. If something is green, we should be able to merge it. Should we actually do downstream tests against the `deployed-to-production` branch? That way incompatible schema changes can't be merged until app fixes are actually released.

Davidslv:
@tijmenb seems this needs to be updated.

PR #355, opened 2016-07-14
Add taxon parents link to taxon

carvil:
We would like to allow taxons to have multiple taxon parents associated. Rather than changing the existing `parent` link (which is used for the breadcrumbs and only allows one association), we created a new link type called `taxon_parents`. This should allow a number of parents on any given taxon.

More details here: https://trello.com/c/nImPlcNS/19-allow-taxons-to-have-multiple-parents


tijmenb:
Excellent!

PR #358, opened 2016-07-15
Add 'temporary_update_type' to specialist documents

tuzz:
When adding an attachment to a document, it sends the
document to the Publishing API with the attachment's
metadata. If the user hasn't chosen an update_type for
the document, we intend to set it to 'minor' and set
this 'temporary_update_type' field to true. We can then
present a choice of update types to the user rather
than pre-selecting 'minor' for them.

Relates to:
https://trello.com/c/oxxALgmy/187-changes-to-update-type-for-raib-reports

### DIFF
--- a/dist/formats/specialist_document/frontend/schema.json
+++ b/dist/formats/specialist_document/frontend/schema.json
@@ -237,6 +237,10 @@
         },
         "change_history": {
           "$ref": "#/definitions/change_history"
+        },
+        "temporary_update_type": {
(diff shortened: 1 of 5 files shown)


steventux:
I assume it's not as simple as enforcing `update_type` validation somewhere within specialist publisher?
:+1: other than that

tuzz:
Afraid not! Nothing's ever that simple :-(

PR #361, opened 2016-07-19
Add policy areas & related policies to detailed guides

binaryberry:
This is required as part of migrating detailed guides.

[Trello](https://trello.com/c/CkHu0suq/317-6-detailed-guides-migration-final-tasks-deploy-1-medium)

### DIFF
--- a/formats/detailed_guide/publisher/links.json
+++ b/formats/detailed_guide/publisher/links.json
@@ -9,6 +9,9 @@
     "related_guides": {
       "$ref": "#/definitions/guid_list"
     },
+    "related_policies": {
+      "$ref": "#/definitions/guid_list"
+    },
     "related_mainstream": {
       "$ref": "#/definitions/guid_list"
     }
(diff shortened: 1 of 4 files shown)


formats/detailed_guide/publisher/links.json
```diff
@@ -9,6 +9,12 @@
     "related_guides": {
       "$ref": "#/definitions/guid_list"
     },
+    "related_policies": {
+      "$ref": "#/definitions/guid_list"
+    },
+    "policy_areas": {
```
tijmenb: `policy_areas` is already part of the allowed links for every format, so it's not necessary to add this here.

https://github.com/alphagov/govuk-content-schemas/blob/master/formats/base_links.json#L31-L34

binaryberry: True, amended now. Thanks Tijmen!

tijmenb: Cool thanks!

PR #364, opened 2016-07-21
Restrict `locale` to list of valid locales

tijmenb:
It's currently possible to provide a bogus locale, which would cause frontends to crash because they set the Rails locale to this string (`I18n.locale`). See [government-frontend](https://github.com/alphagov/government-frontend/blob/e9de1144c40b9a2231904218656d0812f22ebce5/app/controllers/content_items_controller.rb#L41-L43) for an example of this behaviour.

This list was generated by running this on government-frontend rails console:

```
I18n.available_locales.map(&:to_s).sort
```


cbaines:
:+1:

tijmenb:
Follow up to this: https://github.com/alphagov/govuk-content-schemas/pull/366

PR #365, opened 2016-07-21
Make `links` required in frontends

tijmenb:
content-store & publishing-api always return a `links` key, so apps depend on this. Making it required will ensure this will always be the case and allows random content generation.


rubenarakelyan:
👍

PR #366, opened 2016-07-21
Also validate locale in frontend schemas

tijmenb:
When doing https://github.com/alphagov/govuk-content-schemas/pull/364 I didn't realise `locale` was also used in the frontend-links. 

This PR extracts the field definition to the `definitions` and uses it to validate the frontend links.


Davidslv:
This PR is a browser killer

PR #368, opened 2016-07-25
[Do Not Merge] Add note and reason for change for latest change to service manual guides

36degrees:
Because we don't know the timestamp of when we'll publish until we call publish, we use change_history only for previous changes, and get the latest change note and summary directly from the details hash, along with public_updated_at.


36degrees:
We're going with a different approach so this change is no longer required.

PR #372, opened 2016-08-16
Add `gone` schema and examples

gpeng:
This format is required as `gone` content will be rendered by the frontend. This commit updates the schema to allow unpublishing explanations and alternative urls to be displayed on `gone` items.


formats/gone/frontend/schema.json
```diff
@@ -0,0 +1,91 @@
...
+    "routes"
+  ],
+  "properties": {
+    "base_path": {
+      "$ref": "#/definitions/absolute_path"
+    },
+    "content_id": {
+      "$ref": "#/definitions/guid"
+    },
+    "format": {
```
tijmenb: Can we skip this one here? :)

gpeng: Yep. Fixed in #4886448

PR #373, opened 2016-08-16
Allow email alert signups to work with links other than countries

36degrees:
The email-alert-api uses the links hash within subscriber_lists as a matcher, examining new documents that get published for matching `links` and `document` types and sending emails to relevant lists.

The schema currently limits the contents of the links hash to only permit countries, which means that only links to `countries` can be subscribed to.

We are building email alerts into the service manual, allowing users to subscribe to service manual topics. Therefore we need to create email_alert_signups based on documents that link to `service_manual_topics`, so we need to allow for this in the schema.

Rather than adding `service_manual_topics` as another 'special case' this instead allows exactly one property with a alphabetical, underscored key to be provided as long as the value is an array.

### DIFF
--- a/formats/service_manual_guide/frontend/examples/service_manual_guide.json
+++ b/formats/service_manual_guide/frontend/examples/service_manual_guide.json
@@ -11,6 +11,7 @@
         "href": "#what-is-it-why-it-works-and-how-to-do-it"
       }
     ],
+    "change_note": "This is our latest change.",
     "change_history": [
       {
         "public_timestamp": "2015-10-07T08:17:10+00:00",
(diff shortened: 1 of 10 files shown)


formats/service_manual_guide/frontend/examples/service_manual_guide.json
```diff
@@ -11,6 +11,7 @@
         "href": "#what-is-it-why-it-works-and-how-to-do-it"
       }
     ],
+    "change_note": "This is our latest change.\n\nThis is the reason for our latest change.",
```
thehenster: I wonder how much we need the reason for change in here? It's minorly upsetting that the `change_note` field contains the change_note and reason_for_change, but doesn't in the `change_history`.

Considering the actual email isn't the most readable at the moment it might actually improve it and the user can always click through to the frontend to see the reason_for_change.

Already been down this path?

36degrees: I've updated the examples to remove references to reason for change, and modified publisher to just send the change note.

thehenster:
LGTM

PR #377, opened 2016-08-24
Use link expansion for the Service Standard

thehenster:
This PR updates the `service_manual_service_standard` schema and examples to use link expansion.

Currently it doesn't seem possible to have a frontend example of a parent/child relationship which has been link-expanded that matches it's frontend schema.

It seems the definitions of each link type are overwritten for the frontend schemas so that they are "#/definitions/frontend_links". https://github.com/alphagov/govuk-content-schemas/blob/b283419dc1f43d69396e88fdd559ea2f4ce1889b/lib/govuk_content_schemas/frontend_schema_generator.rb#L81

The "#/definitions/frontend_links" definition mostly matches an expanded link but falls down when it gets into the links of the expanded link where it expects every link type to be a "#/definitions/guid_list".

Here are a couple of heavily redacted examples to show what we have and what the frontend schema wants. The difference is that the former has it's /links/children/*/links/parent expanded whereas the later has a guid list.

This is the structure of a link expanded content item in the wild:
```
{
  "base_path": "/service-manual/service-standard",
  "links": {
    "children": [
      {
        "base_path": "/service-manual/service-standard/have-a-multidisciplinary-team",
        "links": {
          "parent": [
            {
              "base_path": "/service-manual/service-standard",
              "links": {}
            }
          ]
        }
      }
    ]
  }
}
```

This is the structure of a content item that the schema expects:
```
{
  "base_path": "/service-manual/service-standard",
  "links": {
    "children": [
      {
        "base_path": "/service-manual/service-standard/have-a-multidisciplinary-team",
        "links": {
          "parent": ["00f693d4-866a-4fe6-a8d6-09cd7db8980b"]
        }
      }
    ]
  }
}
```

I'm not sure how much this matters? For parent/child link expanded stuff, the /links/children/*/links/parent data is also available in the root object.. it's a link expanded reference to itself. I think a slight problem is that our examples won't look exactly like they do in the wild, which will cause confusion and people use real content items as the basis for fixutres. A larger potential problem is for another link type where the relationship isn't self-referential like the parent/child one and we actually want that data..

Perhaps the "#/definitions/frontend_links" definition could be made more flexible to allow the link types to be either a guid_list or another "#/definitions/frontend_links". Or a "#/definitions/frontend_expanded_links" maybe?

### DIFF
--- a/dist/formats/service_manual_service_standard/frontend/schema.json
+++ b/dist/formats/service_manual_service_standard/frontend/schema.json
@@ -91,7 +91,7 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "points": {
+        "children": {
           "$ref": "#/definitions/frontend_links"
         },
         "taxons": {
@@ -193,29 +193,9 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "points": {
-          "type": "array",
-          "items": {
-            "type": "object",
-            "properties": {
-              "title": {
-                "type": "string",
-                "description": "The title of the point. It matches the title of the relevant page."
-              },
-              "summary": {
-                "type": "string",
-                "description": "The summary of the point. It matches the summary on the relevant page."
-              },
-              "base_path": {
-                "$ref": "#/definitions/absolute_path"
-              }
-            },
-            "required": [
-              "title",
-              "summary",
-              "base_path"
-            ]
-          }
+        "body": {
+          "type": "string",
+          "description": "A single line of text that serves as a second, less prominent introduction."
         }
       }
     },
--- a/formats/service_manual_service_standard/publisher/details.json
+++ b/formats/service_manual_service_standard/publisher/details.json
@@ -3,29 +3,9 @@
   "type": "object",
   "additionalProperties": false,
   "properties": {
-    "points": {
-      "type": "array",
-      "items": {
-        "type": "object",
-        "properties": {
-          "title": {
-            "type": "string",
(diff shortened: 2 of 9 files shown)


dist/formats/service_manual_service_standard/frontend/schema.json
```diff
@@ -193,29 +193,9 @@
...
-            },
-            "required": [
-              "title",
-              "summary",
-              "base_path"
-            ]
-          }
+        "body": {
+          "type": "string",
+          "description": "A single line of text that serves as a second, less prominent introduction."
```
36degrees: Where was this before?

formats/service_manual_service_standard/publisher/details.json
```diff
@@ -3,29 +3,9 @@
...
-        },
-        "required": [
-          "title",
-          "summary",
-          "base_path"
-        ]
-      }
+    "body": {
+      "type": "string",
+      "description": "A single line of text that serves as a second, less prominent introduction."
```
36degrees: Where was this before?

thehenster: It was missing.. which actually means it wasn't tested.. Hmm. I'll add the test.

36degrees: 👍

thehenster: It looks like we added the test here: https://github.com/alphagov/service-manual-frontend/commit/9e679c59

The test was just missing in master.

PR #378, opened 2016-08-31
Remove oneOf nesting

tijmenb:
This is a follow-up to https://github.com/alphagov/govuk-content-schemas/pull/374. It removes the now-unnecessary nesting of the schema in a `oneOf`.

### DIFF
--- a/lib/govuk_content_schemas/schema_combiner.rb
+++ b/lib/govuk_content_schemas/schema_combiner.rb
@@ -71,19 +71,12 @@ private
   end
 
   def add_format_field(schema)
-    properties = schema.delete('properties')
-    required = schema.delete('required') || []
-    schema['additionalProperties'] = true
-    schema['oneOf'] = [
-      {
-        "properties" => {
-          "document_type" => document_type,
-          "schema_name" => format_with_name
-        }.merge(properties),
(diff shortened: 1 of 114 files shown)


lib/govuk_content_schemas/schema_combiner.rb
```diff
@@ -71,19 +71,12 @@ def merge_schemas(base_schema, other)
...
-        }.merge(properties),
-        "required" => ['document_type', 'schema_name'] + required,
-        "additionalProperties" => false
-      }
-    ]
+    schema["properties"]["document_type"] = document_type
+    schema["properties"]["schema_name"] = format_with_name
+    schema["required"] ||= []
+    schema["required"] << "document_type"
+    schema["required"] << "schema_name"
```
carvil: Maybe it doesn't apply, but do we need to call `uniq` on `schema['required']` after we add both document_type and schema_name? From line 76 it looks like it could have stuff already there, is there a risk document_type and schema_name are already in that array?

carvil:
@tijmenb looks good to me. I left 1 minor comment in there and the branch needs a rebase so the tests work (I fixed the config this morning, so it should work now).

carvil:
:shipit:

PR #381, opened 2016-09-06
Remove document type from specialist publisher metadata

steventux:
https://trello.com/c/Y0ncSoTO/258-clean-up-remove-document-type-from-pub-api-metadata-payload

Remove duplicate `document_type` property from the `specialist_document` schema metadata.
To do this we have to [be explicit in the enums for the `tribunal_decision_category` property](https://github.com/alphagov/govuk-content-schemas/commit/8212a1e3551017fd3c3dbb0bf52b399f38869f56#diff-1a705374a0f9ac97b29928d07d04b0bfR182) to satisfy [the `oneOf` clause for document specific metadata](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/specialist_document/publisher/details.json#L41).

Needs https://github.com/alphagov/specialist-publisher-rebuild/pull/891 to build correctly.
https://github.com/alphagov/specialist-publisher/pull/737 added for completeness.

### DIFF
--- a/dist/formats/specialist_document/frontend/schema.json
+++ b/dist/formats/specialist_document/frontend/schema.json
@@ -640,12 +640,6 @@
         "bulk_published": {
           "type": "boolean"
         },
-        "document_type": {
-          "type": "string",
-          "enum": [
-            "aaib_report"
-          ]
-        },
         "aircraft_category": {
           "type": "array",
           "items": {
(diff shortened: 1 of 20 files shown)


tuzz:
This is mutually dependent on https://github.com/alphagov/specialist-publisher-rebuild/pull/891. Going to merge this first as there are already things broken with Licensify so it'll be less red on Jenkins.

PR #382, opened 2016-09-08
Add notes_for_editors to taxon details hash

mobaig:
Also includes minor amendment to description field explanation - a taxon
internal name may not necessarily include the parent.


tijmenb:
Merging this, since the licence-finder tests fail unrelatedly.

PR #383, opened 2016-09-09
Make rendering_app optional for contacts

carlosmartinez:
It doesn't make sense for it to be required, and meanwhile it's causing validation errors.


kevindew:
Looks good 👍

PR #384, opened 2016-09-12
Service manual topics now link to email alert signups

36degrees:
Within the service manual we're making it possible for users to sign up for email alerts with particular service manual topics.

To do this we're creating an email_alert_signup for each topic, and associating it with the topic.


thehenster:
:+1:

PR #386, opened 2016-09-13
Remove "signup_link"

tijmenb:
`details.signup_link` is a piece of technical debt from the time we didn't have the links hash, or didn't use it properly. Policy Publisher and Finder Publisher shouldn't be sending this.


rubenarakelyan:
👍

PR #389, opened 2016-09-16
Don't allow additional properties for frontend links

tijmenb:
This changes the way the `links` are validated in frontend schemas.

Since #232 the frontend links have `additionalProperties` set to true. I think this is undesirable because it allows all data in the frontend links, defeating the purpose of schemas.

- Two detailed guide examples have the links incorrectly nested (`links -> parent -> parent` instead of `links -> parent -> links -> parent`). This wasn't picked up by the validation. This could have caused developers to build pages that work only with the incorrect example.
- The service manual example included a `expanded_links` example, which is not used anymore (and should never have been used in frontends).

### DIFF
--- a/formats/frontend_links_definition.json
+++ b/formats/frontend_links_definition.json
@@ -3,7 +3,7 @@
   "type": "array",
   "items": {
     "type": "object",
-    "additionalProperties": true,
+    "additionalProperties": false,
     "required": [
       "content_id",
       "title",
@@ -11,6 +11,19 @@
       "locale"
     ],
     "properties": {
+      "details": {
+        "type": "object",
+        "additionalProperties": true
+      },
+      "document_type": {
+        "type": "string"
+      },
+      "schema_name": {
+        "type": "string"
+      },
+      "public_updated_at": {
+        "type": "string"
+      },
       "content_id": {
         "$ref": "#/definitions/guid"
       },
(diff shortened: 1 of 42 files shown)


formats/frontend_links_definition.json
```diff
@@ -47,11 +60,7 @@
       "links": {
         "type": "object",
         "additionalProperties": true,
-        "properties": {
-          "parent": {
-            "$ref": "#/definitions/guid_list"
-          }
-        }
+        "description": "TODO: investigate if could this be validated recursively."
```
boffbowsh: I'm fairly sure the answer to this is no, definitely with current tooling. Probably within the scope of JSON schemas themselves :(

formats/frontend_links_definition.json
```diff
@@ -47,11 +60,7 @@
       "links": {
         "type": "object",
         "additionalProperties": true,
```
boffbowsh: We can use [`patternProperties`](https://spacetelescope.github.io/understanding-json-schema/reference/object.html#pattern-properties) here. Something like:

```json
       "links": {
          "type": "object",
          "patternProperties": {
            "^[a-z_]+$": {
              "$ref": "#/definitions/frontend_links"
            }
          }
        }
```

This allows us to validate that what is coming from content store has the basic required fields, but that object will also have `additionalProperties: true`

tijmenb:
It seems that the incorrect nesting example was used in government-frontend and caused a bug to be introduced. We'll fix that first.

boffbowsh:
Merging this as it isn't changing production code, whereas the govt-frontend one is.

PR #394, opened 2016-09-21
Generate notification schemas

tijmenb:
Currently this repo has two versions of a schema: a "frontend" version that describes the output of the content-store, and a "publisher" schema which describes the JSON clients are sending to the publishing-api. This PR adds a third "notification" variant, which describes the payload that's being sent over the message queue. Adding this was the original plan when the contract testing was first conceived, but it was somehow never implemented. 

In https://github.com/alphagov/govuk-content-schemas/pull/333, there was an attempt to add a generic schema that would describe all possible payloads. This was never used in the publishing-api and validation against it turned out to be quite complex because of the errors. The schemas have recently been simplified (https://github.com/alphagov/govuk-content-schemas/pull/374 https://github.com/alphagov/govuk-content-schemas/pull/378), which makes adding a schema per-format easier. 

The new schemas can be used to validate the message queue payload and to generate random data in queue-consuming applications. In particular, we will use it in the publishing-api to make sure there are no unexpected changes to the payload. 

The first commit is the most relevant to review. 

Part of: https://trello.com/c/fCvwqU5b

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,12 @@ processable formats.
 
 We use [JSON Schema](http://json-schema.org/) to define the formats.
 
-For each format there are two possible representations:
+For each format there are three possible representations:
 
 * the 'publisher' representation, which is used when a publishing application
   transmits data to the content store.
 * the 'frontend' representation, which is produced by the content store when a
   frontend application requests data
-
-In the future, there may be a third:
-
 * the 'notification' representation, which is used when broadcasting messages about content
   items on the message queue
 
--- a/dist/formats/case_study/notification/schema.json
+++ b/dist/formats/case_study/notification/schema.json
@@ -1,0 +1,613 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "content_id",
+    "details",
+    "document_type",
+    "expanded_links",
+    "govuk_request_id",
+    "links",
+    "locale",
+    "redirects",
+    "routes",
+    "schema_name",
+    "title",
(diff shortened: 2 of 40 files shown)


cbaines:
> In #333, there was an attempt to add a generic schema that would describe all possible payloads. This was never used in the publishing-api and validation against it turned out to be quite complex because of the errors. The schemas have recently been simplified (#374 #378), which makes adding a schema per-format easier.

What are the "errors"? I think having a schema for the data that you intend to validate is simpler than having several schemas, for which one, and only one should be valid (which is why I constructed the message queue schema in this way).

tijmenb:
The main problem with the validation is that the errors of a `oneOf` clause are not readable. Since the schema can't work out which of the `oneOf`'s _should_ be valid, it can only tell you that your payload isn't valid against one of 40 schemas. I've tried to make it work but wasn't able to.

danielroseman:
TAP error seems unrelated, but I'm not sure why it's happening.

36degrees:
For the email alert service it would make sense to have a 'generic' notification schema to test against, but I guess we could just pick a couple of relevant schemas to test with.

cbaines:
> The main problem with the validation is that the errors of a oneOf clause are not readable. Since the schema can't work out which of the oneOf's should be valid, it can only tell you that your payload isn't valid against one of 40 schemas. I've tried to make it work but wasn't able to.

I'd prefer to see this addressed in the schema validator, or by modifying the output it generates, rather than splitting up the schema, as I think this makes it harder to read and use.

I don't mind this change going ahead though, as I'm not aware that this is being used, or of a plan (in the product backlog of the publishing api) to use it.

PR #400, opened 2016-09-30
Allow redirects to gov.uk campaign URLs

robinjam:
This PR relates to:
* alphagov/short-url-manager#62
* alphagov/publishing-api#531

This PR needs merging first:
* alphagov/govuk_schemas_gem#16

The content designers need to be able to configure short URLs for gov.uk campaigns. For example, it needs to be possible to set up a redirect from `https://www.gov.uk/campaign-slug` to `https://my-campaign-title.campaign.gov.uk`.

This PR updates the `redirect` schema to allow gov.uk campaign URLs to be the destination of redirects. The `govuk_campaign_url` regex should match any HTTPS URL where the host is a subdomain of `campaign.gov.uk`.

## Regex explanation

Demo: http://regexr.com/3eblr

```
https://               # gov.uk campaigns must be served over https
(
  [a-zA-Z0-9]          # Each subdomain must start with an alphanumeric character...
  (
    [a-zA-Z0-9-]{0,61} # ...can be up to 63 characters long, and can contain dashes...
    [A-Za-z0-9]        # ...and must end with an alphanumeric character
  )?
  \.
)+                     # Match at least 1 subdomain
campaign\.gov\.uk
```

The remainder of the regex has the same structure as `absolute_fullpath`.

### DIFF
--- a/formats/redirect/publisher_v2/schema.json
+++ b/formats/redirect/publisher_v2/schema.json
@@ -92,6 +92,11 @@
       "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?(\\?([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?(#([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?$",
       "description": "A path with optional query string and/or fragment."
     },
+    "govuk_campaign_url": {
+      "type": "string",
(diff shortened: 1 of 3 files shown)


formats/redirect/publisher_v2/schema.json
```diff
@@ -118,7 +123,11 @@
           "enum": [ "prefix", "exact" ]
         },
         "destination": {
-          "$ref": "#/definitions/absolute_fullpath"
+          "type": "string",
+          "oneOf": [
```
tijmenb: @robinjam @pmanrubia could we change this to `anyOf`?

Using `oneOf` means that the string will be invalid if it matches more than exactly one of the clauses.

In this case it's probably impossible for a destination to be both a valid `absolute_fullpath` AND `govuk_campaign_url`, but you could easily see a next developer adding something here that will make that possible. See https://github.com/alphagov/govuk-content-schemas/pull/397 for the same around specialist publisher documents.

pmanrubia: Sure @tijmenb. It makes sense. We will update it as soon as possible.

PR #401, opened 2016-10-04
Add `ordered_related_items` link type

tijmenb:
This link type will be used for the sidebar on mainstream pages.

https://trello.com/c/mOeDK914

cc @MatMoore

### DIFF
--- a/formats/base_links.json
+++ b/formats/base_links.json
@@ -7,6 +7,10 @@
       "description": "Prototype-stage taxonomy label for this content item",
       "$ref": "#/definitions/guid_list"
     },
+    "ordered_related_items": {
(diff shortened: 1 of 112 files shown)


formats/base_links.json
```diff
@@ -7,6 +7,10 @@
       "description": "Prototype-stage taxonomy label for this content item",
       "$ref": "#/definitions/guid_list"
     },
+    "ordered_related_items": {
+      "description": "Related items. Mostly used for mainstream content to power the sidebar. Ordering of the links is determined by the editor in Panopticon.",
```
MatMoore: This should say what "related items" actually are. Can this be anything with a content id?

tijmenb: Updated. Is that enough?

PR #404, opened 2016-10-05
Add api_path to frontend_links

kevindew:
This allows us to send an api_path as part of content item link payload.

This will allow us to work towards deprecating the environment specific web_url
and api_url.


tijmenb:
Can you add a comment to `web_url` & `api_url` that they're deprecated?

PR #407, opened 2016-10-06
Deprecate api_url and web_url

kevindew:
Adds comments to the description of these fields to indicate that they are now deprecated.

As requested in: https://github.com/alphagov/govuk-content-schemas/pull/404


tijmenb:
❤️

PR #416, opened 2016-10-18
Remove "policies" link type

tijmenb:
This link type isn't used anywhere. It was probably introduced to capture a policy - document relationship, but we use `related_policies` for that now.

We've confirmed on staging that there are no links with this link type.

Thanks @boffbowsh for finding it.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -112,9 +112,6 @@
         "parent": {
           "$ref": "#/definitions/frontend_links"
         },
-        "policies": {
-          "$ref": "#/definitions/frontend_links"
-        },
         "policy_areas": {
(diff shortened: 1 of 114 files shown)


tijmenb:
Closing this for a new PR. The reason `policies` exist in examples and frontend schemas is that the [publishing-api uses it to put the reverse of working-groups links](https://github.com/alphagov/publishing-api/blob/0c23a5a2588c357f84602c38cfc9481ac9de9b25/app/queries/dependent_expansion_rules.rb#L55) there. 

Which means that if you tag a policy to a working group (like [this policy](https://www.gov.uk/government/policies/2012-olympic-and-paralympic-legacy) to [this working group](https://www.gov.uk/government/groups/olympic-and-paralympic-legacy-cabinet-committee)), the _working group_ will have a `policies` in the links.

PR #417, opened 2016-10-18
Don't allow `policies` to be set directly

tijmenb:
Working groups can have a `policies` link type, which is populated with the policies that have this working group in the `working_groups` links. This is done by the dependency resolution functionality [in the publishing-api](https://github.com/alphagov/publishing-api/blob/0c23a5a2588c357f84602c38cfc9481ac9de9b25/app/queries/dependent_expansion_rules.rb#L55).

This makes clear that publishing apps should not use the link type, but that it still can appear on the frontend.

Follow up to https://github.com/alphagov/govuk-content-schemas/pull/416.

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -112,9 +112,6 @@
         "parent": {
           "$ref": "#/definitions/frontend_links"
         },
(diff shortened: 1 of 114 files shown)


jackscotti:
giphy

PR #419, opened 2016-10-19
Remove Lloyds-related content schemas

tijmenb:
The share sale application has been decommissioned, so we don't need these schemas anymore.


carvil:
👍

PR #420, opened 2016-10-19
Add parent to service manual topic example

36degrees:
Service manual topics will usually have the service manual homepage as a parent, so include it in the example.

The exception to this is when the topic should not be included on the list of topics on the homepage. A current use case for this is the ‘communities’ topic, which is linked separately.

### DIFF
--- a/formats/service_manual_topic/frontend/examples/service_manual_topic.json
+++ b/formats/service_manual_topic/frontend/examples/service_manual_topic.json
@@ -10,6 +10,20 @@
   "phase": "alpha",
   "analytics_identifier": null,
   "links": {
+    "parent": [
+      {
+        "content_id": "053d245d-b1cd-4807-9dbc-ce78654121d1",
+        "title": "Service manual",
+        "base_path": "/service-manual",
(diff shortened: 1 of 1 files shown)


formats/service_manual_topic/frontend/examples/service_manual_topic.json
```diff
@@ -10,6 +10,20 @@
   "phase": "alpha",
   "analytics_identifier": null,
   "links": {
+    "parent": [
+      {
+        "content_id": "053d245d-b1cd-4807-9dbc-ce78654121d1",
+        "title": "Service manual",
+        "base_path": "/service-manual",
+        "description": "Helping government teams create and run great digital services that meet the Digital Service Standard.",
+        "api_url": "https://content-store.gov.uk/content/service-manual",
```
tijmenb: Should this be `https://www.gov.uk/api/content/service-manual`?

36degrees: Yes, good catch. It's wrong in a couple of other examples in this file… which I definitely didn't copy and paste… will fix.

PR #422, opened 2016-10-19
Document running dummy content store with bowler

36degrees:


cbaines:
Looks good to me, thanks for writing this :)

PR #424, opened 2016-10-21
Allow format filtering on finder publisher schemas

carvil:
We are moving the rendering app of the people page from whitehall to finder-frontend. In doing so, we are writing a JSON file to represent the new finder page. In that schema, we have to describe how are we going to filter the data for the finder page.

In rummager, we need to use the `format` so that the finder page gets populated with people (this is because the `document_type` of a person is an `edition` and the format is `person`).

Here is an example query: https://www.gov.uk/api/search.json?filter_format=person

This PR make is possible to filter data by format in finders.

Trello: https://trello.com/c/reCVyzpg/259-migrate-the-people-page-to-a-finder

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -272,6 +272,9 @@
               "items": {
                 "type": "string"
               }
+            },
+            "format": {
+              "type": "string"
             }
           }
         },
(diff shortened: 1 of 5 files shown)


carvil:
This PR is waiting on a PR in `manuals-publisher` so that the tests pass.

PR #426, opened 2016-10-24
Add top level change note

cbaines:
This is documented here in the Publishing API: https://github.com/alphagov/publishing-api/blob/f4c0c3d9044033ad9a680e644b4cee389c827521/doc/api.md#json-attributes

### DIFF
--- a/lib/govuk_content_schemas/frontend_schema_generator.rb
+++ b/lib/govuk_content_schemas/frontend_schema_generator.rb
@@ -12,6 +12,7 @@ class GovukContentSchemas::FrontendSchemaGenerator
     redirects
     routes
     update_type
+    change_note
   }.freeze
 
   OPTIONAL_PROPERTIES = %w{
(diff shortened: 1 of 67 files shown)


lib/govuk_content_schemas/frontend_schema_generator.rb
```diff
@@ -12,6 +12,7 @@ class GovukContentSchemas::FrontendSchemaGenerator
     redirects
     routes
     update_type
+    change_note
```
boffbowsh: Is this correct for frontend? Shouldn't change notes be sent as an array in change history or similar?

cbaines: > Is this correct for frontend? 

I believe this matches the behaviour of the publishing_api, the new top level `change_note` field is only sometimes used in the construction of the `change_history` member of the `details` object when presenting to the content store. These `INTERNAL_PROPERTIES` are members of the publisher schema to omit from the frontend schema.

> Shouldn't change notes be sent as an array in change history or similar?

Yes, as above, sometimes the Publishing API will construct the `change_history` from the change notes it has received.

boffbowsh: Sorry, totally inverted the logic of this in my head. This is a blacklist not a whitelist. 👍

PR #428, opened 2016-10-27
Deployment changes

cbaines:
govuk-content-schemas is now deployed as a standard app.


cbaines:
govuk-content-schemas was successfully deployed via "Deploy App" in Jenkins today.

PR #429, opened 2016-10-28
details[:change_history] is no longer required

carlosmartinez:
As part of the work around Publisher API managing change history, we no
longer mandate that apps send a change history in the details hash.


cbaines:
I can understand that this is part of the work, but I still don't understand why this change is being made?

dist/formats/specialist_document/frontend/schema.json
```diff
@@ -229,8 +229,7 @@
       "additionalProperties": false,
       "required": [
         "body",
-        "metadata",
-        "change_history"
```
cbaines: As said on slack, I'm unsure if this change to the frontend schema is intentional.

carlosmartinez: It is unintentional. Will add the relevant fix to GovukContentSchemas::FrontendSchemaGenerator

Gemfile.lock
```diff
@@ -40,5 +40,8 @@ DEPENDENCIES
   rake
   rspec
 
+RUBY VERSION
+   ruby 2.3.1p112
+
 BUNDLED WITH
-   1.10.6
```
cbaines: As far as I know, the changes to the `Gemfile.lock` should be removed, with the reasoning being something to do with the version of bundler specified from puppet.

PR #431, opened 2016-11-02
 Add new attributes for tribunal decisions

lgarvey:
The last PR was reverted because the changes were not backward
compatible causing the PR to fail CI tests.

This PR adds two extra columns - tribunal_decision_categories and
tribunal_decision_sub_categories and leaves the old non pluralised
fields intact.


lgarvey:
@tijmenb

tijmenb:
The test failures here are unrelated by the way, I'll fix those.

lgarvey:
Yes, I'm happy to make those changes. I'll get them done tomorrow.

PR #433, opened 2016-11-03
Asylum finder generated schema updates 

lgarvey:
See #431


tijmenb:
Makes https://github.com/alphagov/specialist-publisher/pull/927 pass.

PR #437, opened 2016-11-10
Add generic schema

rubenarakelyan:
This commit adds a “generic” schema which includes nothing in the `details` or `links` hashes beyond what is defined in the base JSON documents.

The reason is to provide a schema for content items that do not need any information in the content-store to be rendered. (Rather than a `placeholder`, which implies that it needs but doesn't have data _yet_).

Previously, the `special_route` schema has been used for these types of pages, for example the “Find your local council” page ([code](https://github.com/alphagov/frontend/blob/e460d40b0095d6c07570aa14b884fcbdde65b7c6/lib/tasks/publishing_api.rake)). This is wrong, because `special_route` is a special kind of schema that doesn’t have a frontend representation. It was intended for the use of routing, never to be served from the content-store (this was never implemented). 

We think `special_route` could be deprecated in favour of this schema.

Other names we considered:

- base
- basic
- page
- minimal
- minimum

The name of the schema is open for discussion.

### DIFF
--- a/dist/formats/generic/frontend/schema.json
+++ b/dist/formats/generic/frontend/schema.json
@@ -1,0 +1,539 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "links",
+    "title",
+    "details",
+    "locale",
+    "content_id",
+    "document_type",
+    "schema_name"
+  ],
+  "properties": {
+    "base_path": {
+      "$ref": "#/definitions/absolute_path"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ]
+    },
+    "public_updated_at": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "first_published_at": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "publishing_app": {
+      "$ref": "#/definitions/application_name"
+    },
+    "rendering_app": {
+      "$ref": "#/definitions/application_name"
+    },
+    "locale": {
+      "$ref": "#/definitions/locale"
+    },
+    "need_ids": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "analytics_identifier": {
+      "$ref": "#/definitions/analytics_identifier"
+    },
+    "phase": {
+      "description": "The service design phase of this content item - https://www.gov.uk/service-manual/phases",
+      "type": "string",
+      "enum": [
+        "alpha",
+        "beta",
+        "live"
+      ]
+    },
+    "details": {
+      "$ref": "#/definitions/details"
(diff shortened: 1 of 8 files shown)


tijmenb:
These are the current `special_route` documents:

```ruby
- "/"
- "/api"
- "/apple-touch-icon-114x114.png"
- "/apple-touch-icon-144x144.png"
- "/apple-touch-icon-57x57.png"
- "/apple-touch-icon-72x72.png"
- "/apple-touch-icon-precomposed.png"
- "/apple-touch-icon.png"
- "/contact"
- "/courts-tribunals"
- "/design-principles"
- "/favicon.ico"
- "/find-local-council"
- "/google6db9c061ce178960.html"
- "/google991dec8b62e37cfb.html"
- "/googlef35857dca8b812e7.html"
- "/government"
- "/help"
- "/homepage"
- "/humans.txt"
- "/info"
- "/lloyds-share-offer-further-details"
- "/lloydsfactsheet"
- "/performance"
- "/performance/big-screen"
- "/random"
- "/robots.txt"
- "/search.json"
- "/search"
- "/search/opensearch.xml"
- "/service-manual"
- "/sitemap.xml"
- "/sitemaps"
- "/tour"
- "/trade-tariff"
- "/transformation"
- "/ukwelcomes"
```

Other things that can use `generic` schema are the things that don't only use the content item for the breadcrumb and related links. These include https://www.gov.uk/bank-holidays and smart answers.

tijmenb:
@alphagov/team-gov-uk-core-formats-and-publishing @alphagov/team-gov-uk-custom-formats @alphagov/team-gov-uk-publishing-platform does this sound allright?

tijmenb:
Ooh, forgot to mention the immediate need: https://github.com/alphagov/whitehall/pull/2818.

We're migrating the [Services and Information pages](https://www.gov.uk/government/organisations/hm-revenue-customs/services-information) from whitehall to collections frontend. Those pages don't need any stuff in the details hash, because everything is retrieved from rummager. It seems silly to create a entire new schema. Since we now have `document_type`, we can specify the "type" in there but use a generic schema.

kevindew:
Makes sense to me. Big improvement on `special_route` 👍

binaryberry:
All sounds sensible to me - I find the names minimal or minimum to describe what it is better, which is a bare bones schema. "Generic" makes me think it would include the fields most commonly found on other schemas, such as body.

formats/generic/frontend/examples/generic_item.json
```diff
@@ -0,0 +1,15 @@
+{
+  "content_id": "170bff3e-444c-4571-b26f-259d2b8d4f27",
+  "base_path": "/any-thing",
```
rubenarakelyan: Yes, I have changed the example to an actual services and information content item.

tijmenb:
@binaryberry 

> I find the names minimal or minimum to describe what it is better, which is a bare bones schema

Would `bare_bones` describe it even better?

binaryberry:
@tijmenb yes, that would work I think!

tijmenb:
Merged after a chat with @jennyd, who's okay with this name.

PR #441, opened 2016-11-24
Add a new kind of facet for "topical" content

MatMoore:
Topical means anything that changes from one state to another on a predefined date. I'm using this for topical events, but it could also apply to consultations.

The goal is to support a special status field like in the following:
screen shot 2016-11-24 at 15 05 14

This will requires finder frontend to pass a special date filter to rummager, like `filter_end_date=to:2016-11-24`.

I'm not sure about the name "topical" for the type - suggestions welcome!

https://trello.com/c/oVfSAO9W/311-migrate-policy-areas-topical-events-index-to-a-finder

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -622,6 +622,25 @@
           "$ref": "#/definitions/nation_applicability"
         }
       }
+    },
+    "label_value_pair": {
+      "description": "One of many possible values a user can select from",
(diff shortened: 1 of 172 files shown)


cbaines:
@MatMoore There is a deploy this afternoon if you are happy for this to be merged and deployed?

MatMoore:
@cbaines That should be fine but before I merge I'm just going to change the name `option` to be more specific.

PR #446, opened 2016-11-29
Add pry-byebug gem

binaryberry:
For debugging purposes.


cbaines:
If you are going to say why the change is being made in the pull request, it would be good to have this information in the commit as well.

binaryberry:
Fair enough - amended now.

PR #444, opened 2016-11-29
[Don't merge] Remove `format` from frontend content items

tijmenb:
The `format` field is deprecated and is currently an alias for `schema_name`. This regularly confuses people because we use the word "format" elsewhere too.

Todo before merge:

- [x] Stop government-frontend using `format` (https://github.com/alphagov/government-frontend/pull/202)
- [x] Make sure tests pass for this PR
- [x] Make sure that tests for multipage-frontend are run against `deployed-to-production`
- [x] Make sure that multipage-frontend is not using `format` anymore
- [ ] Make content-store stop exposing `format` (https://github.com/alphagov/content-store/pull/250)

https://trello.com/c/C8A9bnuO/344-work-on-making-the-document-type-concept-clearer

PR #447, opened 2016-11-30
Use emphasised organisations definition in statistical data set

bevanloon:
When committing fc4e66a, I forgot to refactor Statistical Data Set's `emphasised_organisation` to use the newly created definition. This is somewhat amusing because writing the `details.json` for Statistical Data Set prompted the refactor of `emphasised_organisation`.

cc @fofr

PR #448, opened 2016-11-30
Reformat examples as part of the build

tijmenb:
This adds a step to the build process to reformat the example JSON files. The main reason is that this allows us to quickly edit all the examples. For example, #444 removes the `format` field simply by adding `data.delete("format")` to the rake task.

This will also prevent errors with duplicated keys in the JSON (converting it to a Ruby hash will solve them).

PR prompted by https://github.com/alphagov/govuk-content-schemas/pull/444#discussion_r90045687.

cc @andrewgarner


andrewgarner:
👍 

After my comment yesterday I was also thinking about whether we could incorporate something like https://github.com/PagerDuty/jsonlint

tijmenb:
With this I think we catch most errors. The examples are validated against the schemas, so any syntax errors are picked up already.

fofr:
:100: 👍

fofr:
Can something similar be applied to `details.json` and `links.json`?

A PR made it through today that has incorrect indentation which I think this would have prevented:
https://github.com/alphagov/govuk-content-schemas/pull/447#pullrequestreview-10723190

tijmenb:
Sure, I don't see any reason not to do that. Seems useful!

PR #450, opened 2016-12-02
Content schemas contains schemas

MatMoore:
It's really confusing when the README and how-to guides talk about "formats" instead of "schemas".

This is because `format` used to be a content store field and it was split into two things: `document_type` and `schema_name`.

The term "format" is often used more like `document_type`, which is different to the schemas. A schema may be used for multiple document types.

Not addressed in this PR:
- Rake tasks still use "format"
- Directory structure still uses "formats"


README.md
```diff
@@ -1,14 +1,14 @@
...
 The aim of it is to support 'contract testing' between the frontend and
 publisher apps by expressing the schema and examples in strict, machine
-processable formats.
+processable form.
 
-We use [JSON Schema](http://json-schema.org/) to define the formats.
+We use [JSON Schema](http://json-schema.org/) to define the schemas.
 
-For each format there are three possible representations:
+For each schemas there are three possible representations:
```
tijmenb: `schema`

MatMoore: doh

README.md
```diff
@@ -1,14 +1,14 @@
 # GOV.UK content schemas
 
-This repo contains schemas and examples of the content formats used on GOV.UK.
+This repo contains schemas and examples of the content schemas used on GOV.UK.
```
cbaines: I don't think this makes sense any more if you replace formats with schemas. The term document type is closer in usage than format.

cbaines: Maybe say "respective content" rather than "content schemas"?

andrewgarner:
🎉 🎉 🎉

fofr:
:100: :+1:

PR #454, opened 2016-12-07
Withdrawn

dougdroper:
Frontend will be able to filter out withdrawn links based off this
withdrawn flag.

https://trello.com/c/Tv6uNqYq/857-expose-withdrawn-state-for-links

### DIFF
--- a/formats/frontend_links_definition.json
+++ b/formats/frontend_links_definition.json
@@ -63,6 +63,7 @@
         "$ref": "#/definitions/analytics_identifier"
       },
       "withdrawn": {
+        "description": "Indicates whether the item has been unpublished as withdrawn, https://github.com/alphagov/publishing-api/blob/master/doc/model.md#workflow",
         "type": "boolean"
       },
       "links": {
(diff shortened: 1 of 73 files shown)


formats/frontend_links_definition.json
```diff
@@ -62,6 +62,9 @@
       "analytics_identifier": {
         "$ref": "#/definitions/analytics_identifier"
       },
+      "withdrawn": {
+        "type": "boolean"
```
tijmenb: Can you add a description to this? Perhaps with a link to the publishing-api docs about what it means?

tijmenb:
Currently the frontend links are a subset of the attributes in the frontend schema. Should we expose this element in the base content item too for consistency?

fofr:
What's the user need for including withdrawn links?

gpeng:
@fofr Publications that are `withdrawn` still require their attachments (which are also `withdrawn` in the Publishing API) to be rendered on the page otherwise they may no longer make sense to users as the 'actual' content is often in the attachment. Document Collections don't currently render links to `withdrawn` content so we need to be able to identify (and ignore) them for that format.

dougdroper:
Spoke to @tijmenb about this and decided against exposing this element in the base content item.

PR #460, opened 2016-12-13
Configure downstream builds in Jenkins 2

suzannehamilton:
This is a spike to work out how best to trigger builds of projects which depend on the schemas.

In this approach, we call the downstream builds directly from the govuk-content-schemas Jenkinsfile. We'll have to test whether this approach updates pull request checks correctly.

It's separate from PR #458, because we might be able to merge that one as soon as this project is being built by Jenkins 2.

https://trello.com/c/S4DZbsPa/285-jenkins-2-migration

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,12 +30,13 @@ node {
       sh('./jenkins.sh')
     }
 
-    stage("Push release tag") {
-      govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
+    if (env.BRANCH_NAME == 'master') {
+      stage("Push release tag") {
+        govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
+      }
+
+      govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, 'release', 'deploy')
     }
-
-    govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, 'release', 'deploy')
-
   } catch (e) {
     currentBuild.result = "FAILED"
     step([$class: 'Mailer',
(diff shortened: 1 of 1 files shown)


suzannehamilton:
Currently blocked. Once one of the downstream projects has been migrated to Jenkins 2 (and those changes are pushed to production), we can add that project to this Jenkinsfile to check that the build triggers work correctly.

suzannehamilton:
This is now unblocked and ready for review because collections-publisher can run schema tests on the deployed-to-production branch.

Jenkinsfile
```diff
@@ -45,3 +46,18 @@ node {
     throw e
   }
 }
+
+// Run schema tests outside of 'node' definition, so that they do not block the
+// original executor while the downstream tests are being run
+if (env.BRANCH_NAME != 'master') {
```
cbaines: If I remember correctly, the current CI setup has tests in apps are run regardless of the branch, however, I think the overall job status is only dependent on all of the other builds succeeding when building the master branch.

This looks like the app tests are only run when not building master, which I think is different from how the CI works currently. Is this an intentional change?

suzannehamilton: > This looks like the app tests are only run when not building master, which I think is different from how the CI works currently. Is this an intentional change?

Good spot! That's a mistake. I didn't realise that the app tests were triggered by the master build when I originally wrote that check last week - I should have spotted it when I updated it today. I've fixed it now.

> I think the overall job status is only dependent on all of the other builds succeeding when building the master branch.

This change is intentional: in the Jenkins 1 world, every app build reported its status to GitHub individually when triggered by a push to schemas. The Jenkins 2 pipeline builds aren't configured to individually update GitHub, so an app build failure needs to cause the parent schema build to fail, regardless of which branch is being built, so that the failure is visible on the PR.

cbaines: > > This looks like the app tests are only run when not building master, which I think is different from how the CI works currently. Is this an intentional change?
> 
> Good spot! That's a mistake. I didn't realise that the app tests were triggered by the master build when I originally wrote that check last week - I should have spotted it when I updated it today. I've fixed it now.

Good good :)

> > I think the overall job status is only dependent on all of the other builds succeeding when building the master branch.
> 
> This change is intentional: in the Jenkins 1 world, every app build reported its status to GitHub individually when triggered by a push to schemas. The Jenkins 2 pipeline builds aren't configured to individually update GitHub, so an app build failure needs to cause the parent schema build to fail, regardless of which branch is being built, so that the failure is visible on the PR.

Ok, every app will still report the status of the triggered builds for the commit referenced by `deployed-to-production` for the respective app, which is not the commit that we are intentionally being tested.

In this case it might not have much effect other than just setting a status for the commit, as the commit referenced by `deployed-to-production` shouldn't be part of a Pull Request, but I think it does replace any existing status information (possibly because the name/id of the information sent to GitHub is the same). From looking at a Pull Request, it appears that the status information is frozen at the point that it was merged, so at least that is not confusing.

I think this situation is ok, at least initially, but its something to be aware of.

PR #463, opened 2016-12-15
Add external related links to generic schema

brenetic:
We are in the process of stopping Publisher using Panopticon to register routes for it's documents. At the same time we are changing the schema that Publisher formats use to `generic`. External related links are currently used by all the formats.

### DIFF
--- a/dist/formats/generic/frontend/schema.json
+++ b/dist/formats/generic/frontend/schema.json
@@ -223,6 +223,24 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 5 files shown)


cbaines:
Given that this makes the generic schema not really generic (as its now got some specific things related to Mainstream Publisher), I'd prefer if the additionalProperties restriction was lifted, or just create another schema that looks like this (maybe "publisher-temporary")?

brenetic:
Closing this as we are going to use a new scheme

PR #464, opened 2016-12-16
Add temporary schema for mainstream formats

brenetic:
Created new `mainstream_temporary` schema for all the mainstream formats used by Publisher. This is only temporary for the purposes of removing the use of the Placeholder format. During migration we will be breaking each format out into their own schemas and then removing the `mainstream_temporary` schema.


tijmenb:
Smart answers will need to use this schema as well (it also uses `external_related_links`). Perhaps we should call this something non-temporary?

PR #465, opened 2016-12-16
Add generic with external related links schema

brenetic:
Created new `generic_with_external_related_links` schema for all the mainstream formats
used by Publisher, Smart Answers and Frontend. This is only temporary for the
purposes of removing the use of the Placeholder format. During migration we
will be breaking each format out into their own schemas and removing the
`generic_with_external_related_links` schema.


brenetic:
cc @tijmenb @cbaines

PR #468, opened 2016-12-21
Add speeches schema and examples

fofr:
* Add details and links schema for speeches
* Add examples for draft transcript, transcript, written to parliament and authored article types of speech
* Add example where speaker doesn't have a GOV.UK page (eg The Queen)

There are no known features missing from the schema.

## Speaker images

The images are pulled from the profile of the assigned speaker. For now the frontend expects this to come from the `image` property in the details. It may in future make more sense for it to come via an image property in the links hash of the speaker – as that's the dependency.

Story: https://trello.com/c/TWsLTRzc/547-10-speech-migration-mvp-content-schema-examples-and-front-end-work

### DIFF
--- a/dist/formats/speech/frontend/schema.json
+++ b/dist/formats/speech/frontend/schema.json
@@ -1,0 +1,654 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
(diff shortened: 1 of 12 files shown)


bevanloon:
LGTM 👍

PR #469, opened 2016-12-21
Remove v1 schemas (in a silly way)

tijmenb:
At the moment the schema generation code is quite complicated - the V2 schemas depend on the V1 schemas to be generated. This makes removing the code that generates the V1 schemas pretty hard.

I've been working on a rewrite of the schema generation code. To make that PR easier to review, this commit adds a rake task to delete all V1 schemas as a last step. 

I'm also curious of there's anything still using V1 schemas - it shouldn't because the V1 endpoint has been removed.


tijmenb:
😢 

Turns out specialist-publisher and travel-advice-publisher still use the old schemas to validate.

https://ci.dev.publishing.service.gov.uk/job/govuk_specialist-publisher_schema_tests/717
https://ci.dev.publishing.service.gov.uk/job/govuk_travel_advice_publisher_schema_tests/1315

That's pretty odd, considering they would be sending incorrect data to publishing-api and that would be rejected. @kevindew any idea why this wouldn't have come up?

tijmenb:
Closing until TAP & SP are fixed: https://trello.com/c/wfg9PR7v

kevindew:
@tijmenb it looks the specialist-document v1 schema is still compatible with the Publishing API https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/specialist_document/publisher/schema.json if they used some of the other attributes (eg links) they'd get an error but they're not.

PR #473, opened 2017-01-04
Add service toolkit format

mcmillan:
We're building a new page as part of the service manual called the "service toolkit", which is a list of links to government products that looks [like this](https://govuk-product-page-prototypes.herokuapp.com/govuk-service-manual/index).

This adds a schema called `service_manual_service_toolkit` which will support that page.

https://trello.com/c/GHY41UJi/546-create-service-toolkit-page


36degrees:
I'll squash this into two commits (a formats changes commit, and a regenerate schemas commit) and rebase once this has been reviewed.

36degrees:
Squashed into three commits to retain authorship.

PR #480, opened 2017-01-11
Remove Government requirement for Consultations and Speeches

andrewgarner:
The publishing applications are able to create draft documents for Consultations and Speeches that will happen in the future. We are not able to determine what the Government of the future will be so we can not require it be provided if we are to persist drafts.

This will be reviewed when the publishing applications are upgraded.

Some additional context can be found in the current publishing application Whitehall: https://github.com/alphagov/whitehall/commit/06a3a30fe7306825d84b01cce9bcd29fa556443e

### DIFF
--- a/dist/formats/consultation/frontend/schema.json
+++ b/dist/formats/consultation/frontend/schema.json
@@ -230,7 +230,6 @@
       "additionalProperties": false,
       "required": [
         "body",
-        "government",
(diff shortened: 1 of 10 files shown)


andrewgarner:
Closing in favour of https://github.com/alphagov/whitehall/pull/2959

PR #482, opened 2017-01-16
Tweak example text for the CMA finder content item

rubenarakelyan:
When a new email alert subscription is created, a short name is generated describing the subscription. If enough subscription items are chosen by the user, the short name becomes longer than 255 characters, which is the GovDelivery limit for this field.

This commit tweaks the text for the finder content item examples to reflect changes made to the CMA case finder to keep the short text under 255 characters.

Trello: https://trello.com/c/KTHRa4a0/422-fix-email-alert-api-errors-when-the-email-alert-short-name-is-longer-than-255-characters


rubenarakelyan:
For now, we'll just reduce the amount of text. There will also be a test in specialist-publisher to make sure that adding new types doesn't expand the text beyond 255 characters.

PR #490, opened 2017-01-18
Remove 'beta' property from topics

suzannehamilton:
Topics are no longer in beta and will be deprecated soon, so it's safe to remove the 'beta' property entirely.

https://trello.com/c/oW7gSMoB/328-remove-beta-tags-from-topics-in-collections

Marked as [DO NOT MERGE] - requires alphagov/collections-publisher#235 to be deployed to production.


tijmenb:
🤔 If collections-publisher depends on `beta`, the tests for this PR should fail:

https://ci.integration.publishing.service.gov.uk/job/collections-publisher/job/deployed-to-production/76/console

Can we check if collections-publisher is actually testing against the schemas?

suzannehamilton:
@tijmenb Yeah, that's strange. I'll take a look!

suzannehamilton:
@tijmenb I've tracked down the problem. The content schema tests were running against the `master` version of collections-publisher, _not_ the `deployed-to-production` version!

The culprit is this line in the Jenkins helper, which is called by the collections-publisher build (and all the other schema test builds):

```
sh('git merge --no-commit origin/master || git merge --abort')
```

Which fast-forwards to master despite the `--no-commit` flag. I've opened a PR for fixing it by adding a `--no-ff` flag: alphagov/govuk-puppet#5366

PR #497, opened 2017-01-25
Add author to need format

binaryberry:
This will allow to provide the author of each revision in Maslow.

https://trello.com/c/8YShGG7D/24-get-maslow-to-render-content-from-the-publishing-api


fofr:
Will content items containing this be publicly accessible?

PR #501, opened 2017-01-30
Add corporate information page format

fofr:
Goes with https://github.com/alphagov/government-frontend/pull/241

* Add schema and examples for corporate information pages
* Use "about page" as the canonical format document_type
* Include examples for media and complaints document types, but not all types as further examples aren't necessary – apart from document_type (which isn't used during rendering), the examples would be identical.
* Include a translated example for an organisation with a custom logo (Land Registry)

## Notable features
* Corporate information pages are always associated with a single organisation, the `guid` for the org is a required field in the details hash
* Document types that are not "About" pages are children of about pages (they live at the URL `/organisation-name/about/XXX`

## Notable omissions
* Custom logos (with images) are not handled in this PR, that's being tracked separately: https://trello.com/c/HtSGDoRb/835-add-organsation-logos-that-are-images-to-content-item
* About pages contain a block of dynamic text at the bottom. This has been omitted for now and is being tracked here: https://trello.com/c/8syU3iNh/836-add-corporate-information-block-to-about-pages-corporate-information-pages

Part of:
https://trello.com/c/8ySlDVSs/578-corp-info-migration-1-mvp-content-schema-examples-and-front-end-work

(At time of raising PR master was failing)


Rosa-Fox:
Would be good to have another pair of eyes to check over, but I ran all of these examples locally on `government-frontend` comparing with the data in live and they looked fine to me.

formats/corporate_information_page/publisher/details.json
```diff
@@ -0,0 +1,42 @@
...
+    "body",
+    "organisation"
+  ],
+  "properties": {
+    "body": {
+      "type": "string"
+    },
+    "organisation": {
+      "description": "A single organisation that is the subject of this corporate information page",
+      "$ref": "#/definitions/guid"
```
gpeng: Should we be strict with this and add `"maxItems": 1`?

fofr: It's not an array. It's a string that must be a GUID.

fofr:
I've rebased against master and switched to using new definitions where required.

formats/corporate_information_page/publisher/details.json
gpeng: Oh yes... So it is.

PR #504, opened 2017-01-30
Remove reason for change from change history definition

mcmillan:
The `reason_for_change` field inside a change history definition was
added in #280 to support `service-manual-publisher`, which wanted both a
reason for change and a change note.

We're now removing the reason for change field as it's no longer
considered useful. This removes it from the shared definition.

In the 10 months that have elapsed since it was added, it doesn't appear
to be used by any other applications. Searching GitHub for non-JSON
references to "reason_for_change" outside of the
service-manual-publisher repo [yields nothing](https://github.com/search?utf8=%E2%9C%93&q=org%3Aalphagov+reason_for_change+-repo%3Aalphagov%2Fservice-manual-publisher+-language%3Ajson&type=Code&ref=searchresults).

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -578,10 +578,6 @@
           "note": {
             "type": "string",
             "description": "A summary of the change"
-          },
-          "reason_for_change": {
-            "type": "string",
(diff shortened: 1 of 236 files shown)


danielroseman:
Seems sensible, but you'll need to re-run the build task and add the generated files in a separate commit.

mcmillan:
@danielroseman D'oh – sorry, that's done. There's a lot of them!

PR #505, opened 2017-01-30
[Discuss] Force definitions for shared features

fofr:
Features that share a name should share a definition. This:

* avoids the risk that a format implements a feature that clashes with
the name of another feature
* avoids the risk of implementing the same feature in a slightly
different way in multiple places
* makes it harder to write `details.json` poorly

I've raised this PR to have a discussion before devoting more time to it. As part of template consolidation we are [documenting the inconsistencies between content schemas](https://trello.com/c/MKJySNJL/100-document-the-content-schemas-inconsistencies) and looking at ways to make it difficult to introduce further inconsistencies without impeding development.

Here is the output of the rake task on master which shows which duplicate features aren't using defintions:

```
The following features are used in multiple schemas without definitions:

'body'
  - formats/case_study/publisher/details.json
  - formats/consultation/publisher/details.json
  - formats/detailed_guide/publisher/details.json
  - formats/document_collection/publisher/details.json
  - formats/fatality_notice/publisher/details.json
  - formats/hmrc_manual/publisher/details.json
  - formats/hmrc_manual_section/publisher/details.json
  - formats/html_publication/publisher/details.json
  - formats/manual/publisher/details.json
  - formats/manual_section/publisher/details.json
  - formats/news_article/publisher/details.json
  - formats/publication/publisher/details.json
  - formats/service_manual_guide/publisher/details.json
  - formats/service_manual_service_standard/publisher/details.json
  - formats/specialist_document/publisher/details.json
  - formats/speech/publisher/details.json
  - formats/statistical_data_set/publisher/details.json
  - formats/take_part/publisher/details.json
  - formats/topical_event_about_page/publisher/details.json
  - formats/working_group/publisher/details.json
  - formats/world_location_news_article/publisher/details.json

'first_public_at'
  - formats/case_study/publisher/details.json
  - formats/consultation/publisher/details.json
  - formats/detailed_guide/publisher/details.json
  - formats/document_collection/publisher/details.json
  - formats/fatality_notice/publisher/details.json
  - formats/news_article/publisher/details.json
  - formats/publication/publisher/details.json
  - formats/speech/publisher/details.json
  - formats/statistical_data_set/publisher/details.json
  - formats/world_location_news_article/publisher/details.json

'change_note'
  - formats/case_study/publisher/details.json
  - formats/placeholder/publisher/details.json
  - formats/service_manual_guide/publisher/details.json

'tags'
  - formats/case_study/publisher/details.json
  - formats/consultation/publisher/details.json
[...]

PR #509, opened 2017-02-02
Use definitions for shared properties in details

fofr:
Follow up to https://github.com/alphagov/govuk-content-schemas/pull/505
Please review commit by commit.

Features that share a name should share a definition. This:

* avoids the risk that a format implements a feature that clashes with
the name of another feature
* avoids the risk of implementing the same feature in a slightly
different way in multiple places
* makes it harder to write `details.json` poorly

## Add definitions for shared features

* Try to add useful descriptions for each definition
* All schema updates are backwards compatible
* Many definitions match original property name, but others use a new name to represent a definition for a more specific concept (eg documents property uses "attachments_with_thumbnails" definition)
* Add body definition that let's body be a string or an array containing two strings – govspeak and HTML
* Add definition for confusing tags property and document usage
* Policies and finders share many properties, these now use definitions
  * Policies had `additionalProperties: true` set on facets and filters, but on inspection of content the structure of the data was identical to finders. They've been switched to use the same definitions, removing any differences and reducing risk of drift

## Content schema technical debt found

* first_public_at needs to be replaced with first_published_at, documented in new definitions
* `public_updated_at` is provided at the top level but coming_soon and unpublishing formats still use their version in the details hash 4ae1373
* Travel advice has an ordered list of countries in the details, this could be an ordered list of links now links can be ordered
  * The date-time that renders for these countries is in an inconsistent format: https://www.gov.uk/api/content/foreign-travel-advice `"public_updated_at": "2016-11-09T14:28:01Z",` rather than `"public_updated_at": "2017-02-02T17:33:02.000+00:00",`
* Formats still have withdraw_notices in their details hash, despite being provided at the top level (eg c8b8ee6)
* The `groups` feature on service manual topics, topics and mainstream browse was slightly different
  * Topics and mainstream have been made consistent. The schema contains some leftover changes from https://github.com/alphagov/govuk-content-schemas/pull/203 - Properties marked as deprecated for future removal
  * Created a separate definition for groups that use content_ids rather than paths
  * `groups` for topics, mainstream browse and service manual could all use content_ids and links hash, similar to the behaviour of groups in document collections
* Subtle differences between `change_notes` and `child_section_groups` (0fddc8b) on manuals and HMRC manuals
  * HMRC manuals are very similar, except they use `section_ids` for many things – in this case change notes are associated by section_id rather than a section content item or path
* The email_alert_signup format uses `summary` for content, it should use `description`
* `email_signup_link` should be using dependency resolution and links, rather than path and URI a2cec84
* Travel advice contains `publishing_request_id` but it's not clear that it works, or is being used. According to @gpeng the team wrapped up before completion of this feature. 8f0673c2610eeaf84b5a107acb3ec1df47615352
* Manual sections (and HMRC manual sections) define their parent using "manual" in the details, this could be "parent" in the links hash
* Manuals define their organisations in the details, they should use links
* Finders have a `beta` flag, it can be set by the publisher and rendered by frontend, no content is currently using and the feature if it remains should switch to using the `phase` property at the top level
* HMRC manuals and email alert signups use custom breadcrumb properties
  * No breadcrumbs display on email alert signups, this needs to be fixed, and should use the common patter – basing breadcrumbs on the parent in the links hash
  * HMRC manuals is harder to update as it's based on section_ids and we don't control the publisher
* Policies uses a custom `nation_applicability` property but should use the more widely used `national_applicability` definition

`+109,999 −5,483` 🤔


formats/hmrc_manual/publisher/details.json
```diff
@@ -7,7 +7,7 @@
   ],
   "properties": {
     "body": {
-      "type": "string"
+      "$ref": "#/definitions/body"
```
h-lame: This feels mildly dangerous.  The frontend for rendering hmrc manuals doesn't understand that a body can be anything but a string.  If someone started submitting manuals with a multi-part govspeak+html body the schema would be accepted, but the frontend would break.  I don't know if other frontend apps would behave the same for bodies that used to be strings but now might be multi-part.

formats/email_alert_signup/publisher/details.json
```diff
@@ -13,17 +13,7 @@
       "minProperties": 1,
       "properties": {
         "tags": {
-          "type": "object",
-          "description": "DEPRECATED: The tags used to match subscribers lists",
```
h-lame: We're losing these custom descriptions.  Is it possible to keep them?

36degrees:
Looks great.

Publisher is now being tested as part of `continuous-integration/jenkins/branch` on new CI – you can safely ignore the "Verify publisher against content schemas" failure which is coming from old CI.

It'd be good to get a better understanding of the `groups` technical debt which affects the service manual, so I can get a card in our backlog to resolve it 👍

formats/definitions.json
```diff
@@ -526,6 +526,30 @@
           "$ref": "#/definitions/absolute_path"
         }
       }
+    },
+    "manual_organisations": {
+      "description": "A manual’s organisations. TODO: Switch to use organisations in links",
```
h-lame: I think we might be able to drop these already.  hmrc manuals don't look like they have the organisation in the details hash at all (but do have it in the links) and govuk manuals appear to have both.  We control govuk manuals so we could drop it and republish things relatively easily.  The frontend has already been updated to prefer the links.organisation over the details.organisation (see: https://github.com/alphagov/manuals-frontend/pull/169).

formats/definitions.json
```diff
@@ -550,6 +550,67 @@
...
       }
+    },
+    "hmrc_manual_change_notes": {
+      "description": "A history of changes to HMRC manuals and the associated section. section_id is included and required making it different to manual_change_notes",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "section_id",
```
h-lame: We could merge these definitions as although section_id is "required" for hmrc manuals change notes - it's not actually used by manuals-frontend when rendering, so a single definition with optional section_id would work.

formats/definitions.json
```diff
@@ -611,6 +611,89 @@
...
+                  "type": "string"
+                },
+                "base_path": {
+                  "$ref": "#/definitions/absolute_path"
+                }
+              }
+            }
+          }
+        }
+      }
```
h-lame: Like the previous definition for change notes I think we could merge these; the rendering app handles missing titles for groups and section_ids for sections so we can make both optional.  One fun quirk is that manuals-frontend has different rendering paths for hmrc manuals vs govuk manuals and doesn't render group titles at all for govuk even though they are required by the schema (I think because all govuk manuals have one group and the title is always "Contents" - the publisher doesn't let you create groups or give them titles).

formats/finder_email_signup/publisher/details.json
```diff
@@ -9,7 +9,7 @@
   ],
   "properties": {
     "beta": {
-      "type": "boolean"
+      "$ref": "#/definitions/finder_beta"
```
h-lame: Is it going to be a problem that this now allows null, when before it had to be true or false?  Is it worth checking the rendering app for this - if it uses the same finder frontend it's probably fine.

formats/definitions.json
fofr: Yay 👍 , I will add as a further commit.

formats/finder_email_signup/publisher/details.json
fofr: Here's the presenter, which I think will be fine:
https://github.com/alphagov/finder-frontend/blob/master/app/presenters/signup_presenter.rb#L14-L16

Interestingly finders themselves have already stopped using this:
https://github.com/alphagov/finder-frontend/blob/master/app/presenters/finder_presenter.rb#L31-L33

formats/email_alert_signup/publisher/details.json
fofr: I left these out as I wasn't certain they applied to all uses of tags. Happy to put them back in.

formats/hmrc_manual/publisher/details.json
fofr: Good point. I will split this into two definitions, one for each.

formats/finder_email_signup/publisher/details.json
h-lame: :+1:

formats/email_alert_signup/publisher/details.json
fofr: On closer inspection, this description was replaced with the following, which I think is clearer:
> Field used by email-alert-api to trigger email alerts for subscriptions to topics (gov.uk/topic) and policies (gov.uk/policies).

formats/definitions.json
fofr: I'm going to approach this in a separate PR, I'd prefer to minimise the risk with this larger set of changes.

h-lame: Fair enough.

formats/email_alert_signup/publisher/details.json
h-lame: I think that description is good, I guess I was concerned by the "DEPRECATED" bit from here which we've lost (I assume this has some semantic meaning about these formats where publisher will stop sending tags soon?) and the description of tags on hmrc_manuals which doesn't indicate its got anything to do with email-alert-api (although it may do - I'm just not sure).

h-lame: Oh.  Seems that "tags" were removed from the hmrc-manuals-api that translates from the hmrc publisher and publishing-api so maybe we can drop tags from that schema entirely.  I think the "DEPRECATED" tag might still be useful, although it's more about the `details/tags` key than the definition per se.

formats/hmrc_manual/publisher/details.json
fofr: Updated in 1d1176d2542f29c044afa084a72ecfbeeb227da2

formats/email_alert_signup/publisher/details.json
fofr: Should it be removed from the HMRC manuals schema altogether?

fofr:
It's worth noting that this PR hasn't looked for cases of the same feature being used across formats but with _different_ property names.

formats/email_alert_signup/publisher/details.json
h-lame: I think removed from hmrc manuals in a separate PR makes a lot of sense, yes.  Will give us a chance to make sure there aren't any copies lurking in the content-store that still have it, which we can republish if needs be.

formats/hmrc_manual/publisher/details.json
h-lame: :+1:

fofr:
Content schema tech debt captured in a Trello board here:
https://trello.com/b/BOULnRrS/content-schema-tech-debt

PR #510, opened 2017-02-03
Add links to v2 metadata

thomasleese:
We are [adding edition-level links][1] to the `publishing-api` to support the [the 'draft link' workflow][0]. The schema needs to be updated to allow publishing apps to send their links to the `publishing-api` using the [put content][2] endpoint.

[0]: https://trello.com/c/Yd7RZaaT/870-edition-link-workflow-3
[1]: https://github.com/alphagov/publishing-api/pull/749
[2]: https://github.com/alphagov/publishing-api/blob/master/doc/api.md#put-v2contentcontent_id

### DIFF
--- a/dist/formats/case_study/publisher_v2/schema.json
+++ b/dist/formats/case_study/publisher_v2/schema.json
@@ -119,6 +119,44 @@
         "republish"
       ]
     },
+    "links": {
+      "type": "object",
+      "patternProperties": {
+        "^[a-z_]+$": {
+          "$ref": "#/definitions/guid_list"
+        }
+      },
+      "description": "Links associated with a particular edition of a document.",
+      "not": {
+        "anyOf": [
+          {
+            "required": [
+              "available_translations"
+            ]
+          },
+          {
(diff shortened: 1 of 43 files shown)


fofr:
Can you include a PR description? (I came to the PR to find out why links needed to be added to v2 metadata)

kevindew:
Would you be able to separate this out into a commit for the code changes and a commit for regenerating the links? 

I notice we allow just any type of link - but I don't think this is something we allow in other places with links (correct me if I'm wrong) so it might be prudent that in the absence of known links to [block ones created as reverse links to the Publishing API](https://github.com/alphagov/govuk-content-schemas/blob/master/lib/govuk_content_schemas/frontend_schema_generator.rb#L23-L43) - I'm not sure though if that is easy though - @tijmenb might have some thoughts on this.

Also be good to have a description for the links in Publisher V2 to explain their purpose, this could also be used to explain this is an experimental feature or similar to describe where we're at with this links workflow

PR #489, opened 2017-01-18
Update specialist frontend schemas to what is on live

mcgoooo:
this updates the content schemas to what is on live

this is needed before we start the template consolidation work

PR #513, opened 2017-02-08
Update specialist frontend examples

mcgoooo:
this updates the content schemas to what is on live

this is needed before we start the template consolidation work

please refer to #489 for previous discussion, git went a bit funny


Gemfile.lock
```diff
@@ -52,5 +52,8 @@ DEPENDENCIES
   rake
   rspec
 
+RUBY VERSION
```
fofr: Can you remove the Gemfile.lock changes from this PR?

fofr:
One minor thing otherwise 👍

Gemfile.lock
mcgoooo: just doing that now

PR #517, opened 2017-02-14
Remove required body field

brenetic:
Currently Publisher does not validate the presence of content within in the body field for the Help Page and Answer formats. This means that we get an error for nil fields.

This PR now allows nil values to be sent in the body field for both the above formats.

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -220,9 +220,6 @@
     "details": {
       "type": "object",
       "additionalProperties": false,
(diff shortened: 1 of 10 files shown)


davidbasalla:
This feels like the right thing to do for the moment. We don't validate presence of `body` in Publisher, so making them 'required' here causes failures that are more difficult to debug and might get overlooked since Publisher would not show any indication of failure to send to the publishing api. Whether we should be having validations in Publisher for this field is probably another question... so 👍 from me (once green obvs)

PR #518, opened 2017-02-14
Add schema for business support schemes

rubenarakelyan:
This commit adds schemas and examples for the migrated business finance support schemes.

Trello: https://trello.com/c/UHsavPI5/485-create-a-new-specialist-document-type-for-business-support-scheme


rubenarakelyan:
Yup, checked with Jane and Constance and they're happy with the naming.

PR #521, opened 2017-02-15
Add/amend fields on business finance support scheme schema

rubenarakelyan:
This commit adds/amends metadata fields on the business finance support scheme schema.

Since no content items are using this schema yet, the fields can be changed without backwards-compatibility concerns.

Trello: https://trello.com/c/UHsavPI5/485-create-a-new-specialist-document-type-for-business-support-scheme

### DIFF
--- a/formats/specialist_document/publisher/details.json
+++ b/formats/specialist_document/publisher/details.json
@@ -232,21 +232,34 @@
         "bulk_published": {
           "type": "boolean"
         },
-        "additional_information": {
-          "anyOf": [
-            {
-              "type": "string"
-            },
-            {
-              "type": "null"
-            }
-          ]
+        "business_sizes": {
+          "type": "array",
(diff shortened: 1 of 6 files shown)


formats/specialist_document/publisher/details.json
```diff
@@ -265,6 +283,27 @@
             }
           ]
         },
+        "locations": {
```
jennyd: Do locations still need to be there? They were the precursor to postcode search (which uses GSS codes), and I thought they weren't used any more (and hoped we'd got rid of them a while back).

rubenarakelyan: After consultation we've removed locations altogether.

rubenarakelyan:
GSS codes have been removed since they were useful when we had postcode search (which we won't have with the new finder).

PR #524, opened 2017-02-17
Add `poster_url` to service standard format

mcmillan:
We're adding a link to a poster of the service standard to the service
standard page in the service manual. This adds a `poster_url` field to
the `service_manual_service_standard` format to allow us to manage this
via service-manual-publisher.

### DIFF
--- a/formats/service_manual_service_standard/publisher/details.json
+++ b/formats/service_manual_service_standard/publisher/details.json
@@ -5,6 +5,11 @@
   "properties": {
     "body": {
       "$ref": "#/definitions/body"
+    },
+    "poster_url": {
+      "type": "string",
+      "format": "uri",
+      "description": "URL to the service standard poster (absolute)"
     }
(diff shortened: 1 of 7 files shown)


formats/service_manual_service_standard/publisher/details.json
```diff
@@ -5,6 +5,11 @@
   "properties": {
     "body": {
       "$ref": "#/definitions/body"
+    },
+    "poster_url": {
```
mcmillan: I'm really not sure that this should be a string – it should probably be an `asset_link` definition, but we'd need to make a bunch of updates in service-manual-publisher for it to work properly. @36degrees – what do you reckon?

36degrees: I think a string is fine for now. We can revisit when updating the standard actually moves into publisher proper, rather than a rake task.

PR #525, opened 2017-02-17
Update travel advice examples based on live content

fofr:
Ensure that examples are accurate before using them for tests in government-frontend.

I've also confirmed that a `no-parts` travel advice could feasibly published, although all countries currently have at least 5 parts.

Part of https://trello.com/c/L5Nd0AdH/137-3-render-travel-advice-in-government-frontend-l


nickcolley:
I've reviewed this by running the [dummy content store](https://github.com/alphagov/govuk-dummy_content_store) and checking these examples render correctly.

👍

PR #529, opened 2017-02-20
Remove format fields

tijmenb:
`format` has been deprecated and is replaced by `document_type` and `schema_name`.


fofr:
> This field will be removed by Tijmen.
True to his word.

PR #530, opened 2017-02-20
Improved example Guide documents

whoojemaflip:
The previous guide frontend example was missing some tagged fields.

For testing purposes, we've added a few more examples with varying numbers of parts.


davidbasalla:
Looks good! 👍

PR #531, opened 2017-02-21
Add Simple smart answer schema

davidbasalla:
Also added definitions for nodes and node options.

Generate simple smart answer schema.

Also regenerate all other schemas as we added definitions.

https://trello.com/c/2ucbq949/643-migrate-simple-smart-answer-format-to-rendered

### DIFF
--- a/formats/simple_smart_answer/frontend/examples/simple-smart-answer.json
+++ b/formats/simple_smart_answer/frontend/examples/simple-smart-answer.json
@@ -1,0 +1,443 @@
+{
+  "analytics_identifier": null,
+  "base_path": "/become-a-driving-instructor",
+  "content_id": "bc1c2f06-8601-452b-8f18-37520ba7d1f8",
+  "document_type": "simple_smart_answer",
+  "first_published_at": "2016-02-29T09:24:10.000+00:00",
+  "format": "simple_smart_answer",
+  "locale": "en",
+  "need_ids": [
+
+  ],
+  "phase": "live",
+  "public_updated_at": "2017-02-22T13:50:49.000+00:00",
+  "publishing_app": "publisher",
+  "rendering_app": "frontend",
+  "schema_name": "simple_smart_answer",
+  "title": "Become a driving instructor",
+  "updated_at": "2017-02-23T10:35:08.669Z",
+  "withdrawn_notice": {
+  },
+  "links": {
+    "mainstream_browse_pages": [
+      {
+        "analytics_identifier": null,
+        "api_path": "/api/content/browse/driving/teaching-people-to-drive",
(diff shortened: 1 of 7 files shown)


formats/definitions.json
```diff
@@ -142,6 +142,51 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "title"
+        ],
+        "properties": {
+          "kind": {
+            "type": "string"
```
cbaines: If kind has a few valid values, an enum [1] might be better.

1: http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.20

davidbasalla:
thanks @cbaines, have updated the definition!

formats/definitions.json
```diff
@@ -142,6 +142,54 @@
         }
       }
     },
+    "nodes": {
```
fofr: Does this need to be a definition if it's only used in simple smart answers?

This data structure is complex and needs some description fields to explain it.

davidbasalla:
Thanks for the review @fofr, have removed the definition as it would only be used by simple smart answers and have added descriptions

fofr:
Could you include some real world examples?

davidbasalla:
@fofr have replaced the example with a real world example! Is the general idea to always add a real example - rather than a toy one?)

fofr:
@davidbasalla Yeah, it's best to have examples that cover the range of real world uses.

formats/simple_smart_answer/frontend/examples/simple-smart-answer.json
```diff
@@ -0,0 +1,447 @@
...
+  "description": "Find out if you can become an approved driving instructor (ADI) and what you need to do to qualify",
+  "details": {
+    "start_button_text": "Start now",
+    "body": "<p>Find out if you can become an approved driving instructor (<abbr title=\"approved driving instructor\">ADI</abbr>) in Great Britain and what you need to do to qualify.</p>\n\n<div role=\"note\" aria-label=\"Information\" class=\"application-notice info-notice\">\n<p>There’s a different process in <a rel=\"external\" href=\"http://www.nidirect.gov.uk/index/information-and-services/motoring/driving-for-a-living/driving-instructors/how-to-become-an-approved-driving-instructor-adi.htm\">Northern Ireland.</a></p>\n</div>\n\n",
+    "nodes": [
+      {
+        "kind": "question",
+        "slug": "question-1",
+        "title": "Are you already registered as a driving instructor in another EU country?",
+        "body": "\n",
```
fofr: This feels wrong to me

davidbasalla:
Good catch @fofr, have updated the presenter to send the `body` field for nodes only when present.

PR #539, opened 2017-02-27
Add finder links to specialist documents

thomasleese:
We're going to be adding links between specialist documents and finders to help template consolidation. The name `finder` is not set in stone and is up for discussion.

See [our Trello card](https://trello.com/c/2wzkWWAR/439-add-links-between-specialist-documents-and-the-finder-blocking-template-consolidation-from-07-03) for background.

This branch is based on changes I've made in the #543 PR, which should get reviewed and merged first.

/cc @tijmenb

### DIFF
--- a/dist/formats/answer/publisher_v2/schema.json
+++ b/dist/formats/answer/publisher_v2/schema.json
@@ -120,52 +120,7 @@
       ]
     },
     "links": {
-      "type": "object",
-      "patternProperties": {
-        "^[a-z_]+$": {
-          "$ref": "#/definitions/guid_list"
-        }
-      },
-      "description": "Links associated with a particular edition of a document.",
-      "not": {
-        "anyOf": [
-          {
-            "required": [
-              "available_translations"
-            ]
-          },
-          {
-            "required": [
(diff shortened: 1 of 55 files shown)


kevindew:
Don't we actually want to make this an edition link? so it'd appear in https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/specialist_document/publisher_v2/schema.json#L122 and I presume the frontend schema, but not be in links one? - not sure if that's easy as I assume schemas aren't configured for that scenario?

tijmenb:
Do you know all the things a specialist document page needs from the `finder`? I was wondering if we could use the `parent` link type for this, so that we decouple the specialist documents from the concept of a finder.

thomasleese:
@tijmenb It uses the `facets` of the finder. We're not opposed to using `parent` as the `link_type`, but we can't decouple the documents on the frontend side -- the parents would still always need to be finders.

tijmenb:
Make sense! I agree with Kevin to make it an edition-level link. `finder` is a fine name for it.

PR #541, opened 2017-02-28
Add auth_bypass_id attribute to metadata

davidbasalla:
We want to be able to give people access to the draft stack for other states than just fact check (i.e. draft or ready) so we rename the field to be less specific. The [docs of authenticating proxy](https://github.com/alphagov/authenticating-proxy#technical-documentation) refer to this feature as "The application also supports bypassing authentication via a valid JWT token" so the new name `auth_bypass_id` feels more appropriate here...

We want to be able to maintain the old fact_check_id working for a short time in conjunction with the new auth_bypass_id attribute.

https://trello.com/c/7oRBYwnn/690-rename-fact-check-id-to-auth-bypass-id

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -458,9 +458,13 @@
             "type": "string"
           }
         },
+        "auth_bypass_ids": {
+          "$ref": "#/definitions/guid_list",
+          "description": "A list of ids that will allow access to this item for non-authenticated users"
+        },
         "fact_check_ids": {
           "$ref": "#/definitions/guid_list",
-          "description": "A list of ids that will allow access to this item for fact checking."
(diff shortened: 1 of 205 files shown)


cbaines:
I think the way the schema tests are currently set up enforces adding backwards compatibility in some way.

As publisher needs changing to generate the new data, I don't think you can make that change in a backwards compatible way. I think the remaining option is to make this change in a backwards compatible way, so add auth_bypass_ids and deprecated, but do not remove fact_check_ids. govuk-content-schemas can then be deployed, and afterwards publisher can be changed to use auth_bypass_ids, and then after that, fact_check_ids can be removed. Does this make sense?

binaryberry:
@cbaines it does, thanks! I shall amend this PR to reflect this.

PR #543, opened 2017-02-28
Add support for stricter edition links

thomasleese:
This way we can use a white list rather than a black list.

### DIFF
--- a/lib/govuk_content_schemas/schema_combiner.rb
+++ b/lib/govuk_content_schemas/schema_combiner.rb
@@ -53,14 +53,15 @@ private
   end
 
   def add_links(schema, target)
-    return unless schemas.key?(:base_links)
+    base_links = schemas[:base_links]
+    edition_links = schemas[:edition_links]
+    linkset_links = schemas[:links]
+
+    order = [edition_links, linkset_links, base_links].compact
+    return if order.empty?
 
     schema[target] = {} unless schema.key?(target)
-    if schemas.key?(:links)
-      schema[target]["links"] = merge_schemas(schemas[:links], schemas.fetch(:base_links))
-    else
-      schema[target]["links"] = embeddable_schema(schemas.fetch(:base_links))
-    end
+    schema[target]["links"] = merge_schemas(*order)
   end
 
   def merge_schemas(base_schema, *others)
--- a/lib/govuk_content_schemas/schema_combiner.rb
+++ b/lib/govuk_content_schemas/schema_combiner.rb
@@ -63,10 +63,14 @@ private
     end
   end
 
-  def merge_schemas(base_schema, other)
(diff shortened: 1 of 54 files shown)


lib/govuk_content_schemas/schema_combiner.rb
```diff
@@ -53,11 +53,18 @@ def add_details(schema)
...
     linkset_links = schemas[:links]
+    base_edition_links = schemas[:base_edition_links]
+    edition_links = schemas[:edition_links]
+
+    order = [
+      edition_links,
+      linkset_links,
+      base_edition_links,
+      base_linkset_links
+    ].compact
```
carlosmartinez: could this be replaced with

```
order = schemas.slice(:edition_links, :links, :base_edition_links, :base_links).compact
```

thomasleese: I don't think we can do that as it won't preserve the order when turning it into an array.

carlosmartinez: ah, right. could require ActiveSupport core extensions and do `slice.values`, but probably not worth it.

lib/tasks/combine_schemas.rake
```diff
@@ -30,7 +30,7 @@ end
 
 def sources_for_v2_schema(filename)
   Rake::FileList.new(
-    "formats/definitions.json",
+    "formats/{definitions,base_edition_links}.json",
```
carlosmartinez: what

thomasleese: To generate a v2 `schema.json` file, information from both `formats/definitions.json` and `formats/base_edition_links.json` is required.

lib/govuk_content_schemas/schema_combiner.rb
```diff
@@ -53,20 +53,31 @@ def add_details(schema)
...
 
-  def merge_schemas(base_schema, other)
+  def merge_schemas(base_schema, *others)
     merged_schema = embeddable_schema(base_schema)
-    other = embeddable_schema(other)
-    merged_schema["properties"] = merged_schema["properties"].merge(other["properties"])
+
+    others
+      .map { |other| embeddable_schema(other) }
+      .each { |other| merged_schema["properties"].merge!(other["properties"]) }
```
carlosmartinez: does the following work? i think it expresses the intent more clearly.

```
    other_properties = others.map do |other|
      embeddable_schema(other)["properties"]
    end
    merged_schema["properties"].merge!(other_properties))
```

carlosmartinez: merged_schema["properties"].merge!(other_properties.reduce({}, :merge)))

PR #549, opened 2017-03-02
Add short-url-manager to `dependentApplications`

h-lame:
Looks to me like the [short-url-manager JenkinsFile](https://github.com/alphagov/short-url-manager/blob/master/Jenkinsfile) is already setup to be run from here. I think this is all we do and we can remove the ci.dev schema_branches job for short-url-manager.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,6 +19,7 @@ def dependentApplications = [
   'policy-publisher',
   'publisher',
   'publishing-api',
+  'short-url-manager',
   'service-manual-frontend',
   'service-manual-publisher',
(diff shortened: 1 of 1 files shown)


h-lame:
Turns out, this needs a puppet change to actually get the build to pass.  See: https://github.com/alphagov/govuk-puppet/pull/5558

h-lame:
This is no longer failing because of short-url-manager; whitehall and manuals-publisher are the culprits.

h-lame:
Oh, yes, I wasn't paying attention to it.  Thanks for the rebase @tijmenb!

PR #550, opened 2017-03-02
Lock down the allowed document types

tijmenb:
This locks down the allowed values for `document_type` to a predefined list. The list of produced from the live content-store, so this should be backwards compatible.

Reasons for doing this:

- It becomes harder to inadvertently change the `document_type` of a document. Most recently we accidentally changed the document type of a number of mainstream documents to `generic_with_external_links` (alphagov/publisher#581). Because the document type is used by analytics, this affected the GOV.UK performance reporting.
- We can use the centrally defined list in the developer docs. We currently rely on search for this, but that isn't accurate because content items with a type like "redirect" or "gone" will never be in
the search index.
- The "document supertype" gem (https://github.com/alphagov/govuk_document_types) can use this list to validate the provided document types. This makes it easier to spot errors.

Note that because the current list is taken from the live content store it contains document types that are obviously wrong, such as the `placeholder` and `placeholder_ministerial_role` documents. This has to be fixed in the content store before removing them here.

Needs https://github.com/alphagov/publishing-api/pull/813 to be merged & deployed before green.

https://trello.com/c/07lAcDtC


cbaines:
Did you consider adding a smaller whitelist of document types to specific/all schemas? For example, quite a few schemas are only used for one document type.

lib/govuk_content_schemas/allowed_document_types.yml
```diff
@@ -0,0 +1,155 @@
+# Produced by running `y ContentItem.distinct(:document_type)` in the content
```
cbaines: Is the `y ` here a typo?

tijmenb: Nope, `y` will output the contents in YAML. Try it at home!

tijmenb:
@cbaines I considered it, but I thought it might be too much overhead when choosing a document type, and in the code that does the modification of the schemas. Also, I was thinking that there might not actually be a strong relationship between a document type and schema. I can't think of a good reason that they should be coupled.

cbaines:
> @cbaines I considered it, but I thought it might be too much overhead when choosing a document type, and in the code that does the modification of the schemas. Also, I was thinking that there might not actually be a strong relationship between a document type and schema. I can't think of a good reason that they should be coupled.

Makes sense :+1:

PR #553, opened 2017-03-06
Travel advice index countries as links

danielroseman:
Allow synonyms in travel advice items, and update example of travel
advice index to show expanded links rather than explicit countries
items in details.

(Test failures are in downstream builds and appear unrelated.)


fofr:
Is there a plan to use this sort of link on travel advice documents as well as the index?

PR #554, opened 2017-03-06
Create schema for publishing-api -> content-store interaction

tijmenb:
This adds a schema for the interaction between publishing-api and the content-store. It is an internal schema which describes what the publishing-api "PUT content" call looks like to the content store. 

With it we can prevent publishing-api from sending invalid things to the content store. On the other side we can make sure that the content store can deal with everything that the publishing-api sends.

cc @kevindew


tijmenb:
Going to close this for a bit to focus on other things, will revisit next week.

PR #555, opened 2017-03-06
Remove manuals-publisher as downstream app (temporarily)

tijmenb:
Manuals publisher tests are currently failing, holding up work on this repo. See https://github.com/alphagov/manuals-publisher/issues/816.


tijmenb:
Not necessary anymore since https://github.com/alphagov/manuals-publisher/pull/822.

PR #556, opened 2017-03-06
Add multiple and no organisations to contacts examples

mcgoooo:
this is so that all examples the content schema could have are covered


formats/contact/frontend/examples/contact_with_two_organisations.json
```diff
@@ -0,0 +1,171 @@
+{
+  "base_path": "/government/organisations/hm-revenue-customs/contact/customs-excise-and-vat-fraud-reporting",
```
fofr: This example has the same base path as the other one. They should be unique.

mcgoooo:
@fofr updated

tijmenb:
@mcgoooo is this PR still needed?

mcgoooo:
not right no, we do need to test this, but it broke a lot of things

PR #558, opened 2017-03-07
[DO NOT MERGE] Updates schema to html publications

elenatanasoiu:
For: https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents

This will add two more fields to be sent along to Publishing API from Whitehall. These two fields will be included in the print version of HTML Publications which are edited on whitehall, but will be displayed by frontend.

You can still see a draft of the HTML Publication on Whitehall, but the frontend app actually needs to display this information.

These schemas will signal to frontend to use two extra fields (that will be displayed in the print version of the HTML Publication). 

They are: `isbn` and `print_meta_data_contact_address`. Both fields are optional.

`isbn` - Identifies the Web ISBN to be displayed when printing an HTML Publication
`print_meta_data_contact_address` - Identifies the address of the institution which has produced the HTML Publication


formats/html_publication/publisher/details.json
```diff
@@ -21,6 +21,12 @@
     },
     "first_published_version": {
       "type": "boolean"
+    },
+    "isbn": {
+      "type": "string"
+    },
+    "print_meta_data_contact_address": {
```
fofr: I don't think this needs to have `print_` in the name, it may not always be used for print.
I'd note it's use in a description field.

Should the ISBN and contact address be displayed on screen for users as well as those printing?

fofr:
Should the ISBN and contact address be displayed on screen for users as well as those printing?

formats/html_publication/publisher/details.json
fofr: The descriptions you added to the PR should be included as descriptions in the schema

fofr:
Could you split the commits into two: one that adds the changes and one that generates the built files in `dist/`?

formats/html_publication/publisher/details.json
elenatanasoiu: This corresponds to a field created in the `whitehall` project. Here is the migration for it: https://github.com/alphagov/whitehall/blob/072d107581c98a97f5d0a6eff6232821892f4099/db/migrate/20170223180546_add_metadata_fields_to_attachments.rb

I had a discussion with @h-lame  about what to name this: 
screen shot 2017-03-09 at 15 32 41

It can be renamed, but it's only being used for the print version. 

See the Trello card for why it was needed: https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents

elenatanasoiu: I've added descriptions to the schema. Please see here: https://github.com/alphagov/govuk-content-schemas/pull/561/files

elenatanasoiu:
The ISBN and contact address are only visible on the print page because the purpose is to make the printed version of the HTML attachment be as similar as possible to the PDF version.

PDF version: 

screen_shot_2016-09-14_at_14 32 28

There is no requirement for them to appear on the screen, as stated in the Trello card: https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents

elenatanasoiu:
I've split the commits into two, please see here: https://github.com/alphagov/govuk-content-schemas/pull/561/commits

PR #561, opened 2017-03-09
[DO NOT MERGE] Updates schema to html publications

elenatanasoiu:
For: https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents

This will add two more fields to be sent along to Publishing API from Whitehall. These two fields will be included in the print version of HTML Publications which are edited on whitehall, but will be displayed by frontend.

You can still see a draft of the HTML Publication on Whitehall, but the frontend app actually needs to display this information.

These schemas will signal to frontend to use two extra fields (that will be displayed in the print version of the HTML Publication). 

They are: `isbn` and `print_meta_data_contact_address`. Both fields are optional.

`isbn` - Identifies the Web ISBN to be displayed when printing an HTML Publication
`print_meta_data_contact_address` - Identifies the address of the institution which has produced the HTML Publication

Is dependent on this PR: https://github.com/alphagov/govuk-content-schemas/pull/561/commits/312e10e87c5534a3151337a2b647310f55184eb1

### DIFF
--- a/formats/html_publication/publisher/details.json
+++ b/formats/html_publication/publisher/details.json
@@ -21,6 +21,14 @@
     },
     "first_published_version": {
       "type": "boolean"
+    },
+    "isbn": {
+      "type": "string",
+      "description": "Identifies the Web ISBN to be displayed when printing an HTML Publication"
+    },
+    "print_meta_data_contact_address": {
+      "type": "string",
+      "description": "Identifies the contact address of the institution which has produced the HTML Publication. To be displayed when printing an HTML Publication"
     }
   }
 }
--- a/dist/formats/html_publication/frontend/schema.json
+++ b/dist/formats/html_publication/frontend/schema.json
@@ -268,6 +268,14 @@
         "first_published_version": {
           "type": "boolean"
         },
+        "isbn": {
+          "type": "string",
+          "description": "Identifies the Web ISBN to be displayed when printing an HTML Publication"
+        },
+        "print_meta_data_contact_address": {
+          "type": "string",
+          "description": "Identifies the contact address of the institution which has produced the HTML Publication. To be displayed when printing an HTML Publication"
+        },
         "change_history": {
           "$ref": "#/definitions/change_history"
         }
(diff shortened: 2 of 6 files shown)


formats/html_publication/frontend/examples/prime_ministers_office.json
```diff
@@ -5,7 +5,9 @@
...
     "headings": "<ol></ol>",
     "public_timestamp": "2016-01-17T14:19:42.460Z",
-    "first_published_version": false
+    "first_published_version": false,
+    "isbn": "978-1-78386-459-1",
```
fofr: The example content items should reflect the live data? This looks like isbn and the contact address have been copied and pasted to each example.

formats/html_publication/publisher/details.json
```diff
@@ -21,6 +21,14 @@
...
     "first_published_version": {
       "type": "boolean"
+    },
+    "isbn": {
+      "type": "string",
+      "description": "Identifies the Web ISBN to be displayed when printing an HTML Publication"
+    },
+    "print_meta_data_contact_address": {
+      "type": "string",
+      "description": "Identifies the contact address of the institution which has produced the HTML Publication. To be displayed when printing an HTML Publication"
```
fofr: Good descriptions 👍

formats/html_publication/frontend/examples/prime_ministers_office.json
elenatanasoiu: @fofr The isbn field is not present in these examples, it's a field that's not normally filled in (but that people should be encouraged to use more often). And as for the print_meta_data_contact_address field, I've only just added the field so there's no way to reflect live data on it yet. 

So I have a chicken and egg problem here. 

Do you think I should just remove them from the examples?

fofr: I would add a specific example that would realistically have it set.

elenatanasoiu: Cool, thanks

nickcolley:
Can you let me know the best way to test these changes?

PR #562, opened 2017-03-10
Send draft links

gpeng:
The publishing API `put_content` endpoint now handles links within the JSON payload.

This PR updates all format schemas to allow links within the publisher `schema.json` file. So, now we generate two publisher files:

* schema.json - content and all links combined
* links.json - just links

As the use cases and distinction between edition based and old style `patch_links` flavour links develops we can use these files to differentiate between the two but currently the links in both files are just duplicated when we combine the schema files.

[Trello](https://trello.com/c/brKgEYBl/622-draft-links-3-days)

### DIFF
--- a/dist/formats/answer/publisher_v2/schema.json
+++ b/dist/formats/answer/publisher_v2/schema.json
@@ -305,6 +305,43 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "taxons": {
+          "description": "Prototype-stage taxonomy label for this content item",
+          "$ref": "#/definitions/guid_list"
+        },
+        "ordered_related_items": {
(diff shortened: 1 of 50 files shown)


tijmenb:
Would we be able to restrict the links in `put_content`? It looks like https://github.com/alphagov/govuk-content-schemas/pull/543 added `base_edition_links.json` for this purpose.

kevindew:
Just to expand on Tijmens point you probably want to define them in an `edition_links.json` file as per [specialist document](https://github.com/alphagov/govuk-content-schemas/tree/master/formats/specialist_document/publisher)?

gpeng:
Ah.. ok. Didn't see that. I'll close this then. Thanks.

PR #564, opened 2017-03-10
Run frontend tests

cbaines:


Jenkinsfile
```diff
@@ -4,14 +4,15 @@ REPOSITORY = 'govuk-content-schemas'
 
 def dependentApplications = [
   'businesssupportfinder',
-  'contacts',
-  'contacts-frontend',
   'collections',
   'collections-publisher',
+  'contacts',
```
36degrees: 🔤 ⏬ 👌

cbaines:
> These changes look fine, but I think you need to change the Jenkinsfile in frontend so that it passes the SCHEMA_BRANCH env when calling contentSchemaDependency – otherwise it'll always checkout the deployed-to-production branch of schemas.

Good spot, I've put up https://github.com/alphagov/frontend/pull/1167

36degrees:
I've approved it, but because I'm not a contributor it means zilch. 😢 

I tried!

cbaines:
> I've approved it, but because I'm not a contributor it means zilch. :cry:
> 
> I tried!

Thanks @36degrees you can be assured that at least I value your efforts and skill!

cbaines:
The tests for frontend are failing at the moment, I've proposed https://github.com/alphagov/frontend/pull/1168 to fix this.

cbaines:
Even more of the setup is missing! I think the frontend tests are now passing, but https://github.com/alphagov/govuk-puppet/pull/5605 will need merging before this will work.

At that point, the schema testing will work, but until https://github.com/alphagov/frontend/releases/tag/release_2079 is deployed, the tests will be using the wrong branch of govuk-content-schemas.

PR #566, opened 2017-03-10
Licence format schema

davidbasalla:
We add two examples for licences, one with continuation link which will link off to an external site and one without continuation link which will make a request to the Licensify API for extra licence details.

### DIFF
--- a/formats/licence/frontend/examples/licence_without_continuation_link.json
+++ b/formats/licence/frontend/examples/licence_without_continuation_link.json
@@ -1,0 +1,248 @@
+{
+  "analytics_identifier": null,
+  "base_path": "/temporary-events-notice",
+  "content_id": "cb16a948-d4c9-4e55-99b9-2f3931481b07",
+  "document_type": "licence",
+  "first_published_at": "2016-02-29T09:24:10.000+00:00",
+  "format": "licence",
+  "locale": "en",
+  "navigation_document_supertype": "guidance",
+  "need_ids": [
+
+  ],
+  "phase": "live",
+  "public_updated_at": "2015-04-27T14:48:11.000+00:00",
+  "publishing_app": "publisher",
+  "rendering_app": "frontend",
+  "schema_name": "licence",
+  "title": "Temporary Events Notice (England and Wales)",
+  "updated_at": "2017-03-10T16:01:40.758Z",
+  "user_journey_document_supertype": "thing",
+  "withdrawn_notice": {
+  },
+  "links": {
+    "mainstream_browse_pages": [
+      {
+        "analytics_identifier": null,
+        "api_path": "/api/content/browse/business/licences",
+        "base_path": "/browse/business/licences",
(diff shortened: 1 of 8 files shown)


issyl0:
For some reason GitHub thinks these commits are the wrong way around, but locally they're the correct way around:

```
* 188e0ed Fri, 10 Mar 2017 16:22:32 +0000 | Generate dist for licence format (HEAD -> licence_format_schema, origin/licence_format_schema) [David Basalla]
* 63d6435 Fri, 10 Mar 2017 14:44:43 +0000 | Add a schema for the Licence format [Isabell Long]
```

davidbasalla:
hi @fofr do you mind having a look when you get the chance? Thank you!

formats/licence/publisher/details.json
```diff
@@ -0,0 +1,32 @@
...
+    "licence_identifier"
+  ],
+  "properties": {
+    "will_continue_on": {
+      "description": "Text to follow the statement 'This will continue on'.",
+      "type": "string"
+    },
+    "continuation_link": {
+      "description": "Link to licence competent authority.",
+      "type": "string"
```
fofr: If this is a link, should type be URI?

formats/licence/publisher/details.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "licence_identifier"
```
fofr: As `licence_identifier` is the only required field, what does the content look like when only this is provided?

formats/licence/frontend/examples/licence_without_continuation_link.json
```diff
@@ -0,0 +1,250 @@
...
+        "links": {
+        }
+      }
+    ]
+  },
+  "description": "You need a Temporary Event Notice if you want to carry out a 'licensable activity' on unlicensed premises in England or Wales",
+  "details": {
+    "licence_identifier": "1071-5-1",
+    "will_continue_on": "",
+    "continuation_link": "",
```
fofr: Should these empty string fields be omitted?

issyl0:
@fofr I think the commit order being the wrong way round is GitHub's fault - locally they seem to be the correct way around: https://github.com/alphagov/govuk-content-schemas/pull/566#issuecomment-285716991. We'll see if GitHub corrects the order when we push changes based on your comments.

formats/licence/publisher/details.json
davidbasalla: Depending on the type of licence (as defined by Licensify via the licence identifier), it might:
- redirect to an authority show page with different actions (eg apply, renew)
- show the title of the licence and a postcode search form
- show a radio button for multiple licence authorities

There's an Overview header currently being rendered by Frontend, which we'll update to be conditional based on presence of 'licence_overview'. Publisher itself only validates the presence of the licence identifier.

davidbasalla:
@fofr thanks for the review, have updated based on your comments

PR #570, opened 2017-03-15
Remove deprecated countries hash from travel index.

danielroseman:
This is no longer being sent by travel-advice-publisher or expected by frontend.


cbaines:
If possible, it would be good to rebase this on to https://github.com/alphagov/govuk-content-schemas/pull/564 when it has merged, as that would mean that the frontend tests run.

cbaines:
> If possible, it would be good to rebase this on to #564 when it has merged, as that would mean that the frontend tests run.

This has finally merged :fireworks: so if you rebase, the frontend test suite should run.

danielroseman:
@cbaines rebased.

PR #573, opened 2017-03-16
Add links for email to migrated formats

bevanloon:
Adds `people` and `roles` to `links.json` for the following formats:

* Consultation
* Fatality Notice
* News Article
* Publication
* Speech

In order to send emails, Email Alert Api relies on the presence of links in data that the Publishing Api sends to the rabbit queues. Whitehall in turn needs to present these links to the Publishing Api. Roles and People are currently required but missing. This will allow Email Alert Api to take over from Govuk Delivery and ultimately allow us to retire Govuk Delivery.

There will be a corresponding PR in Whitehall to present these links to the Publishing Api (currently a work in progress on this branch https://github.com/alphagov/whitehall/tree/add-links-for-email-to-migrated-formats).

[Trello](https://trello.com/c/WizHAZBT)

### DIFF
--- a/dist/formats/consultation/frontend/schema.json
+++ b/dist/formats/consultation/frontend/schema.json
@@ -93,6 +93,14 @@
         "topical_events": {
           "$ref": "#/definitions/frontend_links"
         },
+        "roles": {
+          "description": "Used to power Email Alert Api subscriptions for Whitehall content",
+          "$ref": "#/definitions/frontend_links"
+        },
+        "people": {
+          "description": "Used to power Email Alert Api subscriptions for Whitehall content",
+          "$ref": "#/definitions/frontend_links"
+        },
         "taxons": {
           "description": "Prototype-stage taxonomy label for this content item",
           "$ref": "#/definitions/frontend_links"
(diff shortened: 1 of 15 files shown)


fofr:
Can you elaborate on what the email API is using it for?

fofr:
My fear is that a future person may see that these fields are not being used by any frontend and will remove them. How do we stop that happening?

Data needed by email has typically been a little opaque (see `tags` in the details hash)

bevanloon:
That's a good point. I think once they are part of the schemas proper and Whitehall is presenting them, the Whitehall tests should start failing if they are removed. If I'm right, part of the Content Schemas build job will run the tests against Whitehall's deployed to production branch (https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+59%3A+Workflow+for+making+changes+to+the+schemas). In which case, we should get failures on things like https://github.com/alphagov/whitehall/blob/bead1062bab00be5cee4b7d30e9f05d3bd95dcad/test/unit/presenters/publishing_api/speech_presenter_test.rb#L30 should fail.

Maybe not foolproof, but it might be good enough?

tijmenb:
Would it be possible to add a small description to the definitions? Just something like "Used to power email alert subscriptions in Whitehall" would go a long way.

bevanloon:
👍 @tijmenb added a description into the links.

PR #576, opened 2017-03-17
Add "email" and "government" supertypes

dwhenry:
These replace the "whitehall" super type and will be
used for emailing changes to whitehall content.


tijmenb:
Looks good! Can you split the commits into one with the config change and another one for with the schema regeneration?

dwhenry:
@tijmenb done

PR #582, opened 2017-03-17
Rewrite schema generation

tijmenb:
Paired with @kevindew on this.

This rewrites the schema generation to be easier to follow, faster and easier to extend.

In this PR we go out of our way to keep the `/dist` output in of the rewrite exactly the same as the old generator. This is to make sure that we aren't changing the functionality inadvertently and to make reviewing easy. In a follow-up PRs we can make some of the code nicer and refactor some more.

Needs:

- [x] https://github.com/alphagov/govuk-content-schemas/pull/577 
- [x] https://github.com/alphagov/govuk-content-schemas/pull/579 
- [x] https://github.com/alphagov/govuk-content-schemas/pull/580
- [x] https://github.com/alphagov/govuk-content-schemas/pull/581


lib/schema_generator/publisher_content_schema_generator.rb
```diff
@@ -0,0 +1,118 @@
...
+    end
+
+  private
+
+    attr_reader :schema_name
+
+    def required
+      required_properties = base_content_item["required"] + %w[document_type schema_name]
+
+      # TODO: `contact` is a content item that is can be base path less.
```
MatMoore: > is can be

"is" or "can be"?

tijmenb:
Merged including the typo because I'm afraid I'll be stuck building CI for the rest of the afternoon.

PR #584, opened 2017-03-20
Send current commit sha to downstream jobs.

danielroseman:
Set a flag on those apps that are using the centralised buildProject
script. If so, don't wait for them to complete; they will report their
status individually to the triggering PR (https://github.com/alphagov/govuk-puppet/pull/5625).

https://trello.com/c/5psKmk85

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,33 +2,36 @@
 
 REPOSITORY = 'govuk-content-schemas'
 
+// second boolean parameter describes whether the app is using the central
+// buildProject script - if so it will report its success individually, so this
+// job does not need to wait for it.
 def dependentApplications = [
-  'businesssupportfinder',
-  'collections',
-  'collections-publisher',
-  'contacts',
-  'content-tagger',
-  'email-alert-frontend',
-  'email-alert-service',
-  'finder-frontend',
-  'frontend',
-  'hmrc-manuals-api',
-  'licencefinder',
-  'manuals-frontend',
-  'manuals-publisher',
-  'policy-publisher',
-  'publisher',
-  'publishing-api',
-  'short-url-manager',
-  'service-manual-frontend',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -2,34 +2,37 @@
 
 REPOSITORY = 'govuk-content-schemas'
 
+// second boolean parameter describes whether the app is using the central
+// buildProject script - if so it will report its success individually, so this
+// job does not need to wait for it.
 def dependentApplications = [
```
thomasleese: Is it possible to use some sort of hash literal to avoid the need for a comment?

```groovy
[app: 'businesssupportfinder', usesBuildProject: false]
```

Perhaps that's too verbose though?

Jenkinsfile
```diff
@@ -119,7 +127,10 @@ stage("Check dependent projects against updated schema") {
           [$class: 'StringParameterValue',
             name: 'SCHEMA_BRANCH',
             value: env.BRANCH_NAME],
-        ]
+          [$class: 'StringParameterValue',
+            name: 'SCHEMA_COMMIT',
+            value: env.GIT_COMMIT]
+        ], wait: wait
```
tijmenb: Will this make `build` return immediately? If that's the case the statsd timing below won't work anymore.

danielroseman: Yes, this is the opposite side from the govuk-puppet change that allowed the downstream builds to report to Github individually. Could this timing also go into the buildProject job?

tijmenb: Yeah, I think we could do something like that. Probably best if we remove the timers in this PR to avoid confusion.

Jenkinsfile
tijmenb: Hopefully we can get rid of the second param again soon, once everybody has started using `govuk.buildProject()`.

PR #585, opened 2017-03-20
Edition links in frontend/notification schemas

kevindew:
This is built on top of https://github.com/alphagov/govuk-content-schemas/pull/582 so won't be mergable until that is merged in and this is rebased against any changes.

This includes edition links inside frontend and notification schemas.

It also includes description and maxItems values in those schemas. Logic being that description is useful for people reading these schemas to know the purpose of the links, maxItems is useful as can anticipate how to handle the link. Originally I had it include required and minItems but I realised that we can't guarantee these are included. If this aspect of this PR is controversial let me know and I can break it out into a separate one for discussion.

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -81,51 +81,67 @@
       "additionalProperties": false,
       "properties": {
         "taxons": {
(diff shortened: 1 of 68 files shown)


kevindew:
@tijmenb are you happy with this?

PR #586, opened 2017-03-20
[DO NOT MERGE] Convert hard-coded special_route schema into generated schema

suzannehamilton:
Replace `schema.json` for the `special_route` schemas with the `metadata` and `links` components which will be used to generate the full schemas.

This makes `special_route` more flexible. For example, it can now have any valid `document_type`, which will let us give the homepage and search pages their own document types to use in analytics.

https://trello.com/c/u1H1rgDA/525-some-finding-pages-are-still-being-tracked-as-thing

Note that this adds links to the `special_route` schema. Is that OK? Is there some way of overriding the allowed links with an empty array? It didn't look like it, but I might have missed something in the schema generation code.

cc @tijmenb


suzannehamilton:
Closing this PR because I think we'll need to make the change in two steps to keep the schema tests green.

PR #593, opened 2017-03-24
Add locale and details to special_route schema

suzannehamilton:
- Delete publisher v1 version of the `special_route` schema, since the v1 API has been retired
- Add optional properties to the `special_route` publisher schema which match the properties which will be generated when `special_route` is migrated from a hard-coded schema to an automatically generated one.

This is step 1 of a process of making `special_route` consistent with the other schemas, which will allow it to use the globally-defined list of allowed document types:

1. Add `locale` and `details` as optional properties, and hard-code their definitions
2. Update other applications to provide these properties when publishing a `special_route`
3. Replace the hard-coded `schema.json` with the schema components like `metadata.json`, which will make `locale` and `details` mandatory fields. This step will remove the duplicated locale definitions.

Trello: https://trello.com/c/u1H1rgDA/525-some-finding-pages-are-still-being-tracked-as-thing

cc @tijmenb


cbaines:
Does the allowing of special routes to have details mean that there is a need for special routes to have details in some cases?

suzannehamilton:
Special routes don't specifically require it, but it's a required field in the [base metadata](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/metadata.json).

I guess we could rewrite the schema generation so that it was only a required field on some of the schemas, but it's used by pretty much everything so it seems less likely to cause bugs if we keep it as a required field everywhere and just make it an empty property on `special_route`. That's what `generic` does, which is the only other schema I've seen that doesn't add any extra details.

Does that make sense? I'm hoping that having this redundant property actually makes things simpler overall, but I might be missing a better solution.

cbaines:
> Does that make sense? I'm hoping that having this redundant property actually makes things simpler overall, but I might be missing a better solution.

Mmmm, I'm happy to defer to your judgement. If some document_types don't need details or locale then it would be neater to not require having them, but I don't really mind. It is just a schema.

PR #597, opened 2017-03-27
Add info-frontend to downstream builds

suzannehamilton:
The info-frontend project depends on the schemas, so it should be built when the schemas change.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -109,7 +109,7 @@ stage("Check dependent projects against updated schema") {
     dependentBuilds[app] = {
       start = System.currentTimeMillis()
 
-      build job: "${app}/deployed-to-production",
+      build job: "/${app}/deployed-to-production",
         parameters: [
           [$class: 'BooleanParameterValue',
             name: 'IS_SCHEMA_TEST',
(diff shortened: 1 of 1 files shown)


gpeng:
I think that error you were seeing in the commit message for e06acdc might have been fixed by the puppet deploy I did earlier. I had to do [this](https://github.com/alphagov/govuk-puppet/pull/5654)  (which is what I deployed this morning) to get my `government-frontend` version of this PR to work.

suzannehamilton:
@tijmenb - any thoughts on the [second commit](https://github.com/alphagov/govuk-content-schemas/pull/597/commits/c8ff1341337fed31dfaea09005ab9b3e62c59120) that I've just added to fix the schema tests?

It fixes an error which happens when you have a content schema branch with the same name a downstream project. I named this branch `info-frontend` rather than `add-info-frontend`, not realising it would cause a problem.

suzannehamilton:
@gpeng Sorry, I posted my last comment before seeing yours. I think the issue is the branch name being the same as the downstream error - it gives the same "No job found" error because it's looking in the wrong directory, and I've reproduced it with a different project to check.

PR #598, opened 2017-03-28
[WIP] Finally remove format

tijmenb:
This removes `format` from the frontend representation. 

This PR is expected to fail until content-store is updated, because it's still sending `format`.

- [ ] Should fail on content-store
- [ ] Fix content-store
- [ ] Fix service-manual-frontend


tijmenb:
Picking this up in a couple of weeks.

PR #604, opened 2017-04-05
Assign lgil codes

binaryberry:
All local transactions will be assigned to a lgil_code by default. The term
"override" is therefore misleading and we should rename this variable.

This leaves lgil_override as a deprecated field until local transactions have been
republished.

https://trello.com/c/RnJC1d7a/5-assign-lgil-codes-to-all-local-transactions


issyl0:
"Schemae"! ❤️

sihugh:
Although I guess we should fix the failing test :-)

PR #611, opened 2017-05-03
Better descriptions for dates

tijmenb:
Some extra docs that hopefully make it clearer what the dates mean.

### DIFF
--- a/formats/definitions.json
+++ b/formats/definitions.json
@@ -540,17 +540,17 @@
       }
     },
     "first_public_at": {
-      "description": "The date the content was first published. Used in details. Will be deprecated in favour of top level first_published_at when publishing API allows it to be edited.",
+      "description": "DEPRECATED. The date the content was first published. Used in details. Will be deprecated in favour of top level first_published_at when publishing API allows it to be edited.",
       "type": "string",
       "format": "date-time"
     },
     "first_published_at": {
-      "description": "The date the content was first published",
+      "description": "The date the content was first published.  Automatically determined by the publishing-api, unless overridden by the publishing application.",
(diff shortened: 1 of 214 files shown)


formats/definitions.json
```diff
@@ -540,17 +540,17 @@
       }
     },
     "first_public_at": {
-      "description": "The date the content was first published. Used in details. Will be deprecated in favour of top level first_published_at when publishing API allows it to be edited.",
+      "description": "DEPRECATED. The date the content was first published. Used in details. Will be deprecated in favour of top level first_published_at when publishing API allows it to be edited.",
```
fofr: Can drop the `will be deprecated` part.

formats/definitions.json
```diff
@@ -540,17 +540,17 @@
...
       "type": "string",
       "format": "date-time"
     },
     "first_published_at": {
-      "description": "The date the content was first published",
+      "description": "The date the content was first published.  Automatically determined by the publishing-api, unless overridden by the publishing application.",
```
fofr: Want to include something about it being set to `now` if omitted?

formats/definitions.json
```diff
@@ -540,17 +540,17 @@
...
       "type": "string",
       "format": "date-time"
     },
     "public_updated_at": {
-      "description": "When the content was last significantly changed, date shown to users",
+      "description": "When the content was last significantly changed (a major update). Shown to users.  Automatically determined by the publishing-api, unless overridden by the publishing application.",
```
fofr: Should we mention that this is usually accompanied with a change note?
How does the publishing-api determine this?

PR #612, opened 2017-05-08
Add Visiblity property to taxon schema

klssmith:
For Taxons that are ready to come out of development but shouldn't be viewable by the general public yet. Use the `visible_to_departmental_editors` flag to make the `draft` taxon appear in the `Draft Taxonomies` section of the new tagging interface. 

Note that this is only relevant for taxons at the root of their taxonomy tree.

[Trello](https://trello.com/c/3xt6aNdb/73-make-certain-taxons-visible-as-drafts-to-whitehall-editors)

### DIFF
--- a/dist/formats/taxon/frontend/schema.json
+++ b/dist/formats/taxon/frontend/schema.json
@@ -438,6 +438,10 @@
           "type": "string",
           "description": "Usage notes for editors when tagging content to a taxon."
         },
+        "visible_to_departmental_editors": {
+          "type": "boolean",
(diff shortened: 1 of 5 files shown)


tijmenb:
Because existing draft taxons should be hidden, maybe it's better to flip the meaning. Something like `visible_to_departmental_editors`?

whoojemaflip:
@tijmenb Yes, the lifecycle of a `taxon` becomes:

`Draft + Hidden`, `Draft + not hidden`, `Published + not hidden`

ie. Existing published taxons` should not be hidden by default.

We're working up a script to republish all the taxons with this additional field (whatever it ends up being).

PR #613, opened 2017-05-10
Add optional publishing_request_id frontend field

thomasleese:
This will contain the value of the request ID header, similar to how Travel Advice Publisher currently has `publishing_request_id` in the details hash.

Do not merge until https://github.com/alphagov/info-frontend/pull/68 has been merged.

[Trello Card](https://trello.com/c/JZ5Zabe9/702-remove-publishing-request-id-from-travel-advice-publisher-details-hash-2)

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -336,6 +336,9 @@
     "updated_at": {
       "type": "string",
       "format": "date-time"
+    },
+    "publishing_request_id": {
+      "$ref": "#/definitions/publishing_request_id"
     }
   },
   "definitions": {
--- a/dist/formats/answer/notification/schema.json
+++ b/dist/formats/answer/notification/schema.json
@@ -280,6 +280,9 @@
       "type": "string",
       "format": "date-time"
     },
+    "publishing_request_id": {
+      "$ref": "#/definitions/publishing_request_id"
+    },
     "routes": {
       "type": "array",
       "minItems": 1,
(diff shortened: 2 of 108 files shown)


cbaines:
What is the motivation behind adding this?

thomasleese:
@cbaines TAP is already sending a similar field, the aim was to make this happen across all the apps.

cbaines:
> @cbaines TAP is already sending a similar field, the aim was to make this happen across all the apps.

From what I remember/understand from the Travel Advice Publisher use case, it was misusing the govuk request id. Do you know what the use case for this data being in the content store is for Travel Advice Publisher related content?

thomasleese:
@cbaines It is actually no longer used by the frontend apps, only by some internal TAP publication checks, more information [in the Trello Card](https://trello.com/c/JZ5Zabe9/702-remove-publishing-request-id-from-travel-advice-publisher-details-hash-2).

tijmenb:
I'm also unsure why we would need to add this, who will use it?

If we do need it, the rename makes it less clear probably. I think it was called `publishing_request_id` to make it clear that this is the ID of the request that made the `/publish` call to the publishing-api. Maybe calling it `govuk_request_id` doesn't reveal where it comes from and is likely confused with the actual `govuk_request_id` of the request to the content store.

gpeng:
It was named `publishing_request_id` for the reasons @tijmenb says as we wanted to use it to marry up the publishing request ie the post when the publisher pushed the button to the payload arriving at the frontend. We exposed this in a meta tag in the body of the travel advice. (all of which I just realised is already in the Trello card... 🤕 ) I think it's more explicit to keep that naming (`publishing_request_id`) if the intention is to pass it through from the publishing apps and allow this kind of functionality for all formats.

thomasleese:
So it looks like Travel Advice Publisher will still be using the field, but it's also useful to allow other apps to perform checks in the future that Travel Advice Publisher currently does.

PR #628, opened 2017-06-19
Add world location schema

gpeng:
WorldLocations are no longer going to be rendered by Whitehall (with the exception of the international delegation type) as they are going to be replaced by taxons. They do still need to be linkable by other content though as other things depend on this assocation e.g. email alerts.

They are currently sent as placeholders to the publishing API. This commit creates a new schema for them (which is just a placeholder copy with the name changed) and then removes the requirement for a base path from  that schema.

[Trello](https://trello.com/c/6tio72jl/221-split-world-locations-from-international-delegations-in-whitehall)

### DIFF
--- a/lib/schema_generator/publisher_content_schema_generator.rb
+++ b/lib/schema_generator/publisher_content_schema_generator.rb
@@ -22,8 +22,9 @@ module SchemaGenerator
     def required
       required_properties = base_content_item["required"] + %w[document_type schema_name]
 
-      # TODO: `contact` is a content item that is can be base path less.
-      if schema_name == "contact"
+      # TODO: `contact` and `world_location` are schema types that can be base path less.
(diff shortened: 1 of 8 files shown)


lib/schema_generator/publisher_content_schema_generator.rb
```diff
@@ -23,7 +23,8 @@ def required
       required_properties = base_content_item["required"] + %w[document_type schema_name]
 
       # TODO: `contact` is a content item that is can be base path less.
```
rubenarakelyan: Minor comment: we should update this comment to be consistent with the code (also, it is currently worded incorrectly so that can also be fixed!)

gpeng: Fixed in 041637c. I'm not sure we really need that comment at all. I did consider deleting it.

PR #629, opened 2017-06-23
Update world locations examples and FE schemas

gpeng:
`world_locations` links can now be base_path less. This PR updates the examples to reflect this and also removes translations for them which we are dropping support for.

The existing `frontend_links` definition requires `base_path` so this PR also adds a `base_path_less_frontend_links` definition which is used when the `world_locations` are included as the frontend schemas are generated. This could probably also be used for `contacts` and `governments` but they don't appear in the examples so I've omitted that for now.

Related PRs:

https://github.com/alphagov/publishing-api/pull/953
https://github.com/alphagov/content-store/pull/299
https://github.com/alphagov/government-frontend/pull/386

Trello:

https://trello.com/c/6tio72jl/221-split-world-locations-from-international-delegations-in-whitehall


gpeng:
@kevindew I've fixed the `additionalAttributes` as discussed on https://github.com/alphagov/content-store/pull/299

PR #631, opened 2017-06-29
Deprecate "headings" on HTML publications

fofr:
Formerly a list of headings used to display a contents list, government-frontend no longer uses this data – instead it pulls headings out of the content and constructs a contents list that it can manipulate. This matches other contents lists in the app.

See: alphagov/government-frontend#384

@nickcolley @andysellick


fofr:
Closing while we investigate bugs with number formatting in new contents lists.

fofr:
This is good to review now.

PR #633, opened 2017-07-05
Add "start_button_text" to transaction schema

binaryberry:
This will allow to pass in customised text for start buttons. This also adds
child-maintenance as an example of a transaction with a custom start button, and extracts the start_button_text as a definition since it's also being used by the simple_smart_answer schema.

cc @chrisroos @h-lame @nickcolley 

@pmanrubia @ikennaokpala are you or someone from your team available to review?

https://trello.com/c/arlZczQO/423-2-republish-smart-answer-part-year-profits-to-finalise-your-tax-credits-start-page-as-a-transaction-page

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -1562,6 +1562,10 @@
     "will_continue_on": {
       "description": "Description of the website the adjoining external link will be taking the user to",
       "type": "string"
+    },
+    "start_button_text": {
+      "description": "Custom text to be displayed on the green button that takes you to another page",
+      "type": "string"
     }
   }
 }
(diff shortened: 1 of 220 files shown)


fofr:
Minor comment: `start_button_text` doesn't (at first inspection) indicate that it's custom and overrides a default. The description does elaborate though, and a frontend can assume that the lack of this field means it should display "Start now".

We could always pre-populate this field with "Start now" in the future if we wanted.

binaryberry:
@ikennaokpala : Thanks! I now removed the "to be squashed" mention which I had left by mistake - I intended to keep it a separate commit. Also used "schemas" instead of "schemae" for ed41fcb now.

@fofr : true, and `custom_start_button_text` would have been more descriptive.  Renaming it now would imply also making changes to frontend and publisher for the simple smart answers format (which also uses that field), and for the transaction format to frontend and the smart answers PR, so I'd rather avoid it if possible. We did discuss with @nickcolley the option of sending "Start now" by default from smart-answers but it felt more natural to have frontend default to "Start now" when no text was provided.

PR #634, opened 2017-07-07
New approach to defining Schemas

kevindew:
Hello :wave: During the last firebreak (June 2017) I've worked on a new approach to defining schemas. This achieves a few things that improve schemas:

1. All schemas can be generated from a single file per-format, even the exceptional cases (redirects/gone)
2. Notification and frontend are much more accurate (frontend could be even more accurate, but examples need fixing)
3. Usage of a format ([jsonnet](https://github.com/google/jsonnet)) that allows importing for re-use of common definitions and to try avoid duplications. We can also use comments in the definitions 🎉 
4. Definitions are generated just when needed.

This approach has been documented in an [RFC](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-080-jsonnet-based-single-file-schemas.md) which has been accepted. 

As of late September 2017, this approach now passes tests for all dependent repos (which has involved some accuracy compromises).

I've tried to fix everything that I think could be broken by this, but I'd not be surprised if something gets caught on it. The most likely reason for this is the change of the examples directory, which has required updates in [govuk-content-schema-test-helpers](https://github.com/alphagov/govuk-content-schema-test-helpers/pull/27) and [govuk_schemas](https://github.com/alphagov/govuk_schemas/pull/37).

I'm due to go on leave from September 28th until October 9th so if/once this gets approved I wouldn't plan to merge it until October 9th so I can be about if it breaks things.


kevindew:
@tijmenb reflecting on the problem regarding the document types yaml import, I've put together an approach to still allowing this via a document_types file created as part of schema generation in https://github.com/alphagov/govuk-content-schemas/pull/634/commits/4321416d5a6de18332ea4f7867d452f2d3b5354f which will output a dist/allowed_document_types.yml file as per https://github.com/alphagov/govuk-content-schemas/blob/054a457cc312cacdac5cbe2e064daf3edddf5b73/dist/allowed_document_types.yml

What do you think?

tijmenb:
@kevindew that looks good! The docs would be able to work with this. Btw I forgot that the tests for https://github.com/alphagov/govuk_document_types/blob/master/spec/data_lint_spec.rb#L39 also use the YAML file. 

One thing that I hadn't realised is that this makes it somewhat easier to add document types (by adding them to the schema right?). This may make it easier to have the same document types with a slightly different name.

kevindew:
@tijmenb well remembered, I'll get something in on that too when we merge.

Yes there's a risk of that. I'll have a little ponder over that.

tijmenb:
Yep, don't let that hold you off from merging!

PR #637, opened 2017-07-14
Consistent ordering in schemas

kevindew:
This specifies a set of rules of how a schema should be ordered and applies them when schemas are generated. 

This is to make the output schemas to be less determined by their inputs (which has led to things like [this](https://github.com/alphagov/govuk-content-schemas/blob/master/lib/schema_generator/frontend_schema_generator.rb#L105-L114)) and instead be predictable in their output.

The rules are:

- For an object with a key of "properties" the fields are sorted alphabetically

Otherwise:

- A description field is at the top
- Followed by any fields prefixed with a $ (eg $ref)
- Then we have the type field
- Followed by required
- Other fields are alphabetical
- Any definitions are at the end

The reason this is a somewhat complex ordering rather than a simple alphabetical one is that schema files end up being swamped by "definitions".

### DIFF
--- a/lib/schema_generator/schema.rb
+++ b/lib/schema_generator/schema.rb
@@ -1,15 +1,65 @@
 # Read and write the schemas
 module SchemaGenerator
+  module SortedHash
+    refine Hash do
+      def deep_sort(&block)
+        recursive_deep_sort(self, &block)
+      end
+
+    private
+
+      def recursive_deep_sort(property, parent_key = nil, &block)
+        if property.is_a?(Hash)
+          children_sorted = property.each_with_object({}) do |(k, v), memo|
+            memo[k] = recursive_deep_sort(v, k, &block)
+          end
+
+          sorted = children_sorted.sort do |a, b|
+            block.call(a, b, parent_key)
+          end
+
+          Hash[sorted]
+        elsif property.respond_to?(:map)
+          property.map { |item| recursive_deep_sort(item, nil, &block) }
+        else
+          property
+        end
+      end
+    end
+  end
+
   class Schema
+    using SchemaGenerator::SortedHash
+
     def self.read(filename)
       @parsed_json_cache ||= {}
       @parsed_json_cache[filename] ||= JSON.parse(File.read(filename))
     end
 
     def self.write(full_filename, schema_hash)
-      schema_json = JSON.pretty_generate(schema_hash) + "\n"
+      schema_json = JSON.pretty_generate(ordered_schema(schema_hash)) + "\n"
       FileUtils.mkdir_p(File.dirname(full_filename))
       File.write(full_filename, schema_json)
+    end
+
+    def self.ordered_schema(schema_hash)
+      # Custom consistent sorting for JSON Schema objects
+      sorted_hash = schema_hash.deep_sort do |(a, _), (b, _), parent_key|
+        a, b = [a.to_s, b.to_s]
+        # We don't want to sort any items properties
+        next a <=> b if parent_key == "properties"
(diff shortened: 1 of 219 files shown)


lib/schema_generator/schema.rb
```diff
@@ -1,15 +1,65 @@
...
     end
 
     def self.write(full_filename, schema_hash)
-      schema_json = JSON.pretty_generate(schema_hash) + "\n"
+      schema_json = JSON.pretty_generate(ordered_schema(schema_hash)) + "\n"
       FileUtils.mkdir_p(File.dirname(full_filename))
       File.write(full_filename, schema_json)
     end
+
+    def self.ordered_schema(schema_hash)
```
fofr: Some tests would clarify this behaviour.

lib/schema_generator/schema.rb
```diff
@@ -1,15 +1,65 @@
...
 module SchemaGenerator
+  module SortedHash
+    refine Hash do
+      def deep_sort(&block)
+        recrusive_deep_sort(self, &block)
+      end
+
+    private
+
+      def recrusive_deep_sort(property, parent_key = nil, &block)
```
fofr: Typo here

kevindew: Thanks 👍

lib/schema_generator/schema.rb
kevindew: Good point - I'll add some

kevindew:
@fofr I've addressed both of your comments.

lib/schema_generator/schema.rb
```diff
@@ -1,15 +1,65 @@
...
     def self.write(full_filename, schema_hash)
-      schema_json = JSON.pretty_generate(schema_hash) + "\n"
+      schema_json = JSON.pretty_generate(ordered_schema(schema_hash)) + "\n"
       FileUtils.mkdir_p(File.dirname(full_filename))
       File.write(full_filename, schema_json)
     end
+
+    def self.ordered_schema(schema_hash)
+      # Custom consistent sorting for JSON Schema objects
+      sorted_hash = schema_hash.deep_sort do |(a), (b), parent_key|
```
thomasleese: Do we need `a` and `b` in brackets? Looks a bit odd, and if we do, I'm not sure why.

kevindew: So when iterating the hash it iterates as arrays of keys and values.

eg:
```
[4] pry(main)> { k1: "v1", k2: "v2" }.map { |a| a }
=> [[:k1, "v1"], [:k2, "v2"]]
```
so options we have here to make it easier to understand are like:

```
sorted_hash = schema_hash.deep_sort do |(key_for_a, _), (key_for_b, _), parent_key|
```

or 
```
sorted_hash = schema_hash.deep_sort do |a, b, parent_key|
  key_for_a = a.first.to_s
  key_for_b = b.first.to_s
```

Not really sure if they help a whole lot though?

kevindew: does just: `sorted_hash = schema_hash.deep_sort do |(a, _), (b, _), parent_key|` make life easier or harder?

thomasleese: Ah that makes sense, I do prefer it explicitly as `(x, _)`, but I think I just don't like it because in my mind `(x)` and `x` should be exactly the same thing.

kevindew: Yeah I get you there 👍

kevindew: @thomasleese changed that now

PR #638, opened 2017-07-17
Add vanish schema

klssmith:
The vanish schema will only be used for sending unpublishing messages to
the notification queue.

[Trello card](https://trello.com/c/5a7bQyHI/963-notify-downstream-services-search-indexing-when-documents-are-unpublished-3)

### DIFF
--- a/formats/vanish/notification/schema.json
+++ b/formats/vanish/notification/schema.json
@@ -1,0 +1,46 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "base_path",
+    "publishing_app"
+  ],
+  "properties": {
+    "base_path": {
+      "$ref": "#/definitions/absolute_path"
+    },
+    "content_id": {
+      "$ref": "#/definitions/guid"
+    },
+    "document_type": {
+      "enum": [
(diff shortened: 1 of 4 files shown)


fofr:
Presumably this can't re-use existing definitions for `guid`, `absolute_path` etc? Or have I misinterpreted?

thomasleese:
@fofr Because it's a hand written schema we can't make use of the common definitions.

formats/vanish/notification/schema.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "publishing_app"
```
steventux: Is this right that only the publishing app is required? Is this to allow us to vanish pathless content? If so what's the identifier for the thing we want to vanish? Would specifying a `content_id` OR `base_path` be possible?

thomasleese: Hmm, actually this is based on the schema for `gone` which only listed `publishing_app` as a requirement. Thinking about it though, I think it does make sense to include `base_path` in that list.

thomasleese:
@steventux Would you be able to have another look at this now that it's passing?

PR #639, opened 2017-07-17
Add special_route example

gpeng:
This commit adds an example of the `special_route` format. This format isn't rendered but rendering apps do need to handle it now that content-store will fall through to the first prefix route that it finds
if a specific content item is not found.

e.g. `/government/i-dont-exist` will return the `/government` `special_route` item.

Having this example will allow us to test for this and handle it properly.


fofr:
Is this example an exact replica of https://www.gov.uk/api/content/government?

gpeng:
Yes

PR #640, opened 2017-07-18
Remove designprinciples from permitted apps

danielroseman:
service-manual-frontend is the "publisher" as well as the renderer for
the static content previously owned by designprinciples, so add that
app to publishing_apps as well.


tijmenb:
Superseded by https://github.com/alphagov/govuk-content-schemas/pull/664.

PR #643, opened 2017-08-10
Change display title of medical_specialism finder in examples

h-lame:
For: https://trello.com/c/HbnBgaJd/208-rename-filter-title-in-mhra-finder

The real finder has the title of this field changed to "Medical
speciality" to match the words used by the audience of these alerts
so we make the same change in the schemas for consistency.

This matches up with the changes in https://github.com/alphagov/specialist-publisher/pull/1051


h-lame:
Aside: this change didn't actually cause anything in `dist/` to change when I ran `rake`.  Is that to be expected?

h-lame:
No changes to `dist/` because the changes I made were only in the examples.

PR #646, opened 2017-08-18
Allow nil value for publishing publishing_request_id

tijmenb:
The content-store currently has a lot of nil-values for `publishing_request_id`. This is fine since that means the document wasn't published after the request id was introduced.

Context: I found this problem while investigating the content-store returning invalid content (https://trello.com/c/FesP8XSc). To get a better idea about how many items are invalid in the content-store,  https://github.com/alphagov/content-store/pull/306 was raised to allow us to build a [dashboard tracking invalid data][dash].

screen shot 2017-08-18 at 08 21 08

It turns out most invalid data is related to the issue fixed in this PR, so I'm expecting the percentage of invalid content to go down a lot after this.
 
https://deploy.integration.publishing.service.gov.uk/job/check-content-store/23/consoleFull

[dash]: https://grafana.integration.publishing.service.gov.uk/dashboard/db/content-validation-content-store

https://trello.com/c/FesP8XSc
https://trello.com/c/53cyvyda

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -1390,7 +1390,14 @@
     },
     "publishing_request_id": {
       "description": "A unique identifier used to track publishing requests to rendered content",
-      "type": "string"
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ]
     },
     "redirect_route": {
       "type": "object",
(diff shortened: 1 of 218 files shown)


tijmenb:
Looks like this has fixed more than half the validation errors:

screen shot 2017-08-18 at 09 07 45

https://deploy.integration.publishing.service.gov.uk/job/check-content-store/26/console

thomasleese:
🎉

PR #648, opened 2017-09-01
Add whitehall admin as a rendering app

cbaines:
whitehall-admin, rather than whitehall-frontend is responsible for
responding to requests for /government/uploads. Currently,
govuk-puppet sets up the NGinx configuration such that when the
request ends up at the frontend load balancer, it reverse proxies the
request to whitehall-admin.

Adding whitehall-admin as a valid rendering app will allow creating a
special route for directly routing traffic to whitehall-admin, which
will help with simplifying the routing, and making it more explicit.

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -1456,7 +1456,8 @@
             "spotlight",
             "static",
             "tariff",
-            "whitehall-frontend"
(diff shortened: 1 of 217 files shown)


danielroseman:
It's probably not worth spending any time on simplifying asset routing; GoFreeRange are currently working on moving it all to S3, so all that complexity will be going away fairly soon.

PR #652, opened 2017-10-06
Add document type 'task_list' into allowed document types

Davidslv:
This commit adds the placeholder_task_list, that should allow us to run the BETA
version of the task list pages that we are preparing.

Trello:
https://trello.com/c/WOhDbn9e/208-publish-route-for-the-task-list-page


sihugh:
@Davidslv if you can rebase this, we should get this merged so that it won't cause problems when we actually want to do stuff.

PR #658, opened 2017-10-30
Include feedback as dependent application for schema changes

RoryMacDonald:
Schema changes are currently not running specs against the feedback application, which resulted in a failed test when first running the feedback application.


Jenkinsfile
```diff
@@ -29,6 +29,7 @@ def dependentApplications = [
   ['static', true],
 
   ['contacts', false],
+  ['feedback', false],
```
tijmenb: This can be `true` after https://github.com/alphagov/feedback/pull/220.

PR #659, opened 2017-10-31
Add search user need document supertype

brenetic:
Add `search_user_need_document_supertype` to group documents by core or government. In the short term this will be used in an A/B test to boost core (sometimes referred to as 'mainstream') documents in search results.

The Frontend team have been working on [grouping documents](https://gov-uk.atlassian.net/wiki/spaces/GFED/pages/187564388/Grouping+document+types+by+need) by user need however the groupings do not fit our current requirements because the document types that will be targeted are spread across the proposed Guidance and Service super groups.

https://trello.com/c/z9wtoGTV

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -170,6 +170,10 @@
         "answer"
       ]
     },
+    "search_user_need_document_supertype": {
+      "description": "Document supertype grouping core and government documents",
+      "type": "string"
+    },
     "title": {
       "$ref": "#/definitions/title"
     },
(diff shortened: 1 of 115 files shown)


lib/schema_generator/notification_schema_generator.rb
```diff
@@ -48,6 +48,7 @@ def required
         title
         update_type
         user_journey_document_supertype
+        user_need_document_supertype
```
suzannehamilton: Is it OK to make this a required field even though we haven't added it to the publishing API yet? I'd have thought we'd need to make it optional first.

PR #661, opened 2017-11-09
Move applications out of the "doesn't use buildProject" block

h-lame:
We're slowly changing some of the apps in the "doesn't use buildProject" block of our Jenkinsfile so that they do use buildProject.  In most cases they don't use it because the job and repo names don't match.  This has been fixed in buildProject so we can upgrade those apps.  This PR is for tracking those changes as they are merged.  Each commit will be a different app.


h-lame:
Reminder to self: downstream projects that have been fixed need to be deployed to production before we'll start seeing them self report on github as it's the `deployed-to-production` branch that is built by this job.

h-lame:
Oops, I should have marked this as [WIP] because I hope to do a few more before merging (licence-finder and smart-answers).  Thanks for the pre-emptive approve though 😉 !

tijmenb:
Merging because it's valuable to have these switched, even if it's not all of the apps.

PR #665, opened 2017-11-22
Integrate dummy content store

tijmenb:
This integrates the dummy content store (https://github.com/alphagov/govuk-dummy_content_store).

Differences between this and the gem:

- It's no longer a gem so we don't have to release a new version for each change and bump the version. This makes it easier for the app on Heroku to keep up to date (the app is deployed from this repo to make sure that the schemas used by the app are always up to date).
- By using Sinatra we can drop a lot of the code needed to render responses and render the ERB template. This version is hundreds of lines smaller.
- Instead of proxying the content store, this app redirects to the live content store. This removes complexity and is more robust.
- Makes sure that items returned have the correct base_path. Applications like government-frontend sometimes look at the base_path to determine if they should redirect to a more canonical URL.

Tested by running `./startup.sh` in this repo and then this for government-frontend:

```
PLEK_SERVICE_CONTENT_STORE_URI=http://localhost:4567/api ./startup.sh --live
```

After this is merged govuk-dummy_content_store can be archived.

https://trello.com/c/IoHKyhoX

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,9 +4,13 @@ ruby File.read(".ruby-version").strip
 gem 'json-schema'
 gem 'rake'
 gem 'rspec'
-gem 'govuk-dummy_content_store', '0.0.5'
-gem 'foreman', '0.78.0'
 gem 'jsonnet', '~> 0.2'
+
+# Preview app for examples
+gem 'foreman', '~> 0.84.0'
+gem 'sinatra', '~> 2.0'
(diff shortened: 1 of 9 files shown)


nickcolley:
Is there anything we can add to the README to explain how to use this?

tijmenb:
@nickcolley turns out there was documentation already, I've updated it. I'm planning to add docs to government-frontend and other apps on how to use it.

PR #667, opened 2017-11-28
Add service_sign_in schema

emmabeynon:
For [Trello](https://trello.com/c/VO0HVhk8/240-create-schema-for-interstitial-pages-in-govuk-content-schemas)
Paired with @leenagupte 

We are creating a new journey for users on transaction pages in order to better support multiple login options for a service.  After clicking on the "Start now" button on a transaction page, we intend for users to land on a page that presents the multiple login options (e.g. Government Gateway, GOV.UK Verify) in a radio button format so they can make the right choice for them.  One of the options will be "Create new account", and we felt that since this page is so closely linked to the radio button page and would be very unlikely to exist on it own, we can house these pages under one format - `service_sign_in`.  As such, we create a new schema for this format.

### DIFF
--- a/dist/formats/generic/frontend/schema.json
+++ b/dist/formats/generic/frontend/schema.json
@@ -157,6 +157,7 @@
         "service_manual_service_standard",
         "service_manual_service_toolkit",
         "service_manual_topic",
+        "service_sign_in",
         "service_standard_report",
         "services_and_information",
         "simple_smart_answer",
--- a/dist/formats/generic/notification/schema.json
+++ b/dist/formats/generic/notification/schema.json
@@ -172,6 +172,7 @@
         "service_manual_service_standard",
         "service_manual_service_toolkit",
         "service_manual_topic",
+        "service_sign_in",
         "service_standard_report",
         "services_and_information",
         "simple_smart_answer",
--- a/dist/formats/generic/publisher_v2/schema.json
+++ b/dist/formats/generic/publisher_v2/schema.json
@@ -160,6 +160,7 @@
         "service_manual_service_standard",
         "service_manual_service_toolkit",
         "service_manual_topic",
+        "service_sign_in",
         "service_standard_report",
         "services_and_information",
         "simple_smart_answer",
--- a/dist/formats/generic_with_external_related_links/frontend/schema.json
+++ b/dist/formats/generic_with_external_related_links/frontend/schema.json
@@ -157,6 +157,7 @@
         "service_manual_service_standard",
         "service_manual_service_toolkit",
         "service_manual_topic",
+        "service_sign_in",
         "service_standard_report",
         "services_and_information",
         "simple_smart_answer",
--- a/dist/formats/generic_with_external_related_links/notification/schema.json
+++ b/dist/formats/generic_with_external_related_links/notification/schema.json
@@ -172,6 +172,7 @@
         "service_manual_service_standard",
         "service_manual_service_toolkit",
         "service_manual_topic",
+        "service_sign_in",
         "service_standard_report",
         "services_and_information",
         "simple_smart_answer",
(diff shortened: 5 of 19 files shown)


kevindew:
Just done some pairing with Leena and we think you don't need the pages array as a single page can only have both of these. Whereas pages was used in the context of https://www.gov.uk/api/content/log-in-file-self-assessment-tax-return you can have infinite pages.

It also appears that this schema is not valid as you are not defining the object that is being validated here: https://github.com/alphagov/govuk-content-schemas/pull/667/files#diff-4b7b551fe15ce1bfddd6bc85faad4514R18 - I'm probably not explaining that well, but it'd need to be a long the lines of `{ type: "object" ` etc. 

It does seem a problem though that nothing has caught that this isn't valid before it went into this repo. I'll investigate that tomorrow. You can check validity here: https://json-schema-validator.herokuapp.com/syntax.jsp

We did some work restructuring it into a simpler object based structure in this gist: https://gist.github.com/kevindew/27b5dda8ec80a3bf835b8b741cc559b7

emmabeynon:
Thanks @kevindew - we've made some changes, and they should be validated properly now!

examples/service_signin/frontend/service_signin.json
```diff
@@ -0,0 +1,44 @@
...
+  "content_id": "2b4de3cf-cb13-4e9f-ae37-07090ab1a6e6",
+  "document_type": "service_signin",
+  "first_published_at": "2016-02-29T09:24:10.000+00:00",
+  "locale": "en",
+  "phase": "live",
+  "public_updated_at": "2015-05-28T15:46:51.000+00:00",
+  "publishing_app": "publisher",
+  "rendering_app": "government-frontend",
+  "schema_name": "service_signin",
+  "title": "Self Assessment",
```
nickcolley: Since this is a self assessment example, should the filename reflect that?

formats/service_signin.jsonnet
```diff
@@ -0,0 +1,86 @@
...
+            "text",
+            "url",
+          ],
+          items: {
+            type: "object",
+            properties: {
+              text: {
+                type: "string",
+              },
+              url: {
```
nickcolley: Are we expected this to be a full url or path?

examples/service_signin/frontend/service_signin.json
emmabeynon: I've seen quite a bit variation in naming the example files; some schemas just have the schema_name.json, others have the content name.  Not sure what the best practice is?

formats/service_signin.jsonnet
emmabeynon: For external links we would expect a full url, and for internal links we'd expect base_path + slug.  I'll update the example to reflect the latter 👍

tijmenb:
If you rebase this, the examples should be validated again (https://github.com/alphagov/govuk-content-schemas/pull/668).

emmabeynon:
> If you rebase this, the examples should be validated again (#668).

Thanks Tijmen!

tijmenb:
Bit of pedantry: it seems like a lot of services use ["Sign in" instead of "Signin"](https://www.google.co.uk/search?q=sign+in). Would it be better to name this `service_sign_in`?

emmabeynon:
> Bit of pedantry: it seems like a lot of services use "Sign in" instead of "Signin". Would it be better to name this service_sign_in?

Sounds reasonable 👍

leenagupte:
@tijmenb Pedantry accepted. Also "signin" could be confused with "signing".

A bit more pedantry for you: 

The verb is sign in.
The noun is sign-in.

Source https://english.stackexchange.com/questions/49850/sign-in-signin-or-sign-in/49851

PR #672, opened 2017-12-04
Add search keywords field

brenetic:
Formats such as Finders in Whitehall and Calendars send separate indexable content to Rummager as a top level field. This should be an optional field for any document that wishes to send any keywords for Rummager to pass as indexable content into Elasticsearch.

https://trello.com/c/ut9cWxrT

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -170,6 +170,9 @@
         "answer"
       ]
     },
+    "search_keywords": {
+      "$ref": "#/definitions/search_keywords"
+    },
     "search_user_need_document_supertype": {
       "description": "Document supertype grouping core and government documents",
       "type": "string"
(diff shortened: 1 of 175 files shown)


MatMoore:
@dwhenry what's the advantage of having it in details?

brenetic:
Closing this as we decided there is not enough documents that would make use of it at this time

tijmenb:
Specialist documents have [hidden_indexable_content](https://github.com/alphagov/rummager/search?utf8=✓&q=hidden_indexable_content), would that perform the same function as this?

brenetic:
The field itself, potentially. But the issue was more to do with sending the field from wherever each document is published. For example Whitehall just sends it as a top level field which the Publishing API ignores. In the end we just decided that for 3 documents it is not worth adding another top level field.

PR #673, opened 2017-12-05
Smart answer nodes

dwhenry:
https://trello.com/c/tSPsLjz3/489-make-smartanswers-indexable

### DIFF
--- a/dist/formats/transaction/frontend/schema.json
+++ b/dist/formats/transaction/frontend/schema.json
@@ -274,6 +274,9 @@
         "external_related_links": {
           "$ref": "#/definitions/external_related_links"
         },
+        "hidden_search_terms": {
+          "$ref": "#/definitions/hidden_search_terms"
+        },
         "introductory_paragraph": {
           "$ref": "#/definitions/body_html_and_govspeak"
         },
@@ -397,6 +400,12 @@
     "guid": {
       "type": "string",
       "pattern": "^[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$"
+    },
+    "hidden_search_terms": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
     },
     "locale": {
       "type": "string",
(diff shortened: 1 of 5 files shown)


tijmenb:
Hmmm, I wonder if this is right. It could be confusing for all transactions to have these nodes, while they're smart answer specific. How will rummager know when index the nodes? Would it make more sense to have this content sent in the `search_keywords` https://github.com/alphagov/govuk-content-schemas/pull/672?

dwhenry:
@tijmenb I think nodes is the right location for this data, the issue is more around the smart answer start page re-using the transaction format. I went with adding it here to avoid adding a new format at this stage.

We eventually decided #672 was a non-starter so that is not an option at this stage. We could dump this in hidden-indexable-content, but that feels more hacky than this.

If you have a preference come grab me when you have a moment so we can discuss.

tijmenb:
I guess the main thing is that the `flows` data seems unusable outside of search, because it doesn't contain the smart answer logic ([there's an API for that](https://www.gov.uk/state-pension-age/y/age.json)).

Secondly something feels wrong about adding context-specific data to a schema because search happens to use it. The way I see it, the smart answers application [publishes a start page](https://github.com/alphagov/smart-answers/blob/5a2c32d8927c7e86173bc659df2db077cdac1608/app/services/content_item_publisher.rb#L4-L6) for its "service". But there's also a content item for the services itself, the ["flow" content item](https://github.com/alphagov/smart-answers/blob/5a2c32d8927c7e86173bc659df2db077cdac1608/app/services/content_item_publisher.rb#L8-L10) (as yet pretty much empty).

Start page item: https://www.gov.uk/api/content/state-pension-age
Flow page item: https://www.gov.uk/api/content/state-pension-age/y

That would imply that the smartanswer-specific fields should actually go into the flow content item? 

Because nothing will consume the structured `nodes` except for rummager, it seems that `hidden_indexable_content` is actually the best solution? I think it's a totally legit use case for adding content to search that isn't in any of the fields.

dwhenry:
@tijmenb moved the conversation to https://github.com/alphagov/smart-answers/pull/3280 so that we can cover all options before making a decision

PR #682, opened 2017-12-08
Add some more document_type attributes to frontend examples

steventux:
https://trello.com/c/q0PNOZkj/117-move-formats-with-images-to-universal-layout

Moving 'Part of' links to the navigation sidebar means [some
integration tests on government-frontend will require the sidebar
to render](https://github.com/alphagov/government-frontend/pull/580/commits/bff4b8c805ae3f5eb376bd70cd7d4e9b45eff9b5), so add the `document_type` to the relevant
schema examples used in the first batch of redesigned formats as
the [related navigation relies on this attribute to include the links](https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/content_item.rb#L161).

More to follow as we redesign more formats.

### DIFF
--- a/examples/speech/frontend/speech-speaking-notes.json
+++ b/examples/speech/frontend/speech-speaking-notes.json
@@ -52,7 +52,8 @@
         "api_url": "https://www.gov.uk/api/content/government/organisations/home-office",
         "web_url": "https://www.gov.uk/government/organisations/home-office",
         "locale": "en",
-        "analytics_identifier": "D16"
+        "analytics_identifier": "D16",
+        "document_type": "organisation"
       }
     ],
     "speaker": [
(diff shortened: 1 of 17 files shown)


vanitabarrett:
We also check for document_collections with type "document_collection" -  does this need to be added too?

steventux:
@vanitabarrett good point - for consistency yes it should. Will amend.

examples/speech/frontend/speech-speaking-notes.json
```diff
@@ -63,7 +64,8 @@
         "base_path": "/government/people/theresa-may",
         "api_url": "https://www.gov.uk/api/content/government/people/theresa-may",
         "web_url": "https://www.gov.uk/government/people/theresa-may",
-        "locale": "en"
+        "locale": "en",
+        "document_type": "organisation"
```
vanitabarrett: I think this should be `person`

steventux: Oops, yes very good catch :+1:

steventux:
thanks @vanitabarrett updated that incorrect doc type

PR #683, opened 2017-12-11
Update README reference to JSON Schema

rubenarakelyan:
We now use Jsonnet rather than JSON Schema

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The aim of it is to support 'contract testing' between the frontend and
 publisher apps by expressing the schema and examples in strict, machine
 processable form.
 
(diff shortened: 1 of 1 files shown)


kevindew:
@rubenarakelyan I don't think this is quite accurate to state here, in this context (discussing the schemas themselves) they are still json files defined in compliance with json-schema. Jsonnet is used to generate the json-schema files.

So I guess more accurate rephrasings could be "This project contains JSON Schema schema files, JSON example files and Jsonnet files used to generate the schemas" but that seems a bit unclear and a little overly wordy.

rubenarakelyan:
Oh right, I misunderstood the relationship between JSON Schema and Jsonnet...will close.

PR #686, opened 2017-12-12
Service Manual Frontend & Whitehall now use the standard buildProject

h-lame:
For: https://trello.com/c/QtpGwa2A/288-improve-feedback-on-github-for-changes-to-govuk-content-schemas

This means both these apps can report to github themselves and we don't have to wait for them
explicitly in our Jenkinsfile anymore.

These are the last two, but I've left the wait version of the `dependentApplications` loop just in case we need to add some apps that don't use the standard jenkinsfile and don't talk to github themselves.  But I could be convinced to delete this if pressed as I don't really think we want to encourage it.

See: alphagov/service-manual-frontend#127, alphagov/whitehall#3640 and previous work in #661

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,8 +35,7 @@ def dependentApplications = [
   ['smartanswers', true],
   ['specialist-publisher', true],
   ['static', true],
-
-  ['whitehall', false],
+  ['whitehall', true],
 ]
 
 node {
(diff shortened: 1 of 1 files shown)


h-lame:
Note: the whitehall PR isn't merged + deployed yet so this will need to be rebuilt when that happens to prove that it works.

h-lame:
The service-manual-frontend and whitehall PRs have now been merged and deployed to production, so we should see them self-reporting success/fail in github now.  That's the true test of success for this PR.

PR #688, opened 2017-12-12
Add reversable level_one_taxons link to the homepage

koetsier:
- Add level_one_taxons links to the homepage
- Deprecate root_taxons link on the homepage
- Add root_taxon to taxons which acts as the reverse of the
level_one_taxons

Adding a parent of a parentless taxon previously required an additional call to the publishing api to ensure it is not a level one taxon (i.e. getting the homepage to see if there is a 'root_taxons' link to the taxon). With this change a reverse link is introduced making these types of update easier.

### DIFF
--- a/formats/taxon.jsonnet
+++ b/formats/taxon.jsonnet
@@ -24,6 +24,13 @@
     associated_taxons: "A list of associated taxons whose children should be included as children of this taxon",
   },
   links: (import "shared/base_links.jsonnet") + {
-    parent_taxons: "The list of taxon parents (DEPRECATED: use the edition links instead)",
+    parent_taxons: {
+      "$ref": "#/definitions/guid_list",
+      description: "The list of taxon parents (DEPRECATED: use the edition links instead)",
+    },
+    root_taxon: {
+      "$ref": "#/definitions/guid",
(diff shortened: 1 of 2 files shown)


formats/taxon.jsonnet
```diff
@@ -24,6 +24,13 @@
     associated_taxons: "A list of associated taxons whose children should be included as children of this taxon",
   },
   links: (import "shared/base_links.jsonnet") + {
-    parent_taxons: "The list of taxon parents (DEPRECATED: use the edition links instead)",
+    parent_taxons: {
+      "$ref": "#/definitions/guid_list",
+      description: "The list of taxon parents (DEPRECATED: use the edition links instead)",
```
cbaines: I went back in time, and found the source of this deprecation message ( c31ba3a8aea1b6f958ffde8552c845634e60f353 ).

I think the reasoning makes sense, and this might be a sensible time to switch Content Tagger to use the edition links, as we are making a change relating to the root_taxon link type?

tijmenb: I think that got abandoned because there was something with the publishing-api not doing reverse links for edition links.

PR #690, opened 2017-12-13
Add external_content schema and document type

suzannehamilton:
Add schema and document type which represent links to content which is relevant to GOV.UK but is not hosted there, such as local councils or selected NHS pages.

External URLs are indexed in site search so that we can direct users to relevant non-GOV.UK content.

These URLs are currently sent to rummager (and the search index) by the search admin tool. This new schema will allow search admin to send pages to the publishing API instead, which now sends all publishing events to rummager. This will help us reduce the number of direct dependencies on rummager.

The new schema implements the definition discussed in [RFC 88](https://github.com/alphagov/govuk-rfcs/pull/88)

https://trello.com/c/4J6h4YzC/523-create-document-type-for-external-links

Marked as "do not merge" until the RFC is merged.


formats/external_content.jsonnet
```diff
@@ -0,0 +1,32 @@
...
+      ],
+      properties: {
+        url: {
+          description: "A URL for the external resource",
+          type: "string",
+          format: "uri",
+        },
+        comment: {
+          description: "An internal description of why the link is relevant to GOV.UK users",
+          type: "string",
```
dwhenry: I think I would like to see comment be a required field so that it can't be skipped.

brenetic:
I have heard `external_related_links` and `recommended_links` as terms used to describe this. Are we further confusing things by adding another term for it or should they remain as separate things?

formats/external_content.jsonnet
suzannehamilton: @MatMoore what's your opinion on this?

Comments are currently optional in search-admin, but it looks like every link has one anyway. (Though they're mostly automated comments which link to https://github.com/gds-attic/recommended-links)

suzannehamilton:
@brenetic There was some [discussion about this on the RFC](https://github.com/alphagov/govuk-rfcs/pull/88#issuecomment-351033067). "link" might be confused with the publishing API's concept of links between documents.

formats/external_content.jsonnet
suzannehamilton: Having discussed this with @MatMoore, we're going to remove this field from the schema entirely. We'll continue to store it in the search-admin DB but it's an internal note so it doesn't currently make sense to send it to the publishing API.

MatMoore:
@brenetic Maybe to avoid confusion we can change the search admin UI to also refer to this as external content?

PR #700, opened 2018-01-11
Make some fields non-mandatory in unpublishing notifications

suzannehamilton:
Update notifications for unpublishing types (gone, redirect and vanish) to make all the fields non-mandatory except those which are required by downstream apps, such as `base_path` and `content_id`.

The publishing API does not add these fields anyway, so downstream apps should not be relying on them. This schema change makes the content schemas consistent with the publishing API's messages, which will let us add tests to help us keep it consistent.

https://trello.com/c/uV5zDvzu/97-add-schema-tests-to-the-publishing-api

This should fix most of the test failures in alphagov/publishing-api#1132

I'd especially like review feedback on these things:
- Are any of these fields useful to keep on unpublishing messages?
- Is there a better place to define unpublishing types than `NotificationSchemaGenerator`?


lib/schema_generator/format.rb
```diff
@@ -88,7 +88,7 @@ def generate_publisher?
       generate_publisher.nil? ? true : generate_publisher
     end
 
-    def generate_nofitication?
+    def generate_notification?
```
kevindew: oops thanks for catching!

kevindew:
> Are any of these fields useful to keep on unpublishing messages?

I don't think they're of any use but as a lot of them can be created directly (by publishing a gone or redirect) it makes sense for them to still be optional.

> Is there a better place to define unpublishing types than NotificationSchemaGenerator?

I think that's the only place where they're applicable at the moment. I hadn't actually realised that Publishing API used a presenter with less fields for the unpublishing types when I wrote it. The content store always all the same attributes no matter what format. 

The situation where it would be good to be more generic is if we generated a schema of what should be going into content store, which is something we should do as that's something that can't even be validated at the moment. That would be almost identical to the notification schema but a few less fields.

suzannehamilton:
> I don't think they're of any use but as a lot of them can be created directly (by publishing a gone or redirect) it makes sense for them to still be optional.

They are still optional (which you might have spotted), but I've worded it badly so it sounds like they're now forbidden. I'll update the commit message to make it clear that I've just made them non-mandatory. 👍

PR #705, opened 2018-01-17
Add residential property tribunal decision specialist document

h-lame:
For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

Residential property tribunal decisions have a decision date, a category, and a sub-category.  Only one of each.


h-lame:
Removed WIP as we got confirmation (and also that we need to add sub category).

elenatanasoiu:
After testing this branch against `government-frontend`, I've found that a test breaks when trying to fetch the new `residential_property_tribunal_decision` because it expects an example in:

```
govuk-content-schemas/examples/specialist_document/frontend/residential_property_tribunal_decision.json
```

In addition, I've checked and there are a few other examples that are missing from here (and don't have tests in `government-frontend`). In case you're interested in completing the list, they are: 

```
"dfid_research_output"
"esi fund" (although there is a "employment-structural-investment-funds" example)
"medical safety alert"
```

There's also an extra example in that folder of examples that's not a specialist_document: "drug-device-alerts"

h-lame:
@elenatanasoiu - good spot, thanks.  I've added an example.  However we'll need to revisit it once we have some example documents because it's basically full of lorem ipsum at the moment, unlike the other examples which are real documents.

formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1138,6 +1141,83 @@
...
+          "rents---market-rent-assured-shorthold-tenancy",
+          "rents---fair-rent",
+          "right-to-buy---rtb-appeals-elderly-persons",
+          "tenant-associations---recognition-of-tenants-association",
+          "tenant-associations---request-for-information"
+        ],
+      },
+      tribunal_decision_decision_date: {
+        type: "string",
+        pattern: "^[1-9][0-9]{3}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[0-1])$",
```
suzannehamilton: This date pattern is hardcoded a few times in this file. Does the jsonnet format let you extract a shared definition?

h-lame: My jsonnet / json-schema knowledge isn't super great, but this feels possible.  I've put this down on my "clean this up afterwards" list for a separate PR.  Thanks!

PR #713, opened 2018-01-30
Rename subcategories for residential tribunal decisions

elenatanasoiu:
For: https://trello.com/c/BvCDLMMQ/27-moj-residential-tribunals-finder

PR #715, opened 2018-01-31
Renames subcategory values for residential decision

elenatanasoiu:
For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

Amends work we forgot to add in https://github.com/alphagov/govuk-content-schemas/pull/713


formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1204,7 +1204,7 @@
           "park-homes---entitlement-to-terminate-an-agreement-for-occupation",
           "rents---market-rent-assured-shorthold-tenancy",
           "rents---fair-rent",
-          "right-to-buy---rtb-appeals-elderly-persons",
+          "right-to-buy---right-to-buy---right-to-buy-appeals-elderly-persons",
```
h-lame: this should be: `right-to-buy---right-to-buy-appeals-elderly-persons` (according to: https://github.com/alphagov/specialist-publisher/blob/master/lib/documents/schemas/residential_property_tribunal_decisions.json#L236)

elenatanasoiu: darn, thanks!

PR #718, opened 2018-02-02
Add  logic to step nav

sihugh:
This enables us to show `and` / `or` as part of the step by step navigation
journey.

### DIFF
--- a/dist/formats/step_by_step_nav/frontend/schema.json
+++ b/dist/formats/step_by_step_nav/frontend/schema.json
@@ -326,6 +326,13 @@
             ]
           }
         },
+        "logic": {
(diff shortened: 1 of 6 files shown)


Davidslv:
@sihugh has this decision been made already? 
What you think of calling it `conjunction` instead of `logic` as both `and` & `or` are conjunctions, grammatically speaking.

sihugh:
It wouldn't make much sense to have other conjunctions in there though (`but`, `yet` etc.).  It's to indicate whether you have to do a step or not.   i can imagine that our use might change given more step navigation samples and further testing, but I'm OK with `logic` at the moment given the lack of great alternatives.

PR #737, opened 2018-03-16
Add legacy_taxon <-> topic_taxonomy_taxons links to Taxon type

tomsabin:
A new `legacy_taxon` link type exists to represent where previous taxons
used to exist, before the single site-wide Topic Taxonomy was created.
We also require a reverse link, `topic_taxonomy_taxons`, to direct these
'legacy' taxons to the newer taxon in the Topic Taxonomy.

Trello: https://trello.com/c/2n0C0KdR/151-map-legacy-taxons-to-merged-taxons

### DIFF
--- a/lib/schema_generator/expanded_links.rb
+++ b/lib/schema_generator/expanded_links.rb
@@ -29,6 +29,11 @@ module SchemaGenerator
       # Content items that are linked to with a `pages_part_of_step_nav` link type
       # will automatically have a `part_of_step_navs` link type with those items
       "part_of_step_navs",
+
+      # Taxons that have been created by merging old 'legacy' taxons will have
+      # a reverse link to determine where the replacement Topic Taxonomy taxon
+      # now resides
+      "topic_taxonomy_taxons",
     ].freeze
 
     def initialize(format)
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -137,6 +137,10 @@
           "description": "Prototype-stage taxonomy label for this content item",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
+        "topic_taxonomy_taxons": {
+          "description": "Link type automatically added by Publishing API",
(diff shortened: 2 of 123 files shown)


lib/schema_generator/expanded_links.rb
```diff
@@ -29,6 +29,11 @@ class ExpandedLinks
       # Content items that are linked to with a `pages_part_of_step_nav` link type
       # will automatically have a `part_of_step_navs` link type with those items
       "part_of_step_navs",
+
+      # Taxons that have been created by the merging old 'legacy' taxons will
+      # have a reverse link to determine where the site-wide Topic Taxonomy
+      # taxon now reside.
+      "topic_taxonomy_taxons",
```
cbaines: While this is correct, I think the function of this array is to add these links to every schema. That shouldn't be necessary in this case, as this link type should only be used in the case of policies, policy areas, topics... We could add it specifically to those, but I don't mind adding it here either.

tomsabin: We attempted to localise the document types that this link type would show up in (mainstream_browse_page, topic, policy_area, policy), and while this produced the expected schemas (except policy_area), the Publishing API [CI build](https://ci.integration.publishing.service.gov.uk/job/publishing-api/job/legacy-taxons-reverse-link/1/display/redirect) is expecting that all document types in the `/dist/formats` directory to have the reverse link defined.

I think this might be quite a big task to change, and there are also numerous cases in this list which is would also benefit from the same refactor. All reverse links that exist in the [expansion_rules.rb](https://github.com/alphagov/publishing-api/blob/fc18111cd81f93c26349d8f5a6fec5a05dcbec4b/lib/expansion_rules.rb#L28-L35) file in Publisher are also declared in this `LINK_TYPES_ADDED_BY_PUBLISHING_API` array.

However, I've added the relevant document types that this relates to in the legacy_taxon link type in formats/taxon.jsonnet which should help a little. I don't think this refactor should hold out this link addition, as it'll block us from moving forward on adding the legacy taxon mapping into the Publishing API.

PR #744, opened 2018-03-26
Skip checking dependent branches when building master

cbaines:


cbaines:
> Does this leave open a window between the time that the PR is raised and passes CI, and new commits being merged into the dependent applications?

With this change, there wouldn't be any checks. Without it, things would fail in CI.

It's been long enough since I raised this now, that I forget if I was thinking that the jobs were redundant, or that they are rarely looked at (as they don't appear on GitHub), or both... and I didn't put my full reasoning in the commit message...

tijmenb:
@cbaines should we close this?

cbaines:
> @cbaines should we close this?

I don't know of anyone using these jobs, so it saves Jenkins a bit of work if they are generally unused. But I'm not too fussed. I haven't noticed this since, probably because the `govuk-content-schemas` are rarely updated.

PR #748, opened 2018-04-19
Add new organisation schema

rubenarakelyan:
This commit adds a new organisastion schema that will be used when populating organisations in the content-store. This will allow everything that is currently represented on these pages to be stored in the content-store. It also deprecates a link type in the placeholder schema that will be replaced by
an equivalent in the organisation schema.

Trello: https://trello.com/c/xebVLRcp/24-work-out-what-information-we-need-to-publish-to-the-content-store-for-organisations-and-add-a-schema-for-it


formats/placeholder.jsonnet
```diff
@@ -53,13 +70,423 @@
...
+              },
+              image_href: {
+                type: "string",
+                format: "uri",
+              },
+            },
+          },
+          description: "Used for organisations, to allow us to publish board member information. Remove when organisations are migrated.",
+        },
+        ordered_military_personnel: {
```
Davidslv: Do we have intentions of extracting this into another schema in the future?

rubenarakelyan: I think once we migrate people and roles then we'll make these links in the links hash so they get updated automatically using dependency resolution.

formats/organisation.jsonnet
```diff
@@ -0,0 +1,571 @@
...
+                null,
+              ],
+            },
+            image: {
+              "$ref": "#/definitions/image",
+            },
+          },
+          description: "The organisation's logo, including the logo image and formatted name.",
+        },
+        ordered_corporate_information_pages: {
```
tijmenb: I think `ordered_corporate_information_pages` and `ordered_featured_links` are all migrated, could they go into the edition links?

formats/organisation.jsonnet
```diff
@@ -0,0 +1,571 @@
...
+                "$ref": "#/definitions/image",
+              },
+              unpaid: {
+                type: "boolean",
+              },
+            },
+          },
+          description: "A list of the organisation's special representatives, if appropriate. Turn into proper links once organisations are fully migrated.",
+        },
+        ordered_contacts: {
```
tijmenb: I think these also might be migrated, not sure though.

formats/organisation.jsonnet
```diff
@@ -0,0 +1,571 @@
...
+                "$ref": "#/definitions/image",
+              },
+            },
+          },
+          description: "A set of links to social media profiles for the organisation.",
+        },
+        external_related_links: {
+          "$ref": "#/definitions/external_related_links",
+        },
+        tags: {
```
tijmenb: What are `tags` for?

formats/organisation.jsonnet
rubenarakelyan: They are but they can also be external links so we need the flexibility...

formats/organisation.jsonnet
rubenarakelyan: Hangover from copying the placeholder schema - removed

formats/organisation.jsonnet
tijmenb: Ah that's unfortunate. We should make a note to represent the external links too ([see Mat's RFC](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-088-external-content.md)) 

Does this apply to `ordered_corporate_information_pages` as well?

formats/organisation.jsonnet
rubenarakelyan: Looks like they are published to the publishing-api, so moving them to links

formats/organisation.jsonnet
rubenarakelyan: Yeah both `ordered_corporate_information_pages` and `ordered_featured_links` can contain external links - for example to the Civil Service Jobs site or organisation-specific services on other sites.

PR #753, opened 2018-04-26
Make certain organisation fields nullable

rubenarakelyan:
This commit allows `name_prefix` and `role_href` for people and `public_updated_at` and `document_type` for featured documents be `null`, since these fields are not always expected to be populated. It also moves the person definition into a shared file since all person definitions for organisations are identical.

Trello: https://trello.com/c/IftdS5pN/25-get-whitehall-to-publish-organisation-data-to-the-content-store


sihugh:
Green!

PR #755, opened 2018-04-30
Update Person Schema

Davidslv:
We are moving Organisations away from Whitehall, in this commit we would like to expand the Person schema so we can start sending information to Publishing API.

There might be more fields to be added in the future, such as person role.

Trello: https://trello.com/c/32LO2xT8/30-work-out-what-information-we-need-to-publish-to-the-content-store-for-people-and-fix-the-schema-for-it

### DIFF
--- a/formats/person.jsonnet
+++ b/formats/person.jsonnet
@@ -4,6 +4,19 @@
       type: "object",
       additionalProperties: false,
       properties: {
+        analytics_identifier: {
+          "$ref": "#/definitions/analytics_identifier",
+        },
+        full_name: {
+          type: "string",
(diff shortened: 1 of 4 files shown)


formats/person.jsonnet
```diff
@@ -4,6 +4,19 @@
...
       properties: {
+        analytics_identifier: {
+          "$ref": "#/definitions/analytics_identifier",
+        },
+        full_name: {
+          type: "string",
+          description: "Name of the person, including titles and any letters, eg: 'Sir Lord Snoopy DOG'",
+        },
+        minister: {
+          type: "boolean"
```
rubenarakelyan: Probably best to add a `,` at the end for consistency

Davidslv: I've addressed this, thank you

PR #756, opened 2018-05-02
Add organisation links by type

rubenarakelyan:
This commit adds new link types for organisations grouped by their type. This enables index pages and similar content that requires all organisations of a certain type.

Trello: https://trello.com/c/kMgJ1gqF/56-link-to-all-organisations-from-the-organisation-list-content-item

### DIFF
--- a/formats/organisations_homepage.jsonnet
+++ b/formats/organisations_homepage.jsonnet
@@ -1,0 +1,18 @@
+(import "shared/default_format.jsonnet") + {
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {},
+    },
+  },
+  links: (import "shared/base_links.jsonnet") + {
+    ordered_executive_offices: "All organisations of the type 'executive office', ordered alphabetically ignoring any prefixes.",
+    ordered_ministerial_departments: "All organisations of the type 'ministerial department', ordered alphabetically ignoring any prefixes.",
+    ordered_non_ministerial_departments: "All organisations of the type 'non-ministerial department', ordered alphabetically ignoring any prefixes.",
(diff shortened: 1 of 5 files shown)


rubenarakelyan:
`organisation` schema has a load of required fields that are not needed for the index page, and I couldn't find anything more appropriate. I also felt that creating a new schema for one document was a bit of an overkill!

tijmenb:
Ah, that makes sense! 

A schema for a single page is perfectly acceptable though - [homepage](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/homepage.jsonnet) and [service_manual_homepage](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/service_manual_homepage.jsonnet) do that too. The main benefit is that you'll be able to make the link types required in the edition links and avoid confusion, someone looking at the services and information pages could ask themselves why it's allowed to have a `ordered_agencies_and_other_public_bodies` link type.

rubenarakelyan:
Ah OK, I have created a new schema with the links

formats/organisations_homepage.jsonnet
```diff
@@ -0,0 +1,18 @@
+(import "shared/default_format.jsonnet") + {
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {},
+    },
+  },
+  links: (import "shared/base_links.jsonnet") + {
```
tijmenb: Not a blocker, but you could make these edition links so that they go into the publishing-api in a single call.

rubenarakelyan: I decided to not make them edition links since there will be a process that will update the links when new organisations are created etc which can just ran `patch_links` without having to republish the entire content item since that will very rarely change.

PR #758, opened 2018-05-08
Roles schema

Davidslv:
Currently there's no Role schema, this is the first step to add roles
into publishing api.

Trello: https://trello.com/c/oQ0w74iH

### DIFF
--- a/dist/formats/role/frontend/schema.json
+++ b/dist/formats/role/frontend/schema.json
@@ -1,0 +1,760 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "required": [
+    "base_path",
+    "content_id",
+    "description",
+    "details",
+    "document_type",
+    "links",
+    "locale",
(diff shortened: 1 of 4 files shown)


formats/role.jsonnet
```diff
@@ -0,0 +1,85 @@
...
+          type: "boolean",
+        },
+        cabinet_member: {
+          type: "boolean",
+        },
+        slug: {
+          type: "string",
+          description: "eg: 'prime-minister'.",
+        },
+        chief_of_the_defence_staff: {
```
sihugh: I've got a feeling that this isn't actually used anywhere.  The current CDS is not flagged as `true` for this and neither is anyone else.  I also don't think it's exposed in the UI anywhere.

sihugh:
@Davidslv if you've got an example you can append to this, that'd be 👌

Davidslv:
@sihugh I'm going to do that as part of the following card: https://trello.com/c/EB6Dofoj

PR #759, opened 2018-05-22
Add biography to person schema

rubenarakelyan:
This commit adds a biography field to the person schema.

### DIFF
--- a/dist/formats/person/frontend/schema.json
+++ b/dist/formats/person/frontend/schema.json
@@ -410,6 +410,14 @@
         }
       ]
     },
+    "body_html_and_govspeak": {
+      "description": "The main content provided as HTML with the govspeak markdown it's rendered from",
+      "anyOf": [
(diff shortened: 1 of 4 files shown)


rubenarakelyan:
True dat

dist/formats/person/frontend/schema.json
```diff
@@ -448,6 +448,10 @@
         "analytics_identifier": {
           "$ref": "#/definitions/analytics_identifier"
         },
+        "body": {
+          "description": "Biography of the person",
+          "type": "string"
```
tijmenb: Does this need to be `body_html_and_govspeak`?

https://github.com/alphagov/govuk-content-schemas/blob/master/formats/help_page.jsonnet#L9?

rubenarakelyan: Yes, that works :)

PR #761, opened 2018-05-23
Add links between people, roles and organisations

rubenarakelyan:
This commit adds links from people -> roles, roles -> people, roles -> organisations and organisations -> roles.

Trello: https://trello.com/c/UDOsklt8/71-link-people-to-roles-in-publishing-api and https://trello.com/c/rhJHzjPP/72-link-roles-to-organisations

### DIFF
--- a/dist/formats/organisation/frontend/schema.json
+++ b/dist/formats/organisation/frontend/schema.json
@@ -253,7 +253,7 @@
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
         "ordered_contacts": {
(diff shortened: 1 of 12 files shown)


tijmenb:
Have we considered starting to model appointments in the publishing-api. Each `appointment` would link to a `role` and `person`? 

By setting up reverse linking the person and role content items will be updated every time a person changes role, without having to touch the actual content items for person and role. 

This would also enable you to add metadata to the appointment, so we can say something like "Theresa May has been PM since 2015"

PR #762, opened 2018-05-23
Add role appointment schema

rubenarakelyan:
This commit adds role appointments as a new schema. This schema is basepathless, and will model the relationship between people and roles, including both current and previous roles.

It also removes the `role_type` field from the role schema, which is unused and can be determined by the `document_type`.

Trello: https://trello.com/c/UDOsklt8/71-link-people-to-roles-in-publishing-api


formats/role_appointment.jsonnet
```diff
@@ -0,0 +1,23 @@
+(import "shared/default_format.jsonnet") + {
+  base_path: "forbidden",
+  routes: "forbidden",
+  definitions: {
```
tijmenb: Does this need a `role` and `person` edition link type?

rubenarakelyan: Indeed, added

PR #764, opened 2018-06-01
Fix manuals schema import of links

chao-xian:
Manuals was importing `base_edition_links` instead of `base_links`, which isn't right. This was discovered when trying to set a primary_publishing_organisation for manuals.


chao-xian:
[(2) Set primary org in Manuals Publisher](https://trello.com/c/XZdljjFi/377-2-set-primary-org-in-manuals-publisher)

chao-xian:
Thanks @tijmenb for helping me track down the underlying issue here. Addressed in https://github.com/alphagov/govuk-content-schemas/pull/764/commits/9e7860f355d6899f304ff11170d500746f14eecf

PR #770, opened 2018-06-12
Expand Whitehall summary_organisations

Davidslv:
We need more information in `summary_organisations`.
The reason behind this is that we need to maintain the `/api/organisations` current endpoint, this endpoint is managed by Whitehall::Api and since organisations are being extracted from Whitehall we are moving this functionality to be rendered by collections app.

The main objective is to keep the endpoint exactly as it is but being fetched from content-store.
Without this change, the alternative is to find each content_item (organisation) by base_path and retrieve the necessary information, although this would cause several more requests. We think this is the best approach.

Trello: https://trello.com/c/p7ywlbk1/130-migrate-organisations-api

### DIFF
--- a/dist/formats/organisations_homepage/frontend/schema.json
+++ b/dist/formats/organisations_homepage/frontend/schema.json
@@ -599,7 +599,55 @@
         ],
         "additionalProperties": false,
         "properties": {
+          "abbreviation": {
+            "type": "string"
+          },
+          "analytics_identifier": {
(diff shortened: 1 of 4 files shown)


rubenarakelyan:
I don't think the new fields need to be in `details` since `summary_organisations` isn't meant to be a full content item on its own. So the new fields can be at the same level as the existing ones.

PR #778, opened 2018-06-15
Revert "Add organisations to statutory instruments"

steventux:
Reverts alphagov/govuk-content-schemas#777

These should have been placed in `edition_links`


elliotcm:
@steventux Did you intend to also put them in the right place here?  Or do they already exist in that place in the schema?

PR #781, opened 2018-06-18
Correct links schemas

MatMoore:
The links schemas do not accurately describe the data we have in the publishing API.

Make it so that all existing is content is valid. I'm raising a separate PR on publishing API so that we can validate links when they are set.

There are more issues than I've addressed here but I'd like to deploy as is.
https://github.com/alphagov/publishing-api/pull/1258 will flag any further issues.


MatMoore:
Thanks @thomasleese and @sihugh 👍

PR #790, opened 2018-06-21
Make title and intro mandatory for step by steps

sihugh:
They're always populated by the publisher and are required for rendering in the
view.

This was causing an intermittent failure on Collections tests because one of the examples didn't have the title.


sihugh:
I'd missed a publisher example from the update

PR #796, opened 2018-07-10
Add more countries to IDF documents

rubenarakelyan:
This commit syncs the list of countries for International Development Fund documents to match those defined in specialist-publisher.


elliotcm:
@rubenarakelyan Does this list need reviewing by product/content, as it seems to include removing [Sudan](https://www.gov.uk/world/sudan) and [TOPT](https://www.gov.uk/world/the-occupied-palestinian-territories)

rubenarakelyan:
TOPT was a mistake so I've added it back in. Sudan is in there, it's just changed position.

PR #800, opened 2018-07-25
Add ISBN lines to official paper examples

keithiopia:
These examples of official documents (command papers and house of commons papers) need the ISBN ref visible on the page as well as in the metadata.


steventux:
:+1: Test failures appear to be unrelated poltergeist network issues.

PR #803, opened 2018-08-03
Duplicate news article links as edition links

benthorner:
https://trello.com/c/4sTwhdPN/87-make-associations-available-to-select-when-creating-a-document-l

The new Content Publisher app will take the approach of using edition
links wherever possible, the key advantage being that edition links are
part of the published content and can therefore be scheduled/versioned.

Initially we would like to support a subset of the current links, such
as world_locations and organisations; other kinds of links, such as
roles and people, are technically more challenging and will come later.

Note this should have no effect on current publishing tools, as existing
linkset links will be used if edition links do not exist for a document.

### DIFF
--- a/dist/formats/news_article/frontend/schema.json
+++ b/dist/formats/news_article/frontend/schema.json
@@ -171,6 +171,7 @@
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
         "topical_events": {
+          "description": "The topical events this content item relates to.",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
         "topics": {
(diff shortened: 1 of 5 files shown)


cbaines:
> So for taxons, we'll be doing linkset links and telling the user they'll go live immediately; for organisations, we thought this should be a matter for discussion by the team making the change.

Ok, I don't quite follow about the team reference, are you talking about whoever is making the content change?

cbaines:
> @cbaines I think it'd be quite rare that we can change organisations on content without the content itself needing updating.
> 
> Frequently content references the organisation within the content

Interesting, I have some memories of the Organisations team last quarter discussing changes (tagging, content, I'm not sure) being needed when organisations changed, but I know nothing about the actual process.

PR #806, opened 2018-08-08
Duplicate stat data set links as edition links

benthorner:
The new Content Publisher app will take the approach of using edition
links wherever possible, the key advantage being that edition links are
part of the published content and can therefore be scheduled/versioned.

Note this should have no effect on current publishing tools, as existing
linkset links will be used if edition links do not exist for a document.


benthorner:
Previous PR for news articles, for reference...

https://github.com/alphagov/govuk-content-schemas/pull/803/files

PR #807, opened 2018-08-10
Add more examples to test hide_chapter_navigation

andysellick:
We need some more examples in order to be able to test the new guide property, `hide_chapter_navigation`.

- remove `hide_chapter_navigation` from `guide-with-step-navs` and add a new example for this, `guide-with-step-navs-and-hide-navigation`
- add example of a single page guide with step navs and `hide_chapter_navigation`


andysellick:
@leenagupte thanks - any idea why the tests are failing? Doesn't seem related to this PR, but I could be wrong?

PR #814, opened 2018-09-24
Add roles and people to news edition links

kevindew:
Trello: https://trello.com/c/3GT17tkH/211-associate-content-with-ministers-l

This makes these two link types available as edition links so that they
can be used by Content Publisher.

It also updates the descriptions to not be email specific since these
aren't only used for email.

These links do seem a bit fishy as they all used to link to both aspects
of a role appointment without the link to the role appointment. It would
probably make more sense that these linked to a role appointment and
these were expanded. Though this would be a more substantial change.


kevindew:
Whitehall passes against this branch: https://github.com/alphagov/whitehall/pull/4370

PR #815, opened 2018-09-25
Allow bulk_publishing parameter for Publishing API payloads

kevindew:
Loosely related to: https://trello.com/c/3GT17tkH/211-associate-content-with-ministers-l

Despite this field having been available in Publishing API for 2 years
it's not actually accepted at a schema level. Presumably, in the case of
PutContent, this means it's not been used in the last couple of years as
it would have returned a 422 response.

As it's quite a useful option to have I felt it was better to add it
into the schemas rather than remove the feature from Publishing API.

### DIFF
--- a/lib/schema_generator/publisher_links_schema_generator.rb
+++ b/lib/schema_generator/publisher_links_schema_generator.rb
@@ -24,7 +24,8 @@ module SchemaGenerator
     def properties
       {
         "links" => links,
-        "previous_version" => { "type" => "string" }
+        "previous_version" => { "type" => "string" },
+        "bulk_publishing" => { "type" => "boolean" }
       }
     end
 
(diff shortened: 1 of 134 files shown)


lib/schema_generator/publisher_links_schema_generator.rb
```diff
@@ -24,7 +24,8 @@ def generate
     def properties
       {
         "links" => links,
-        "previous_version" => { "type" => "string" }
+        "previous_version" => { "type" => "string" },
+        "bulk_publishing" => { "type" => ["null", "boolean"] }
```
cbaines: Could this just be boolean? Is there a use case in passing in this value, but it being null?

kevindew: I think you're right, not sure how I came to the null or boolean conclusion.

kevindew:
Whitehall passes against this branch: https://github.com/alphagov/whitehall/pull/4370

PR #818, opened 2018-10-02
Remove base_path requirement for some links (and add org examples)

sihugh:
This adds examples for:

- courts and tribunals pages which are served from under "/courts-tribunals/"
- examples for specific org pages (served from under "/government/organisations/") which enable us to reduce the code in the tests for org pages

Org pages link to several things that don't have base_paths, such as organisation contacts and foi contacts.  The new examples fail the tests unless we can generate linked items without base paths which only currently happens for linked items that are `world_locations`

In theory there is no restriction on which content types are linked, so this check feels a bit like a false safety net because a content item without a base path could be added to any link section.  In practice, this doesn't seem to happen.

I tried removing the requirement for a base path from this section, but a lot of downstream tests failed on apps that use govuk_schemas to generate example content items.  I don't have time to fix them all at the moment so I'll reuse the existing pattern here instead.

https://trello.com/c/IRoHCQ3N/143-render-courts-and-tribunals-pages-from-collections

### DIFF
--- a/lib/schema_generator/format.rb
+++ b/lib/schema_generator/format.rb
@@ -267,6 +267,7 @@ module SchemaGenerator
 
     class Links
       ALLOWED_KEYS = %w(description required minItems maxItems).freeze
+      LINKS_WITHOUT_BASE_PATHS = %w(ordered_contacts ordered_foi_contacts world_locations).freeze
 
       attr_reader :links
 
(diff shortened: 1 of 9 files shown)


sihugh:
@kevindew I chickened out :-(

sihugh:
@rubenarakelyan @kevindew I don't suppose one of you would mind giving this some 👀 ?

lib/schema_generator/format.rb
```diff
@@ -290,11 +291,10 @@ def required_links
 
       def frontend_properties
         links.each_with_object({}) do |(k, v), hash|
-          # It's possible for all link types to contain items without base_paths
-          # however apps aren't coded for this so fail on this, therefore
+          # It's possible for all link types to contain items without base_paths# however apps aren't coded for this so fail on this, therefore
```
rubenarakelyan: Something happened here with two lines of comments...

sihugh: Gah. Thanks!  Got to massage this through a bunch of temperamental e2e tests again now 😬

PR #827, opened 2018-10-25
Allow a high resolution URL to be set for images

tijmenb:
Currently, applications like Whitehall set `details.image.url` to a small sized image, for display on the frontend. For example, the [small lead image for a news story](https://www.gov.uk/government/news/government-announces-additional-funding-towards-mayflower-400-commemorations).

This image is 300px wide on the page, so Whitehall sends a 300px version. Since the introduction of the [machine readable metdata component](https://govuk-publishing-components.herokuapp.com/component-guide/machine_readable_metadata) we also expose the image in the meta tags to services like Facebook, Twitter, Google, and Slack. However, they often need a larger resolution image than what we're currently providing. This causes blurry pictures on Twitter:

screen shot 2018-10-25 at 14 47 41

This PR adds a second attribute that contains a "high resolution" version. The component will use this instead of the smaller version.

https://trello.com/c/eqcMykln

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -532,7 +532,13 @@
             }
           ]
         },
+        "high_resolution_url": {
+          "description": "URL to a high resolution version of the image, for use by third parties such as Twitter, Facebook or Slack. Used by the machine readable metadata component. Don't use this on user-facing web pages, as it might be very large.",
+          "type": "string",
+          "format": "uri"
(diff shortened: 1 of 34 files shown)


kevindew:
Is this going to be a 960 x 640 image? asking in relation to what content-publisher should send

tijmenb:
Yes, that's what I'm trying in https://github.com/alphagov/whitehall/pull/4446.

PR #828, opened 2018-10-25
Remove policy page related bits

cbaines:
Following on from the retirement of the Policy pages on GOV.UK, and Policy Publisher, this cleans up the govuk-content-schemas with respect to these changes.

https://trello.com/c/0jgKr3UT


cbaines:
This should fix the Rummager issues: https://github.com/alphagov/rummager/pull/1302

PR #829, opened 2018-10-31
Make base_path optional for nested links

tijmenb:
As described in https://github.com/alphagov/govuk_publishing_components/pull/595#discussion_r229322211, https://github.com/alphagov/govuk-content-schemas/pull/818 made the `base_path` optional in the `links` for certain schemas. However, those same schemas will still need a `base_path` if they're in nested links.


sihugh:
Thanks :-D

PR #837, opened 2018-11-19
Require locale in redirect gone and vanish messages

cbaines:
Relates to:

https://github.com/alphagov/publishing-api/pull/1390
https://github.com/alphagov/publishing-api/pull/1391
https://github.com/alphagov/publishing-api/pull/1392


cbaines:
This should fix the Content Performance Manager tests hopefully https://github.com/alphagov/content-performance-manager/pull/1010

PR #839, opened 2018-11-20
Add EU exit business finder metadata tags to all relevant schemas

barrucadu:
Should match Rummager schema stuff:

- https://github.com/alphagov/rummager/blob/master/config/schema/field_definitions.json
- https://github.com/alphagov/rummager/search?q=appear_in_find_eu_exit_guidance_business_finder&unscoped_q=appear_in_find_eu_exit_guidance_business_finder

[Trello card](https://trello.com/c/Z3jG94IX/68-add-tags-to-the-content-items)

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -320,6 +320,9 @@
         },
         "external_related_links": {
           "$ref": "#/definitions/external_related_links"
+        },
+        "tags": {
+          "$ref": "#/definitions/tags"
         }
       }
     },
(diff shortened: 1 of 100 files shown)


formats/shared/definitions/eu_exit_business_finder_metadata.jsonnet
```diff
@@ -0,0 +1,29 @@
+{
```
thomasleese: I don't think we need this file anymore.

thomasleese:
I guess we don't need this PR now?

rubenarakelyan:
It would be good to split your changes and the ones made by running `rake` into separate commits, otherwise it's very difficult to see what you changed vs. what the script changed.

barrucadu:
I don't think we need this PR for now, we'll need to think about how stuff is actually getting tagged.

PR #843, opened 2018-12-03
Add variants for transactions

rubenarakelyan:
This commit adds a `variant` field to the `transaction` schema. Each `variant` has a number of fields related to those of the parent content item, as well as a `slug`. When a variant of a transaction is accessed via `/name-of-transaction/slug-of-variant`, the fields that are defined for the variant are rendered in preference to those of the parent content item.

RFC: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-097-verify-specific-start-pages.md

### DIFF
--- a/dist/formats/transaction/frontend/schema.json
+++ b/dist/formats/transaction/frontend/schema.json
@@ -346,6 +346,41 @@
           "type": "string",
           "format": "uri"
         },
+        "variants": {
+          "description": "List of transaction variants",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": [
+              "title",
+              "slug"
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "introductory_paragraph": {
+                "$ref": "#/definitions/body_html_and_govspeak"
+              },
+              "more_information": {
+                "$ref": "#/definitions/body_html_and_govspeak"
+              },
+              "other_ways_to_apply": {
+                "$ref": "#/definitions/body_html_and_govspeak"
+              },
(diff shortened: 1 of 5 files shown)


formats/shared/definitions/multi_variant_pages.jsonnet
```diff
@@ -0,0 +1,28 @@
...
+      ],
+      properties: {
+        title: {
+          type: "string",
+        },
+        slug: {
+          type: "string",
+          format: "uri",
+        },
+        transaction_start_link: {
```
thomasleese: I wonder if we could pull this out into a definition since it's referenced here and in the original `transaction.jsonnet` schema file.

formats/shared/definitions/multi_variant_pages.jsonnet
```diff
@@ -0,0 +1,28 @@
+{
+  transaction_variants: {
```
thomasleese: Since this is only used in the `transaction` format, I wonder if it's better for the definition to live in the `transaction.jsonnet` file rather than its own file.

rubenarakelyan: I've moved the whole lot back into the `transaction` file since it's only used there.

formats/shared/definitions/multi_variant_pages.jsonnet
rubenarakelyan: I've moved the whole lot back into the `transaction` file since it's only used there.

PR #844, opened 2018-12-05
Add has_brexit_update flag

bilbof:
This enables a has_brexit_update flag on Whitehall documents to be
stored in the content store and displayed by government-frontend.

Pair effort with @oscarwyatt


bilbof:
Closing this but keeping branch, as we are considering this change and may need this later.

PR #852, opened 2019-01-22
Remove supertypes from frontend schema

tijmenb:
This commit removes the supertypes from the allowed output of the content store. We're removing it from the content items in favour of a system where we get the supertype of a document from directly from the gem: https://github.com/alphagov/govuk_document_types.

This will make it easier to iterate the supertypes. We won't have to republish everything if a new type is created or one is modified.

We are keeping the notification schemas the same for now, because the content performance manager project still uses the values. Once those are removed, we'll retire the groups here entirely.

I expect the content store tests to fail at the moment because it hasn't been updated yet. 

https://trello.com/c/hp5BJD2i


formats/shared/default_properties/notification.jsonnet
```diff
@@ -14,4 +14,40 @@
   payload_version: {
     "$ref": "#/definitions/payload_version",
   },
+  email_document_supertype: {
```
thomasleese: Worth noting that email-alert-api also currently uses this value and `government_document_supertype` to match a piece of content with its subscriber list: https://github.com/alphagov/email-alert-api/blob/master/app/queries/subscriber_list_query.rb#L2

tijmenb: 👍 cool thanks, I suspected as much so we'll keep the supertypes in the notifications until that's fixed.

thomasleese: As far as I know it's taking it directly from the notification, https://github.com/alphagov/email-alert-service/blob/394b64617689ecdd3b3ee285021ef284e41fc7e0/email_alert_service/models/major_change_message_processor.rb#L41 comes from RabbitMQ, and then it's used here https://github.com/alphagov/email-alert-service/blob/394b64617689ecdd3b3ee285021ef284e41fc7e0/email_alert_service/models/email_alert.rb#L34-L35 .

tijmenb: Ah good, was confused for a minute 👍

PR #855, opened 2019-01-25
Add suggested ordered related items

karlbaker02:
This PR adds `suggested_ordered_related_items` to `links` for use in the upcoming Related Links A/B test. The `case_study` example for `frontend` is also updated, which will be used to test changes in _government_frontend_.

Trello: https://trello.com/c/byNrHEMw

### DIFF
--- a/formats/shared/base_links.jsonnet
+++ b/formats/shared/base_links.jsonnet
@@ -17,4 +17,5 @@
   },
   original_primary_publishing_organisation: "The organisation that published the original version of the page. Corresponds to the first of the 'Lead organisations' in Whitehall for the first edition, and is empty for all other publishing applications.",
   lead_organisations: "DEPRECATED: A subset of organisations that should be emphasised in relation to this content item. All organisations specified here should also be part of the organisations array.",
+  suggested_ordered_related_items: "New A/B test suggestions for related items, can be any page on GOV.UK. Mostly used for mainstream content to power the sidebar. Ordering of the links is determined by the editor in Content Tagger.",
 }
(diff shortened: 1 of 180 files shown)


karlbaker02:
@sihugh I'll take a look at this.

karlbaker02:
@sihugh This has now been updated to split the schema changes from the actual changes 👍

formats/shared/base_links.jsonnet
```diff
@@ -17,4 +17,5 @@
   },
   original_primary_publishing_organisation: "The organisation that published the original version of the page. Corresponds to the first of the 'Lead organisations' in Whitehall for the first edition, and is empty for all other publishing applications.",
   lead_organisations: "DEPRECATED: A subset of organisations that should be emphasised in relation to this content item. All organisations specified here should also be part of the organisations array.",
+  suggested_ordered_related_items: "New A/B test suggestions for related items, can be any page on GOV.UK. Mostly used for mainstream content to power the sidebar. Ordering of the links is determined by the editor in Content Tagger.",
```
tijmenb: > Mostly used for mainstream content to power the sidebar. Ordering of the links is determined by the editor in Content Tagger.

Is this correct for this link type?

Minor nitpick: I guess you'd call these `ordered_suggested_related_items` if the ordering is important, or `suggested_related_items` if it isn't.

karlbaker02: This is the same as the description for `ordered_related_items`: given that this link type is for the B variant of aforementioned link type, if it's not correct it should be fixed in both places.

karlbaker02: For the minor nitpick, ordering is important - this is being determined by the ML and will be updated in the content store in priority order. If we change the name here it's going to have to change in three other repositories too - happy to make this change but checking first.

tijmenb: Not sure if I understand correctly, but "Ordering of the links is determined by the editor in Content Tagger" seems wrong if the ordering is determined by the ML. 

There's nothing restricting https://github.com/alphagov/government-frontend/pull/1223 to mainstream content, so "Mostly used for mainstream content to power the sidebar" seems wrong as well.

The link name is fine as-is, I thought the convention was the prefix the link type with `ordered` if ordering was important, but there's no actual prior art for that so 🤷‍♂️

karlbaker02: @tijmenb Updated the description, if it's good I'll regenerate the schemas.

PR #865, opened 2019-02-20
Add facet group, facet and facet value types

steventux:
https://trello.com/c/zn5LBg3L/7-spike-how-can-we-improve-the-business-finder-tagging-system
https://trello.com/c/IVf8x6lk/224-spike-proof-of-concept-publishing-api-accepts-tagged-content-items
https://trello.com/c/dOGEoWAJ/111-add-facetgroup-facet-and-facetvalue-as-formats-in-content-schemas

Improve tagging by incorporating facets and facet values as standalone formats in the publishing pipeline.


steventux:
@rubenarakelyan @thomasleese @kevindew @bevanloon @barrucadu 
This is a very rough proposal of how we'd model tags/facets/metadata/stuff/doodads (I'm not sold any particular name) as content items to start paying off metadata tagging tech-debt.
We're doing a spike to see how this works in Publishing API in a dev VM and how it flows downstream to Rummager.
Any comments, pointers, ideas really welcome.

\cc @peteglondon @vanitabarrett @leenagupte

thomasleese:
This is definitely the sort of thing I had in mind, it looks good to me. I think the `facet` naming makes sense.

The only suggestion I have is that I wonder if we should be looking at how the email sign up page will be rendered at the same time. Currently the content item for it is unrelated to the finder itself, which means that in theory the facets for the email sign up page can go out of sync with the finder. It would be nice [to reconcile the two](https://trello.com/c/eSxYNXB2/110-finders-and-finder-email-signup-pages-use-different-facets) so we should make sure that the new `facet` and `facet_value` schemas have [enough information to also render the sign up pages](https://github.com/alphagov/govuk-content-schemas/blob/9c2f3afc8d3e10fe7a4031b9297f8152081f6be5/formats/finder_email_signup.jsonnet#L53-L66). I think that means we'd need to at least add a `prechecked` field to the `facet_value` schema. The `topic_name` maps to `label` and `key` maps to `value` so they're already covered. I'm unsure whether we'd need `radio_button_name` because in every example I've found, it's the same as the `topic_name`.

bilbof:
This is an interesting idea!

I wonder if it might make sense to move away from depending on the content store in finder-frontend, and instead use a postgres db behind finder-frontend.

This would:
- make managing the finders easier, if all the data is in one relational db,
- reduce the complexity of the codebase, as there would be less manipulating of content store hashes, and more class-based logic in models,
- and potentially make search faster.

For tagging purposes, rather than creating new content items for facet group, facet and facet value types, we could add relations to the db.

This is hijacking the spike a bit, but I thought I would bring this up, as if we do introduce more content items that finder-frontend will depend on it might be harder to do down the line :)

steventux:
@bilbof Given the object of these changes is to find a way to apply metadata to any GOV.UK content in a consistent manner.The eventual goal being to produce something backwards compatible with specialist document metadata _and_ finder results which have been arbitrarily tagged as part of ongoing EU exit finder enhancements, I'm not sure if introducing a different data-store would simplify matters. It also lies outside our established pattern of frontend apps talking directly to dbs*
We've modelled the taxonomy as content items, this proposal is similar, the difference being that this form of categorisation is of a fixed depth and arbitrarily broad as opposed to taxons which can be nested to varying depths.


* There's always an exception to this rule, but it's one we are trying to retire :)

barrucadu:
I definitely prefer having these tags in the content store rather than anywhere else.  Content items already have a lot of search-related metadata in them, so saying that some metadata is special and lives elsewhere feels like a lot of confusion for not that much gain to me.

bilbof:
OK that makes sense. I had assumed that additional calls to content store from finder-frontend would be made for facet data but that doesn't seem to be the case. If we still receive one content item then the request latency and complexity of finder frontend presumably wouldn't change as a result of this change.

steventux:
@bilbof That's correct, we don't intend to increase the calls to the content store for facet data, we'd hope eventually to press the entire `facet_group` (including expanded facets and facet values) into the finder links so it can read the facet config from this, but that's a few steps away yet.

PR #867, opened 2019-03-07
Add recent finder-frontend schema changes

andysellick:
Update schemas to include recent finder frontend changes. Specifically:

- add `option_lookup` for mapping keys to values that can be used to associate `allowed_values` with multiple values
- add `default` option for `label_value_pair`, so that radio facets can have a default option
- add `hide_facet_tag` key radio facets, where we don't want it to have a facet tag when an option is chosen. This is because radios have to have a default value, so removing it (via the facet tag) is confusing, because it can't be removed, only set to the default state.

Partly related to: https://github.com/alphagov/finder-frontend/pull/922

Trello card: https://trello.com/c/pNXeZLrW/403-introduce-subset-of-supergroup-facet-behaviour-on-stats-finder-m

### DIFF
--- a/formats/shared/definitions/finder.jsonnet
+++ b/formats/shared/definitions/finder.jsonnet
@@ -178,6 +178,7 @@
             "checkbox",
             "date",
             "hidden",
+            "radio",
             "taxon",
             "text",
             "topical",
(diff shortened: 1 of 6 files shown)


formats/shared/definitions/finder.jsonnet
```diff
@@ -190,6 +191,16 @@
             "$ref": "#/definitions/label_value_pair",
           },
         },
+        option_lookup: {
+          description: "A map of keys to values that can be used to associate allowed_values with multiple values",
+          type: "object",
+          items: {
```
bilbof: I'm not sure that items is valid here. I have a feeling the below would work. What you could do is try it by generating the schema locally, then validating your example with something like this:

```
GovukContentSchemaTestHelpers::Validator.new("finder", "schema", your_json)
```

```
        option_lookup: {
          description: "A map of keys to values that can be used to associate allowed_values with multiple values",
          type: "object",
          additionalProperties: true,
          patternProperties: {
            "^[a-z_]+$": {
              type: "array",
              items: {
                type: "string"
              }
            }
          }
```

andysellick: Fixed, thanks.

PR #868, opened 2019-03-08
Add option lookup to finder email signup

bilbof:
Similar to the finder facet option_lookup field,
this enables one to carry the same behaviour
(associating a single displayed value with multiple
values behind the scenes) from a finder to the email
signup page.

This will enable publication of the policy and consultations email signup page.

Related to https://github.com/alphagov/finder-frontend/pull/936.

https://trello.com/c/tIXZB4LG/482-make-post-review-changes-to-policy-papers-and-consultations-finder

### DIFF
--- a/dist/formats/finder_email_signup/frontend/schema.json
+++ b/dist/formats/finder_email_signup/frontend/schema.json
@@ -324,6 +324,19 @@
               },
               "filter_value": {
                 "type": "string"
+              },
+              "option_lookup": {
+                "description": "An array of key values where the key is the value of a selected facet and the value(s) are what these are converted to in a Rummager query",
+                "type": "object",
+                "additionalProperties": true,
+                "patternProperties": {
+                  "^[a-z_]+$": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  }
+                }
               }
             }
           }
--- a/dist/formats/finder_email_signup/notification/schema.json
+++ b/dist/formats/finder_email_signup/notification/schema.json
@@ -424,6 +424,19 @@
               },
               "filter_value": {
                 "type": "string"
+              },
+              "option_lookup": {
+                "description": "An array of key values where the key is the value of a selected facet and the value(s) are what these are converted to in a Rummager query",
+                "type": "object",
+                "additionalProperties": true,
+                "patternProperties": {
+                  "^[a-z_]+$": {
+                    "type": "array",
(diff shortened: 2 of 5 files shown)


dist/formats/finder_email_signup/frontend/schema.json
```diff
@@ -320,6 +320,19 @@
               },
               "filter_value": {
                 "type": "string"
+              },
+              "option_lookup": {
+                "description": "A map of keys to values to associate a displayed value with multiple options",
```
oscarwyatt: I'm not sure this is super clear, without more context it's not clear what we're mapping keys to/from, where they're displayed or why we might want to do this.

bilbof: I've changed the text and regenerated the schemas. Let me know if the new text is OK :)

bilbof:
The `content-performance-manager` integration test is failing due to an error in that repo. This is being investigated by the team working on that project and should be fixed later on today.

dist/formats/finder_email_signup/frontend/schema.json
oscarwyatt: Perhaps it's just the tiredness but based on this description I'm still not sure what this does (and I _think_ it's probably a good idea to make it clear here as I'm not sure where else people will be able to see what it does). Let's chat about it tomorrow mornign

bilbof: Sure thing. The current text matches that of the other `option_lookup` for finder facets: https://github.com/alphagov/govuk-content-schemas/pull/867/commits/d06f82b4b3abb43a898f52b587464adf13c7f0c6#diff-6cf8ee2493a65fcb847de8a4dbf8e0daR195.

bilbof:
Oscar requested this text:

> An array of key values where the key is the value of a selected facet and the value(s) are what these are converted to in a Rummager query

I'll make this change then rebuild the schema.

dist/formats/finder_email_signup/frontend/schema.json
karlbaker02: I'm happy to chat about this too, as I was the one who wrote the original description and perhaps thought that it could be clearer.

bilbof: Thanks @karlbaker02, see https://github.com/alphagov/govuk-content-schemas/pull/868#issuecomment-472045661

PR #878, opened 2019-03-27
Allow facet groups as filter in finder

peteglondon:
We are replacing `appear_in_find_eu_exit_guidance_business_finder` with `facet_groups` to 
query for things appearing in the EU exit business readiness finder,

The schema is preventing us from doing that.

Trello: https://trello.com/c/Sd7omEFH/263-mapping-rummager-request-response-in-finder-frontend

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -528,6 +528,12 @@
             "type": "string"
           }
         },
+        "facet_groups": {
+          "type": "array",
+          "items": {
+            "type": "string"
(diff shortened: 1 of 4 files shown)


peteglondon:
[Mapping Rummager request/response in Finder Frontend](https://trello.com/c/Sd7omEFH/263-mapping-rummager-request-response-in-finder-frontend)

sihugh:
I did a fix to make schema tests compile assets before starting, so I think your govt-frontend build should pass now

PR #880, opened 2019-03-28
Allow email subs with taxonomy filters

sihugh:
We want to allow filtering email subs by topic.

This corresponds with a Citizen Readiness finder: alphagov/rummager@87fc52b

Initially we're setting this up to just allow signups to a particular topic, but in future we want to allow combinations of eg brexit AND education


sihugh:
Send my regards to Jøn

PR #884, opened 2019-04-10
Add facet links pathless schema config

steventux:
https://trello.com/c/0wiO10tC/16-email-trigger-mechanism-for-new-content-tagger-process

Facet groups, facets and facet values don't have base paths so add them to the appropriate config in the schema generator.

### DIFF
--- a/lib/schema_generator/format.rb
+++ b/lib/schema_generator/format.rb
@@ -267,7 +267,7 @@ module SchemaGenerator
 
     class Links
       ALLOWED_KEYS = %w(description required minItems maxItems).freeze
-      LINKS_WITHOUT_BASE_PATHS = %w(ordered_contacts ordered_foi_contacts world_locations).freeze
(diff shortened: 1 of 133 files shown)


lib/schema_generator/format.rb
```diff
@@ -267,7 +267,7 @@ def determined_definition
 
     class Links
       ALLOWED_KEYS = %w(description required minItems maxItems).freeze
-      LINKS_WITHOUT_BASE_PATHS = %w(ordered_contacts ordered_foi_contacts world_locations).freeze
+      LINKS_WITHOUT_BASE_PATHS = %w(facet_groups facet_values ordered_contacts ordered_foi_contacts world_locations).freeze
```
leenagupte: Shouldn't `facet` be in this list as well?

leenagupte: Shouldn't `facet` be in this list as well?

steventux: @leenagupte good point - added.

steventux:
This isn't necessary any more

PR #891, opened 2019-05-01
Add finder to "links" in all content items

DilwoarH:
Trello: https://trello.com/c/uR7v1Rri
Depends on: https://github.com/alphagov/info-frontend/pull/444

## What
Allows "finders" to be added to the `links` hash of content items. At moment this is only needed on the business finder. Most people land on content, but then have no way of easily finding other Brexit related content that is relevant to their organisation.

## Why
If a piece of content knows about the finders it is surfaced by, then we can use this information to create a link back to the finder on page in the breadcrumbs and related links.

### DIFF
--- a/examples/guide/frontend/guide-with-facet-groups.json
+++ b/examples/guide/frontend/guide-with-facet-groups.json
@@ -1,0 +1,716 @@
+{
+  "analytics_identifier": null,
+  "base_path": "/national-curriculum",
+  "content_id": "be9f6021-f60a-47d2-a184-d1c31bc6eba0",
+  "document_type": "guide",
+  "first_published_at": "2016-02-29T09:24:10.000+00:00",
+  "locale": "en",
+  "phase": "live",
+  "public_updated_at": "2014-12-02T10:28:58.000+00:00",
+  "publishing_app": "publisher",
+  "rendering_app": "frontend",
+  "schema_name": "guide",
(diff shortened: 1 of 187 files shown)


examples/guide/frontend/guide-with-facet-groups.json
```diff
@@ -63,6 +63,8 @@
           "facet_group": [
             {
               "content_id": "52435175-82ed-4a04-adef-74c0199d0f46",
+              "base_path": "/blah",
```
steventux: Facet groups do not have base paths, so this example isn't representative of the link as it appears in the content store response.
I think the problem lies with how frontend schemas are generated, they don't cater for pathless links like publishing schema generation does. See https://github.com/alphagov/govuk-content-schemas/pull/884 for context. It might be better to fix this rather than create an unrealistic example.
I also don't understand the need for this example in this PR, is this example necessary to represent the link to a finder in the content item's `links["finder"]` guid list?

leenagupte: You're right, the `facet_group` isn't needed, but it does make for a more complete example when the `finder` is added to the links.

leenagupte: Thanks for the link to the fix! Implementing it now.

leenagupte:
@steventux I've updated the example. Can you re-review?

PR #893, opened 2019-05-09
Add alternative format email for orgs

benthorner:
https://trello.com/c/OcuVomSb/820-determine-the-alternative-format-email-address-for-an-attachment

This adds a new field to store an email that specifies how a user can
request an alternative format from the organisation e.g. for an
attachment.

### DIFF
--- a/dist/formats/organisation/frontend/schema.json
+++ b/dist/formats/organisation/frontend/schema.json
@@ -334,6 +334,13 @@
             "null"
           ]
         },
+        "alternative_format_contact_email": {
+          "description": "The organisation's email for requesting an alternative format for attachments",
+          "type": [
+            "string",
+            "null"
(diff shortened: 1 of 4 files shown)


dist/formats/organisation/frontend/schema.json
```diff
@@ -334,6 +334,13 @@
             "null"
           ]
         },
+        "alternative_format_contact_email": {
+          "description": "The organisation's email for requested an alternative format for attachments",
```
1pretz1: I think you mean 'requesting' here instead of 'requested'.

benthorner: Ah, thanks for checking - fixed :-).

kevindew:
Nice one 👍

PR #894, opened 2019-05-17
Add secondary to step by steps

oscarwyatt:
Adds secondary to step by steps so that a
content item can link to a step by step as it 
may be a part of the process but it is not
essential to completing it.

Trello: https://trello.com/c/5zRe5f7F/96-step-by-step-update-the-content-schemas-to-allow-secondary-content-to-be-linked-to-a-step-by-step

### DIFF
--- a/examples/publication/frontend/best-practice-guidance.json
+++ b/examples/publication/frontend/best-practice-guidance.json
@@ -190,6 +190,236 @@
         "links": {
         }
       }
+    ],
+    "secondary_to_step_by_steps": [
+      {
+        "title": "Learn to drive a car: step by step",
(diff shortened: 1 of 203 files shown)


examples/publication/frontend/best-practice-guidance.json
```diff
@@ -190,6 +190,21 @@
         "links": {
         }
       }
+    ],
+    "secondary_to_step_by_steps": [
```
leenagupte: This should have the full step by step content item in it, e.g. https://github.com/alphagov/govuk-content-schemas/blob/master/examples/guide/frontend/guide-with-step-navs.json#L18

oscarwyatt: 👍 Added that in now. I'm looking at publishing-api

PR #898, opened 2019-05-23
Link reject finders

SamJamCul:
Adds an additional possible rejection filter to the finder schema, so we can block results with particular links. This is being done as part of [this ticket](https://trello.com/c/qAwjASRS/176-remove-the-search-page-from-search-results-searchception-s)

### DIFF
--- a/dist/formats/finder/frontend/schema.json
+++ b/dist/formats/finder/frontend/schema.json
@@ -597,6 +597,12 @@
             "type": "string"
           }
         },
+        "link": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "policies": {
           "type": "array",
           "items": {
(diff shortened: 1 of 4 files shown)


dist/formats/finder/frontend/schema.json
```diff
@@ -597,6 +597,12 @@
             "type": "string"
           }
         },
+        "links": {
```
sihugh: I think this might need to be singular `link` (like the content purpose supergroup one above) as it gets transformed to `reject_link` in the query

SamJamCul: Ah no problem, should I also change the <policies> above to <policy>?

SamJamCul: Done, it all says link instead now 👍

PR #900, opened 2019-05-29
Add combine_mode to email alert signup

oscarwyatt:
This parameter will be added to email-alert-api
and a finder-frontend will need to pass it to
the api selectively.

Existing finders will want to continue using
'and' (which can be represented either with
a blank string or omitting the parameter)
while the business finder will want to use
'or' joining

Trello: https://trello.com/c/vLZ7zTdR/132-add-parameter-to-finder-frontend-to-make-email-alert-api-know-this-finder-is-special

### DIFF
--- a/formats/finder_email_signup.jsonnet
+++ b/formats/finder_email_signup.jsonnet
@@ -178,6 +178,14 @@
             },
           },
         },
+        combine_mode: {
+          type: "string",
+          description: "Controls which logic facets on the subscriber list should be joined by. Default is blank which maps to 'and'",
+          enum: [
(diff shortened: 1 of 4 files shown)


formats/finder_email_signup.jsonnet
```diff
@@ -178,6 +178,14 @@
             },
           },
         },
+        join_facets_with: {
```
brucebolt: I'm wondering if `join_facets_with` could be named `combine_mode` to make it consistent with the finder content item (and also optimise for deletion later on)?

PR #901, opened 2019-05-30
Adds official_documents filter to finder schema

hannako:
Adds `has_official_documents` to the list of fields that a finder can filter on.

See trello card [724](https://trello.com/c/6xFJqfAw/724-spin-up-finder-for-content-that-has-official-doc-status-s)

### DIFF
--- a/formats/shared/definitions/finder.jsonnet
+++ b/formats/shared/definitions/finder.jsonnet
@@ -91,6 +91,9 @@
       format: {
         type: "string",
       },
+      has_official_document: {
+        type: "boolean"
+      },
       organisations: {
         type: "array",
         items: {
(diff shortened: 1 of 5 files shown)


formats/shared/definitions/finder.jsonnet
```diff
@@ -203,6 +206,7 @@
             "hidden_clearable",
             "radio",
             "research_and_statistics",
+            "official_documents",
```
bilbof: This looks like a new type of facet, is it possible we can share some logic between research_and_statistics and this one?

hannako: Yes, we have two custom facets now, and there is some duplicated logic. My thought was if a third comes along we should refactor.

PR #912, opened 2019-06-27
Use organisation ids and user ids for access limiting

1pretz1:
This changes the publishing_api's content schema to accept organisation
ids as well as user ids for access limiting. This is due to the changes
content publisher are making on how access limiting works.

Trello:
https://trello.com/c/wxp1XkSg/969-update-govuk-to-support-access-limiting-by-organisation

### DIFF
--- a/formats/shared/definitions/publishing_api_base.jsonnet
+++ b/formats/shared/definitions/publishing_api_base.jsonnet
@@ -15,6 +15,12 @@
           type: "string",
         },
       },
+      organisations: {
+        type: "array",
+        items: {
+          type: "string",
+        },
+      },
       auth_bypass_ids: {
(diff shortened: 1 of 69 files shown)


formats/shared/definitions/publishing_api_base.jsonnet
```diff
@@ -15,6 +15,12 @@
           type: "string",
         },
       },
+      organisations: {
+        type: "array",
+        items: {
+          type: "string",
+        },
+      },
```
kevindew: This should be a `"$ref": "#/definitions/guid_list"` like auth_bypass_ids. It'd also be useful to have a description and have that state that this is organisation content ids.

PR #914, opened 2019-07-08
Wait and propagate dependant app builds

cbaines:
And propagate the build result, so that the build results for
dependant applications are taken in to account for the GitHub branch
protection configuration.

Previously, as the main job is the only one taken in to account for
the branch protections, Pull Requests could be merged when the changes
were incompatible with applications.


cbaines:
Closing this. I remembered why we stopped doing this previously... if any dependant build fails, you have to build everything again.

I'll try changing the GitHub branch protection configuration instead.

PR #915, opened 2019-07-17
This uses the given definition for a guid list for orgs

JonathanHallam:
Stops hardcoding the org list to only contain strings and to
use the given definition for a list of guids

based on comments on #912

### DIFF
--- a/dist/formats/answer/publisher_v2/schema.json
+++ b/dist/formats/answer/publisher_v2/schema.json
@@ -131,10 +131,8 @@
           "$ref": "#/definitions/guid_list"
         },
         "organisations": {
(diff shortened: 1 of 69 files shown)


dist/formats/answer/publisher_v2/schema.json
```diff
@@ -131,10 +131,8 @@
           "$ref": "#/definitions/guid_list"
         },
         "organisations": {
-          "type": "array",
-          "items": {
-            "type": "string"
-          }
+          "description": "A list of organisation ids permitted access to this item",
```
kevindew: I think this comment should reference that they are organisation content ids

PR #920, opened 2019-08-19
 Start removing superflous fields for `facet`

tijmenb:
In my (really quite limited) understanding, the fields that are defined in the `facet` schema aren't actually used for anything. 

- They're not used in the content-tagger tagger UI that people use to tag content
- They're not used in finder-frontend and search-api because we never expose the `facet` of a content item (only the `facet_group` and `facet_value`)

I'm 100% confident about this! Please tell me if I'm wrong.

### DIFF
--- a/formats/facet.jsonnet
+++ b/formats/facet.jsonnet
@@ -7,53 +7,48 @@
     details: {
       type: "object",
       additionalProperties: false,
-      required: [
-        "filterable",
-        "key",
-        "name",
-        "type",
-      ],
       properties: {
         filter_key: {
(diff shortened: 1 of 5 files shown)


formats/facet.jsonnet
```diff
@@ -7,53 +7,48 @@
     details: {
       type: "object",
       additionalProperties: false,
-      required: [
-        "filterable",
-        "key",
-        "name",
```
bilbof: I think these are used by finder-frontend at https://github.com/alphagov/finder-frontend/blob/master/app/models/facet.rb#L7.

tijmenb:
Closing this as the team won't use the facet system anymore, so leaving this for later!

PR #923, opened 2019-10-02
Add Myanmar to Schema's

callumknights:
Part of work to finish the country name change from burma to myanmar.


formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -589,6 +589,7 @@
             "montenegro",
             "morocco",
             "mozambique",
+	    "myanmar",
```
cbaines: The indentation looks a bit off here.

formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1631,6 +1632,7 @@
...
```
cbaines: Same here https://github.com/alphagov/govuk-content-schemas/pull/923/commits/11c95b96e364acfeb2fbccc3ea302382222d23fd#r330930781

formats/shared/definitions/_specialist_document.jsonnet
callumknights: Done!

formats/shared/definitions/_specialist_document.jsonnet
callumknights: Also done!

PR #925, opened 2019-10-03
Switch rummager -> search-api in dependant applications

cbaines:
As rummager was renamed to the search-api quite a while ago.


cbaines:
I've manually removed `rummager` from the list of required status checks to allow merging this Pull Request.

We also have configuration to update, and I've raised a Pull Request for this here https://github.com/alphagov/govuk-saas-config/pull/35

PR #928, opened 2019-10-10
Allow auth_bypass_ids outside access_limited

kevindew:
Trello: https://trello.com/c/DnYanQdY/132-separate-auth-bypassing-from-access-limiting-in-publishing-api

This adds a new field that Publishing API can accept of auth_bypass_ids,
this is part of work to separate auth_bypass_ids from access_limits.
These are being separated because auth_bypass_ids are not a part of
access limiting and can be set independent of that feature.

They auth_bypass_ids field in access_limited has been marked as
deprecated and will continue to be so until there is a drive to update
all GOV.UK publishing apps to not use it.

### DIFF
--- a/dist/formats/answer/publisher_v2/schema.json
+++ b/dist/formats/answer/publisher_v2/schema.json
@@ -19,6 +19,10 @@
     "analytics_identifier": {
       "$ref": "#/definitions/analytics_identifier"
     },
+    "auth_bypass_ids": {
+      "description": "A list of ids that will allow access to this item for non-authenticated users",
+      "$ref": "#/definitions/guid_list"
+    },
     "base_path": {
       "$ref": "#/definitions/absolute_path"
     },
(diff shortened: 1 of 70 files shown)


formats/shared/default_properties/notification.jsonnet
```diff
@@ -14,6 +14,12 @@
   payload_version: {
     "$ref": "#/definitions/payload_version",
   },
+  access_limited: {
+    "$ref": "#/definitions/access_limited",
+  },
+  auth_bypass_ids: {
+    "$ref": "#/definitions/guid_list",
+  },
```
1pretz1: Unless I'm misunderstanding something, why do we want allow the addition of access_limited and auth_bypass_ids to the data for a message in the message queue as these fields will be superfluous to that part of the system?

kevindew: I was just writing they are already sent and will continue to be sent - however I've just realised that because we only send live to the message queue they won't actually need to be sent so can be removed. Thanks I'll sort that out.

kevindew:
@1pretz1 updated

PR #929, opened 2019-11-04
Done pages have a new 'electric vehicle' promo choice

issyl0:
- The government has a goal to reach zero carbon emissions by 2050, and
  electric cars are part of that.
- Content designers can now select "electric vehicles" as their promo of
  choice.
- These schemas need to be updates to include 'electric_vehicle' as a
  valid option.

https://trello.com/c/FmE5xU1p/1478-3-add-new-bespoke-message-to-specific-completed-transaction-pages

### DIFF
--- a/formats/completed_transaction.jsonnet
+++ b/formats/completed_transaction.jsonnet
@@ -17,6 +17,7 @@
                 "mot_reminder",
                 "organ_donor",
                 "register_to_vote",
+                "electric_vehicle",
               ],
             },
             url: {
@@ -48,6 +49,12 @@
             {
               properties: {
                 category: { enum: ["register_to_vote"] }
+              },
+              required: ["url"]
+            },
+            {
+              properties: {
+                category: { enum: ["electric_vehicle"] }
               },
               required: ["url"]
             },
--- a/dist/formats/completed_transaction/frontend/schema.json
+++ b/dist/formats/completed_transaction/frontend/schema.json
@@ -333,6 +333,18 @@
                   ]
                 }
               }
+            },
+            {
+              "required": [
+                "url"
+              ],
+              "properties": {
+                "category": {
+                  "enum": [
+                    "electric_vehicle"
+                  ]
+                }
+              }
             }
           ],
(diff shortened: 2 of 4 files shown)


formats/completed_transaction.jsonnet
```diff
@@ -51,6 +52,12 @@
               },
               required: ["url"]
             },
+            {
+              properties: {
+                category: { enum: ["electric_vehicle"] }
+              },
+              required: ["url"]
```
issyl0: Is it required? The [Trello card](https://trello.com/c/FmE5xU1p/1478-3-add-new-bespoke-message-to-specific-completed-transaction-pages#comment-5db70e607675b9862d474390) gave us some text to hardcode. We decided to do it the "proper" way, but do content designers need to set the URL, or can this be entirely hardcoded in https://github.com/alphagov/frontend/pull/2089?

thomasleese: The card says we can hard code the text, but I do like doing it the “proper” way.

However, since the text of the link is hard coded in frontend anyway I don’t see any benefit to allowing the URL to be customisable.

Perhaps we should hard code it at the beginning and change it if we need to?

thomasleese:
You’ll also need to rebuild all the content schemas. I think ‘bundle exec rake’ should do that.

formats/completed_transaction.jsonnet
thomasleese: Although I see now the way it works in publisher is that the URL input is always visible. Perhaps it’s easiest to be consistent with the others and require the content designer to choose the URL.

issyl0:
I did that in the previous force push after CI complained. Just did it again and there were no changes, so I think we're good?

thomasleese:
@issyl0 Ah yes. Sorry I was going by the number of commits 🤦‍♂. Usually the rebuild step would go in a commit on its own, although this change is pretty small so it matters less.

issyl0:
@thomasleese I split the commits. Could you re-review please?

PR #932, opened 2019-11-19
Add reverse links for appointments

thomasleese:
This new type of reverse link is being added to the Publishing API so we need to add it to the Content Schemas.

See https://github.com/alphagov/publishing-api/pull/1645 for more details.

[Trello Card](https://trello.com/c/roHfuPUV/1362-use-reverse-links-for-role-appointments)


thomasleese:
I've chosen the name `appointments` since it's the same as `ordered_current_appointments` except that I don't believe we can make any guarantees on the ordering any more and because it will include both the `current` and the `previous` appointments.

PR #933, opened 2019-11-19
Add more fields to role appointments

thomasleese:
This adds two new fields to role appointments to support the new reverse links. This came out of comments on https://github.com/alphagov/publishing-api/pull/1645.

#### `current`

This will be set to true if the role appointment is "current". Another way of finding this out is to look at whether then `ended_on` field is set, but I thought it would be good to pull this out as denormalised data to simplify the logic of the frontends.

#### `person_importance_ordering`

This is a new field on role appointments which can be used to order appointments with the same `started_on` date by "importance".

This is used to order the appointments in some places on the frontend, notably the title which gets shown above the name of the person.

An alternative to putting this into the role appointment would be a field in the person, but that adds a dependency between the person and the role appointment meaning that if a new role appointment is created, a new edition of a person would need to be drafted to include the new role appointment in the details.

I struggling come up with a name for this field so any suggestions are welcome.

[Trello Card](https://trello.com/c/roHfuPUV/1362-8-use-reverse-links-for-role-appointments)

### DIFF
--- a/dist/formats/role_appointment/frontend/schema.json
+++ b/dist/formats/role_appointment/frontend/schema.json
@@ -303,12 +303,18 @@
         "change_history": {
           "$ref": "#/definitions/change_history"
         },
+        "current": {
+          "type": "boolean"
+        },
         "ended_on": {
           "type": [
             "string",
             "null"
           ],
           "format": "date-time"
+        },
+        "person_appointment_order": {
+          "type": "integer"
         },
         "started_on": {
           "type": "string",
--- a/dist/formats/role_appointment/notification/schema.json
+++ b/dist/formats/role_appointment/notification/schema.json
@@ -423,12 +423,18 @@
         "change_history": {
           "$ref": "#/definitions/change_history"
         },
+        "current": {
+          "type": "boolean"
+        },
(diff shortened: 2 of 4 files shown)


kevindew:
current sounds good 👍 

I can understand the challenges in the person ordering naming. I'm not too sure about the person_importance prefix as I think it implies the person's importance rather than the appointments precedence.

I do wonder if the ordering is needed though as I think if we have an ordered list of role appointments then this order would persist on the person to go the other way. Unless (yet again) I'm missing something.

thomasleese:
@kevindew We need the ordering because the `person` -> `role_appointments` link (used by the frontend to render the person page) is the reverse link added by the Publishing API which makes no guarantees on the ordering. The explicit link is the `role_appointment` -> `person` one which has no associated ordering (because the target will always be a single person, not a list of people).

The reason we wanted the `person_` prefix was so that when the `role_appointment` appeared expanded from a role, it's very clear that "importance" is not relevant for the role.

How about `importance_for_person_order` instead?

thomasleese:
After discussion outside we've decided to go with `person_appointment_order` as the name.

PR #935, opened 2019-11-22
Add a government schema

cbaines:
In preparation for Whitehall publishing governments to the Publishing API.

### DIFF
--- a/formats/government.jsonnet
+++ b/formats/government.jsonnet
@@ -1,0 +1,25 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: "government",
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {
+        started_on: {
+          type: "string",
+          format: "date-time",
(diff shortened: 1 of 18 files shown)


cbaines:
This depends on https://github.com/alphagov/content-data-api/pull/1367

formats/government.jsonnet
```diff
@@ -0,0 +1,25 @@
+(import "shared/default_format.jsonnet") + {
+  document_type: "government",
+  definitions: {
+    details: {
+      type: "object",
+      additionalProperties: false,
+      properties: {
+        start_date: {
```
thomasleese: Could we call this `started_on` and the other one `ended_on` for consistency with similar naming in role appointments?

cbaines: Yeah, that sounds fine :+1:

cbaines: I've updated this Pull Request, and the govuk-content-schemas change now.

PR #938, opened 2019-11-27
Add government link to examples

cbaines:
Where the example includes "government" in the details.


cbaines:
I'm still fighting the govuk-content-schemas... :(

I think I should include locale in the expansion of the governments: https://github.com/alphagov/publishing-api/pull/1657 and then I'll need to update all the relevant examples to include the locale here.

PR #942, opened 2019-12-10
Add political status to organisations schema

AgaDufrat:
We want to present organisations political status from Whitehall to Publishing
API so we can use this information in Content Publisher. It is currently
hardcoded in the YAML file and requires manual updates.

[Trello card](https://trello.com/c/zHm2B8VX/1242-present-organisation-political-status-from-whitehall-to-publishing-api)

### DIFF
--- a/formats/organisation.jsonnet
+++ b/formats/organisation.jsonnet
@@ -318,6 +318,10 @@
           ],
           description: "The type of organisation.",
         },
+        "political": {
+          "description": "Political status of the organisation used for history mode",
+          "type": "boolean"
+        },
         social_media_links: {
           type: "array",
           items: {
--- a/dist/formats/organisation/frontend/schema.json
+++ b/dist/formats/organisation/frontend/schema.json
@@ -653,6 +653,10 @@
             "tribunal"
           ]
         },
+        "political": {
+          "description": "Political status of the organisation used for history mode",
+          "type": "boolean"
+        },
         "secondary_corporate_information_pages": {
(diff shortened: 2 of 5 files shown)


formats/organisation.jsonnet
```diff
@@ -318,6 +318,10 @@
           ],
           description: "The type of organisation.",
         },
+        "political": {
+          "description": "Political status of the organisation used for history mode",
+          "type": "boolean"
+        },
```
cbaines: So this key is present in lots of content already, but are the semantics the same?

For other types of content, this is described as "If the content is considered political in nature, reflecting views of the government it was published under.".

Even though the definition in the schema gives a different description, I think the possibility of confusion is quite high given the same name is used.

kevindew: Hey @cbaines I'm just replying since I wrote the card.

I agree political is an overloaded term and far from ideal. I however felt that since this "political organisation" was already cemented by Whitehall then calling this by any other name just meant this concept would have two names and confusion. 

There also seemed to be a lack of good alternative names that aren't very verbose. Suggestions seem to be "publishes_current_government_policies" or "associated_with_government_policies". Neither of which seem very attractive options.

If you feel strongly about this and/or have good ideas of naming I'm all ears to chat further.

kevindew: Btw I'm also keen to revise this description of this field.

cbaines: I don't feel particularly strongly about this, but `organisation_political` comes to mind as an alternative. I haven't checked if that's used anywhere else, but it is similar to some of the other `organisation_` prefixed values in the details.

PR #944, opened 2019-12-10
Rename political to organisation_political

gclssvglx:
Renamed due to 'political' already in use and an overloaded term.

[Trello card](https://trello.com/c/zHm2B8VX/1242-present-organisation-political-status-from-whitehall-to-publishing-api)

### DIFF
--- a/dist/formats/organisation/frontend/schema.json
+++ b/dist/formats/organisation/frontend/schema.json
@@ -632,6 +632,10 @@
             }
           }
         },
+        "organisation_political": {
+          "description": "Determines whether content published by this organisation represents governments policies and can be eligible for history mode",
+          "type": "boolean"
+        },
         "organisation_type": {
           "description": "The type of organisation.",
           "type": "string",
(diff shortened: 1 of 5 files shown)


dist/formats/organisation/frontend/schema.json
```diff
@@ -632,6 +632,10 @@
             }
           }
         },
+        "organisation_political": {
```
1pretz1: Having a ponder on the name of this and wouldn't it make more sense to be 'political_organisation'?

gclssvglx: Your point is well taken, and there has been much pondering over the name.  On balance, we've chosen to go with @cbaines suggestion from [here](https://github.com/alphagov/govuk-content-schemas/pull/942#discussion_r356040867).

1pretz1: Okay dokay, I'll not re-open a can of worms 😆

PR #950, opened 2020-01-16
Add brexit no deal notice

koetsier:
Adds the possibility to add a notice to Whitehall content explaining where updated content is.

The links can be external or internal. A title has to be provided for each link.

https://trello.com/c/0URepvoe/270-update-whitehall-content-schema-to-include-a-no-deal-notice


koetsier:
I'll hold off on merging until we know 100% these are the formats

koetsier:
the formats have been confirmed as:
Publication
+ (HTML attachment)
Detailed Guide
Document Collection
Case Study

PR #951, opened 2020-01-17
Remove hide facets by default field

bilbof:
As of alphagov/search-api@1ce164d this field will no longer be used.

Should be merged and deployed after alphagov/search-api#1920 has gone out.

https://trello.com/c/PtSyQ5hi/1285-show-facets-by-default


bilbof:
Search API e2e tests won't pass until the PR mentioned above is merged.

PR #952, opened 2020-01-17
Add required field to email signup facets

bilbof:
The required boolean on email facets will enable us to
specify in the content item whether a facet must have a
value. This is useful for CMA cases and Research and Statistics
finder email signup content items.

https://trello.com/c/NS1ZPbBF/1249


bilbof:
@barrucadu I'm planning to make this a required field once I've updated the content items.

PR #953, opened 2020-01-20
Add format to permitted filter fields

bilbof:
Related to https://github.com/alphagov/specialist-publisher/pull/1563

We want to permit format as a default filter in email signups.

https://trello.com/c/yLsG8ecR/1292-fix-specialist-publisher-email-signups

### DIFF
--- a/examples/finder_email_signup/frontend/cma-cases-email-signup.json
+++ b/examples/finder_email_signup/frontend/cma-cases-email-signup.json
@@ -76,6 +76,9 @@
   "description": "You'll get an email each time a case is updated or a new case is published.",
   "details": {
     "beta": false,
+    "filter": {
+      "document_type": "cma_case"
+    },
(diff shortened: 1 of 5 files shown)


examples/finder_email_signup/frontend/cma-cases-email-signup.json
```diff
@@ -76,6 +76,9 @@
   "description": "You'll get an email each time a case is updated or a new case is published.",
   "details": {
     "beta": false,
+    "filter": {
+      "document_type": "cma_case"
```
barrucadu: Should this example use `format`?

bilbof: They should be, but the tests in finder-frontend depend on these examples so will fail. I'll think about how to change the finder-frontend tests so they're more flexible to schema changes like this.

PR #959, opened 2020-02-11
Remove appear_in_find_eu_exit_guidance_business_finder field

laurentqro:
This was initially used by the EU exit business finder, but no longer was. And the business finder has now been decommissioned anyway.


leenagupte:
@laurentqro It looks like this can be closed, as the changes were covered in: https://github.com/alphagov/govuk-content-schemas/pull/961

PR #969, opened 2020-03-10
Set Whitehall document attachments to be file_attachment_asset

kevindew:
Prior to this being merged https://github.com/alphagov/whitehall/pull/5448 will have to be merged and deployed to production.

Previously all of these documents were set to use the publication attachment schemas, however none of these documents make use of publication specific fields or types. Thus this reduces the options that can apply at a schema level for them.

@barrucadu I've opened this up and marked it as a draft prior to you get your input on this. I'm wondering if there was a reason these document types were set as publication attachment assets in the first place that I've missed and/or if there are consequences to this change? might it break any Whitehall document types.

### DIFF
--- a/dist/formats/corporate_information_page/frontend/schema.json
+++ b/dist/formats/corporate_information_page/frontend/schema.json
@@ -337,7 +337,7 @@
           "description": "An ordered list of asset links",
           "type": "array",
           "items": {
-            "$ref": "#/definitions/publication_attachment_asset"
+            "$ref": "#/definitions/file_attachment_asset"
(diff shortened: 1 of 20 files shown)


kevindew:
> As long as we're sure that none of these formats (eg) allow HTML or external attachments, this seems fine.

Thanks Michael - turns out all of them were leaking publication data into Publishing API I've opened https://github.com/alphagov/whitehall/pull/5448 to try resolve them. May need to get some tips from you on how to represent everything to clear that up.

PR #975, opened 2020-03-25
Add new examples for direct attachment rendering

benthorner:
https://trello.com/c/wM9wqe9H/1587-support-featured-attachments-directly-on-govuk-frontend

Previously we relied on pre-rendered HTML for the attached 'documents'
of publications and consultations. This adds an example to support work
to directly render attachments based on metadata in the payload, and an
array of featured attachment IDs. See RFC 116 for more details.

https://github.com/alphagov/govuk-rfcs/blob/master/rfc-116-store-attachment-data-in-content-items.md


benthorner:
I did attempt to update some of the other examples, but the content changes mean this could cause random test failures wherever we're referencing the content directly.

PR #976, opened 2020-03-25
Define new coronavirus_landing_page schema

hannako:
Once this is merged I will update the the schema name of the [content item](https://www.integration.publishing.service.gov.uk/api/content/coronavirus) from `generic` to `coronavirus_landing_page`

### DIFF
--- a/examples/coronavirus_landing_page/frontend/coronavirus_page.json
+++ b/examples/coronavirus_landing_page/frontend/coronavirus_page.json
@@ -1,0 +1,351 @@
+{
+  "base_path": "/coronavirus",
+  "content_id": "774cee22-d896-44c1-a611-e3109cce8eae",
+  "document_type": "coronavirus_landing_page",
+  "description": "Find out about the government response to coronavirus (COVID-19) and what you need to do.",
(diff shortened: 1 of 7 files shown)


examples/coronavirus_landing_page/frontend/coronavirus_page.json
```diff
@@ -0,0 +1,352 @@
...
+  "public_updated_at": "2020-03-25T14:40:22Z",
+  "rendering_app": "collections",
+  "schema_name": "coronavirus_landing_page",
+  "title": "Coronavirus (COVID-19): what you need to do",
+  "locale": "en",
+  "updated_at": "2020-03-25T14:40:22Z",
+  "links": {
+  },
+  "details": {
+    "title": "Coronavirus (COVID-19): what you need to do",
```
sihugh: We could drop title and meta_description from the details hash, just keep them in the top level

hannako: I've moved title out of details, but having meta_description in the top level doesn't seem to be permitted.

PR #979, opened 2020-04-16
Add ministerial to roles

MahmudH:
If the role is ministerial we want to present the ministers index page.

Trello: https://trello.com/c/i1ffrXiZ/1913-5-link-ministers-to-the-government-ministers-content-item

### DIFF
--- a/dist/formats/role/frontend/schema.json
+++ b/dist/formats/role/frontend/schema.json
@@ -107,6 +107,11 @@
           "description": "The user needs this piece of content meets.",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
+        "ministerial": {
+          "description": "Link to the ministers index page, present if this is a ministerial role.",
+          "$ref": "#/definitions/frontend_links_with_base_path",
+          "maxItems": 1
+        },
         "ordered_parent_organisations": {
           "description": "Organisations that own this role.",
           "$ref": "#/definitions/frontend_links_with_base_path"
--- a/dist/formats/role/notification/schema.json
+++ b/dist/formats/role/notification/schema.json
@@ -127,6 +127,11 @@
           "description": "The user needs this piece of content meets.",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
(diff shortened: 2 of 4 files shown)


thomasleese:
This looks like a good start, I have a few questions.

It looks like the link has been added to role appointments, I wonder if it makes more sense for the link to be on the roles themselves? In Whitehall [that's handled by the role](https://github.com/alphagov/whitehall/blob/master/app/models/ministerial_role.rb#L25-L27) and it means we wouldn't need to republish all the role appointments for a role if a role changes and becomes ministerial in the future.

I'm also unsure about the naming `is_ministeral`. It implies that the value will be a boolean in my mind, not a link. I wonder if we can instead [rely on the existing `parent`/`children` link in this case](https://github.com/alphagov/publishing-api/blob/15b975acb4e964d83173db2a5f35a7c135262340/lib/expansion_rules.rb#L37), so a role links to the ministers page if it's a ministerial role.

barrucadu:
> It looks like the link has been added to role appointments, I wonder if it makes more sense for the link to be on the roles themselves? In Whitehall that's handled by the role and it means we wouldn't need to republish all the role appointments for a role if a role changes and becomes ministerial in the future.

That makes sense to me

> I'm also unsure about the naming is_ministeral. It implies that the value will be a boolean in my mind, not a link. I wonder if we can instead rely on the existing parent/children link in this case, so a role links to the ministers page if it's a ministerial role.

I'm also not too keen on the name (even though I'm the one who suggested it).  I was unsure about using the `parent` link, but I guess we're not going to have any other page as the parent of these.

thomasleese:
> I was unsure about using the `parent` link, but I guess we're not going to have any other page as the parent of these.

Yeah, having thought about this a bit more, I think there are benefits to the link type being explicit for this case, so maybe it is better to have our own link type.

MahmudH:
I've taken the suggestions on board and have moved the changes into role and renamed to ministerial. These changes make more sense, thank you @thomasleese.

PR #981, opened 2020-04-27
Add phase_message optional field

laurentqro:
This PR adds support for a `phase_message` optional field for all content items.

This will be useful especially for the `publisher` app where we want to be able to set this for a given content item, rather than globally across an application.

Related to https://github.com/alphagov/publisher/pull/1250.

### DIFF
--- a/formats/shared/default_properties/all.jsonnet
+++ b/formats/shared/default_properties/all.jsonnet
@@ -23,6 +23,9 @@
       "live",
     ],
   },
+  phase_message: {
+    "$ref": "#/definitions/body_html_and_govspeak",
+  },
   publishing_app: {
     "$ref": "#/definitions/publishing_app_name",
   },
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -181,6 +181,9 @@
         "live"
       ]
     },
+    "phase_message": {
+      "$ref": "#/definitions/body_html_and_govspeak"
+    },
     "public_updated_at": {
       "anyOf": [
         {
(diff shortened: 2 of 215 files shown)


formats/shared/default_properties/all.jsonnet
```diff
@@ -23,6 +23,10 @@
       "live",
     ],
   },
+  phase_message: {
+    description: "The message associated with the design phase for this content item",
+    type: "string",
```
kevindew, line 28: Rather than a string I think this is likely to contain either Govspeak or HTML. 

The precedent for this is to use the [multiple_content_types definition](https://github.com/alphagov/govuk-content-schemas/blob/7865213ca35962fc5ef80dd90b8652de0e5799b2/formats/shared/definitions/publishing_api_base.jsonnet#L70-L88)

kevindew:
@laurentqro first off sorry, out of habit I force pushed a branch and overwrote yours. I've checkout your commit and re-pushed that.

I had a play around in: https://github.com/alphagov/govuk-content-schemas/commit/dc430ae7b3d24dad31e554c8401d3d650ba5f492 which should be good for allowing rendered govspeak into the content item.

I did realise we will hit a pain point though once we render govspeak on the frontend as the component will be expecting only inline elements but this will block. I'm going to have a little play but I'll probably run out of time before tomorrow.

Gosh so much pain to just configure a small thing.

kevindew:
I had a play with components here: https://github.com/alphagov/govuk_publishing_components/pull/1473

It does feel rather painful the amount of changes needed for this hmm I wonder if there are simpler options. Suppose we could take message, link and link text fields though that's quite a lot of fields.

laurentqro:
Not pursuing this feature, so closing this PR.

PR #986, opened 2020-05-04
Remove details-based people links

thomasleese:
We now have this information in the links rather than in the details hash so we don't need these in the schema anymore.

See https://github.com/alphagov/whitehall/pull/5575 for more details.

[Trello Card](https://trello.com/c/NRhnejWA/1942-5-send-links-to-people-instead-of-details-hash-in-the-organisations-in-whitehall)

### DIFF
--- a/lib/schema_generator/format.rb
+++ b/lib/schema_generator/format.rb
@@ -270,7 +270,21 @@ module SchemaGenerator
 
     class Links
       ALLOWED_KEYS = %w(description required minItems maxItems).freeze
-      LINKS_WITHOUT_BASE_PATHS = %w(facet_groups facets facet_values ordered_contacts ordered_foi_contacts world_locations).freeze
+      LINKS_WITHOUT_BASE_PATHS = %w(
+        facets
+        facet_groups
+        facet_values
+        ordered_board_members
+        ordered_chief_professional_officers
+        ordered_contacts
+        ordered_foi_contacts
+        ordered_military_personnel
+        ordered_ministers
+        ordered_roles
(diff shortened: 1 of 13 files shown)


lib/schema_generator/format.rb
```diff
@@ -270,7 +270,20 @@ def determined_definition
 
     class Links
       ALLOWED_KEYS = %w(description required minItems maxItems).freeze
-      LINKS_WITHOUT_BASE_PATHS = %w(facet_groups facets facet_values ordered_contacts ordered_foi_contacts world_locations).freeze
+      LINKS_WITHOUT_BASE_PATHS = %w(
+        facet_groups
+        facets
+        facet_values
+        ordered_contacts
+        ordered_foi_contacts world_locations
```
AlanGabbianelli, line 278:
```suggestion
        ordered_foi_contacts
        world_locations
```

thomasleese: Thanks, surprised the linter didn't catch this!

thomasleese:
> I can't help but feel the need to include huge content items here for testing is a bit of a smell...

Yeah, I'm never sure whether the examples are supposed to be minimal testing examples, or real life examples that real exists...

sihugh:
There's a few different versions of org pages. I keep meaning to re-organise the tests in collections to make it obvious which one's which.  We built them incrementally, so grew into a shape that's probably not optimal.  We were concerned when we were migrating that we weren't catching all the special cases (like the Number 10 door banner) which is why these examples exist. You're right that a little bit of a tidy up would be useful

PR #1001, opened 2020-08-24
Add research_for_development_output specialist publication

edwardkerry:
This PR adds a new Research for Development Output schema published by [SpecialistPublisher](https://github.com/alphagov/specialist-publisher/pull/1692)

[Trello](https://trello.com/c/wye58SZu/2109-update-x2-dfid-specialist-finders-to-fcdo)


examples/specialist_document/publisher_v2/research_for_development_output.json
```diff
@@ -0,0 +1,66 @@
...
+        "AL",
+        "AF"
+      ],
+      "fcdo_theme": [
+        "agriculture",
+        "infrastructure"
+      ],
+      "fcdo_authors": [
+        "Mr. Potato Head",
+        "Mrs. Potato Head"
```
AlanGabbianelli, line 52: 🥔

edwardkerry: these are existing examples, not my own!

formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1955,6 +1958,286 @@
...
     },
   },
+  research_for_development_output_metadata: {
+    type: "object",
+    additionalProperties: false,
+    properties: {
+      bulk_published: {
+        type: "boolean",
+      },
+      fcdo_review_status: {
```
thomasleese, line 1968:
```suggestion
      review_status: {
```

I would if we could make all these fields that reference `fcdo` generic so we don't have to deal with this in the future... I know they're not presented on the frontend, but if we're going to change them from `dfid_` I feel like we may as well make it generic now.

formats/shared/definitions/_specialist_document.jsonnet
```diff
@@ -1955,6 +1958,286 @@
...
+            "health",
+            "humanitarian_disasters_and_emergencies",
+            "infrastructure",
+            "research_communication_and_uptake",
+            "social_change",
+            "water_and_sanitation",
+          ],
+        },
+      },
+      fcdo_document_type: {
```
thomasleese, line 2209: I guess this one could be problematic because of it's name... Maybe `research_document_type`?

thomasleese (approved):

PR #1004, opened 2020-09-01
Remove dfid research outputs schema

edwardkerry:
The DFID Research Output has been superceded by the [Research for Development Output](https://github.com/alphagov/specialist-publisher/pull/1692) document type, and so this code can be removed.

[trello](https://trello.com/c/wXE33m26/2117-remove-dfidresearchoutputs-finder)


thomasleese (approved):
This is fine, I'm just wandering if there are any implications of removing a document type from here that still exists historically (the older editions will still be of document type `dfid_research_output`)... I don't know if we've ever done it before. I guess if anyone comes across it they can find this PR so I think it's okay, it won't affect anything on the site I don't think.

edwardkerry:
Closing this until a decision is made

PR #1005, opened 2020-09-03
Update to Ruby 2.7

Rhian:
These are the necessary changes for Ruby 2.7, tested against 2.7.1

Since there is an automated process for updating Ruby versions in govuk apps, I changed the version locally, made the necessary updates, but raised a PR with version 2.6.6. 

- Deprecation warning from jsonnet gem
```
/ruby-2.7.1/gems/jsonnet-0.3.0/lib/jsonnet/vm.rb:35: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/ruby-2.7.1/gems/jsonnet-0.3.0/lib/jsonnet/vm.rb:87: warning: The called method `evaluate_file' is defined here
```

This is fixed in master but not the latest version
Master: 
- https://github.com/yugui/ruby-jsonnet/blob/master/lib/jsonnet/vm.rb#L35 
- https://github.com/yugui/ruby-jsonnet/blob/master/lib/jsonnet/vm.rb#L87
v0.3.0: 
- https://github.com/yugui/ruby-jsonnet/blob/v0.3.0/lib/jsonnet/vm.rb#L35 
- https://github.com/yugui/ruby-jsonnet/blob/v0.3.0/lib/jsonnet/vm.rb#L87


cdccollins:
Jenkins couldn't find bundler version 2.1.4

Gemfile
```diff
@@ -1,20 +1,21 @@
 source "https://rubygems.org"
+
 ruby File.read(".ruby-version").strip
```
cdccollins, line 3: Think we've decided to remove the ruby version from Gemfile.lock

Gemfile.lock
```diff
@@ -1,97 +1,115 @@
...
+  rack-test
   rake
   rspec
-  rubocop-govuk (~> 3.3.0)
-  sinatra (~> 2.0)
-
-RUBY VERSION
-   ruby 2.6.6
+  rubocop-govuk
+  sinatra
```
cdccollins, line 112: when I bundled this locally, these were removed. Don't know what's happening there but might be worth checking out.

cdccollins (changes requested):

lib/schema_generator/expanded_links.rb
```diff
@@ -1,58 +1,58 @@
...
-     # items.
-     "role_appointments" => "frontend_links",
-
-     # `Role` content items that are ministerial roles will automatically
-     # have a `ministers` link type from the main `ministers` index page.
-     "ministers" => "frontend_links",
-   }.freeze
+      # The Publishing API will automatically link to any translations (content
+      # with the same content_id but a different locale).
+      "available_translations" => "frontend_links_with_base_path",
```
cdccollins, line 6: Just checking if you mean to add an extra tab?

cdccollins:
Another PR supersedes https://github.com/alphagov/govuk-content-schemas/pull/1012

PR #1008, opened 2020-09-18
Add whip organisation to role schema

callumknights:
Part of the work to migrate rendering of the government/ministers page from
whitehall to collections.

https://trello.com/c/loAlkIp7/2063-8-migrate-government-ministers-to-collections

This change is needed as I have added a new field to the `details` in
`role_presenter`, this change adds that field to ensure tests on CI pass.

### DIFF
--- a/formats/role.jsonnet
+++ b/formats/role.jsonnet
@@ -49,8 +49,25 @@
         supports_historical_accounts: {
           type: "boolean",
         },
-        seniority :{
+        seniority: {
           type: "integer",
+        },
+        whip_organisation: {
+          type: "object",
+          properties: {
+            sort_order: {
+              type: [
+                "integer",
+                "null",
+              ],
+            },
+            label: {
+              type: [
+                "string",
+                "null",
+              ],
+            },
+          },
         },
       },
     },
--- a/dist/formats/role/frontend/schema.json
+++ b/dist/formats/role/frontend/schema.json
@@ -358,6 +358,23 @@
         },
         "supports_historical_accounts": {
           "type": "boolean"
+        },
+        "whip_organisation": {
+          "type": "object",
+          "properties": {
+            "label": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "sort_order": {
+              "type": [
+                "integer",
+                "null"
+              ]
+            }
+          }
         }
       }
     },
--- a/dist/formats/role/notification/schema.json
+++ b/dist/formats/role/notification/schema.json
@@ -479,6 +479,23 @@
         },
         "supports_historical_accounts": {
           "type": "boolean"
+        },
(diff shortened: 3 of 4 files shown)


formats/role.jsonnet
```diff
@@ -49,9 +49,26 @@
         supports_historical_accounts: {
           type: "boolean",
         },
-        seniority :{
+        seniority: {
```
ChrisBAshton, line 52: Good spot 👍

formats/role.jsonnet
```diff
@@ -49,9 +49,26 @@
...
           type: "integer",
         },
+	whip_organisation: {
```
ChrisBAshton, line 55: Looks like you've used tabs instead of spaces. Whilst that's an age-old debate we shouldn't have in a PR, what threw me was that the commas on lines 72 and 73 don't line up, so it looked as though there was one `}` too many (which would be invalid JSON). I then noticed the problem starts on this line - we should be at the same indentation level as the closing `}` on line 54.

I suggest you go through this file and fix up the spacing like this:

```suggestion
        whip_organisation: {
```

ChrisBAshton (commented):
Content looks good to me, just need to look at the whitespace! 👍

formats/role.jsonnet
callumknights: Argh, this happens every time I do a quick edit in vim and use tabs out of habit 😆

formats/role.jsonnet
```diff
@@ -49,9 +49,26 @@
         supports_historical_accounts: {
           type: "boolean",
         },
-        seniority :{
+        seniority: {
           type: "integer",
         },
+	      whip_organisation: {
+	       type: "object",
```
ChrisBAshton, line 56: Still looks like some tabs have crept in 😅 I don't usually have this problem myself because VS Code fixes space formatting on save (I think due to [this extension](https://marketplace.visualstudio.com/items?itemName=chenxsan.vscode-standardjs)). There may be an equivalent for Vim...?

callumknights: Oh my god just saw this review, will fix now 👍

formats/role.jsonnet
```diff
@@ -49,9 +49,26 @@
         supports_historical_accounts: {
           type: "boolean",
         },
-        seniority :{
+        seniority: {
           type: "integer",
         },
+        whip_organisation: {
+	        type: "object",
```
ChrisBAshton, line 56: Line 55 looking good now, but still seeing tabs in all the lines below, and you can see the closing `}` on line 71 is too indented as a result.

callumknights: Finally I think it's okay, dear lord 😱

ChrisBAshton (approved):
LGTM now 👍

PR #1009, opened 2020-09-21
Updating specialist document formats for North Macedonia

Tetrino:
Country name changed a couple years ago but we had some leftovers in some document types.


cbaines (approved):
The commit will "Update specialist document formats for North Macedonia", rather than "Updating specialist document formats ..."

The diff looks fine though.

PR #1014, opened 2020-09-25
Allow North Macedonia and Macedonia temporarily

benthorner:
https://trello.com/c/J6DOuBBp/990-country-updates

This is part of our cleanup work for the renaming of this country.
We should ensure these documents can still be updated while the
actual renaming is taking place. Once this has been done, we can
safely remove "Macedonia" from the allowed countries.


JonathanHallam (approved):
👍

PR #1011, opened 2020-09-22
Reinstates "Updating specialist document formats for North Macedonia"

Tetrino:
Country name changed a couple years ago but we had some leftovers in
some document types.

This reverts commit c96d031485f2b3643bb99969bdcf68dc1fe70fc0.

PR #1015, opened 2020-09-28
Remove "Macedonia" (migrated to "North Macedonia")

benthorner:
https://trello.com/c/J6DOuBBp/990-country-updates

Closes: https://github.com/alphagov/govuk-content-schemas/pull/1011


dist/formats/specialist_document/frontend/schema.json
```diff
@@ -1467,6 +1465,7 @@
               "montenegro",
               "montserrat",
               "morocco",
+              "north-macedonia",
```
JonathanHallam, line 1468: Think these need to go slightly further down between North Korea and North Sudan as you have for the health certificates

benthorner: Thanks for spotting @JonathanHallam 💯. I've remembered how the alphabet works now, and pushed up a fix.

JonathanHallam (approved):
looks good to me 🎉

PR #1019, opened 2020-10-14
Run tests against content-publisher

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

This is another application that uses content schemas.


kevindew (approved):
Nice one, I'm surprised we missed this all those moons ago.

PR #1022, opened 2020-10-26
Make 'ordering' property a required field in Take Part pages

ChrisBAshton:
Depends on https://github.com/alphagov/whitehall/pull/5856

It should be safe to make this a required attribute, as it is
enforced on save in Whitehall:
https://github.com/alphagov/whitehall/blob/master/app/models/take_part_page.rb#L7

Trello: https://trello.com/c/aAW8oqCk/2174-5add-ordering-to-take-part-pages-in-schema

### DIFF
--- a/dist/formats/take_part/frontend/schema.json
+++ b/dist/formats/take_part/frontend/schema.json
@@ -300,7 +300,8 @@
       "type": "object",
       "required": [
         "body",
-        "image"
+        "image",
+        "ordering"
       ],
       "additionalProperties": false,
       "properties": {
--- a/dist/formats/take_part/notification/schema.json
+++ b/dist/formats/take_part/notification/schema.json
@@ -402,7 +402,8 @@
       "type": "object",
       "required": [
         "body",
-        "image"
+        "image",
+        "ordering"
       ],
       "additionalProperties": false,
       "properties": {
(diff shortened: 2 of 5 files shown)


AlanGabbianelli (approved):

ChrisBAshton:
(Force pushed a rebase against master).

ChrisBAshton:
Hmmm, making `ordering` required makes [this Whitehall importer test in Content Publisher](https://github.com/alphagov/content-publisher/commit/051a09e00ad7c23a135a11321b7a6447e1eca3e6#diff-f2603f5ce0fb1de34fcdc5d9d88f3b307cc28a9939b490bea01489f68d002671R46-R60) fail. Not sure what's going on yet - there is no instance of 'take_part' in Content Publisher, and I've tried adding `ordering { 1 }` to all of the whitehall import factories and that doesn't fix it either.

It could be that, as hinted by the Trello card, it may not be possible to make this a 'required' field. But it seems a shame for it to fail on an import script in a currently-on-hold publishing app.

ChrisBAshton:
Given the other priorities in this sprint and the lack of developers available, I'm inclined to close this. Hopefully @Tetrino might be able to shed more light on this at some point.

PR #1027, opened 2020-11-30
Add brexit_current_state_notice

hannako:
## What

Add a new field to the schema of:
- CaseStudy
- DetailedGuide
- DocumentCollection
- Publication
- HTML publication

## Context

- We are adding a button to Whitehall to allow publishers to display a new type of call out box on their transition related content. Here's the [PR](https://github.com/alphagov/whitehall/pull/5894) .
- Currently publishers can add a call out box for post transition content, this is called the `brexit_no_deal_notice` and is present on docs such as [this](https://www.gov.uk/api/content/guidance/driving-in-the-eu-from-1-january-2021). 
- A document will only ever have either a `brexit_no_deal_notice` or a `brexit_current_state_notice`. This is determined by a validation on the Whitehall database. 

---
https://trello.com/c/IwXzb5FM/589-design-whitehall-publisher-functionality-to-generate-current-state-notice

### DIFF
--- a/dist/formats/case_study/frontend/schema.json
+++ b/dist/formats/case_study/frontend/schema.json
@@ -279,6 +279,26 @@
       "description": "The main content provided as HTML rendered from govspeak",
       "type": "string"
     },
+    "brexit_current_state_notice": {
+      "description": "A list of URLs and titles about the current state of guidance that might change after brexit.",
+      "type": "array",
+      "items": {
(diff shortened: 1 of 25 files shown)


sihugh (approved):

sihugh:
You can deploy this one now if you're confident that we've got everything

PR #1028, opened 2020-12-02
Add body to step by steps

sihugh:
Site search can only ingest the title and description of a step by step at present. 

The body of a step by step is written in markdown-ish, which we can convert to html using kramdown or similar. By adding this html to the body field of the content item, we'll be able to search for them much more easily.


leenagupte (approved):
We spoken IRL about what the `body` field is going to be populated with, and it all makes sense to me.  It's certainly a nicer option than populating `hidden_search_terms`. 👍

PR #1030, opened 2020-12-03
Add schema for Get Involved static page for use with Government-Frontend

Tetrino:
Functionally this is a duplicate of the History schema (see
/formats/history.jsonnet) however we determined it best to keep it
separate rather than rename the History schema in case we wish to expand
it at a later date.

This is related with other work to migrate the Get Involved page from
Whitehall to Government-Frontend.

https://trello.com/c/7BjH9PyC/2083-8-migrate-government-get-involved-to-government-frontend


thomasleese (commented):
The code changes look good to me, I think it would be good to separate out the commits where you've added to the new schema and where you've rebuilt them all to make it easier to see your specific change.

Tetrino:
@thomasleese Good shout, new commits put in.

thomasleese (approved):

PR #1048, opened 2021-02-25
Update contact headings

chris-gds:
## What?

Updating schemas to reflect work done on this PR https://github.com/alphagov/govspeak/pull/206

## Why?

Headings have been changed to paragraph copy within contact information due to a11y.  They shouldn't be bold as it looks like a heading but is not also the h3 is not always correct when used in the flow of a document

## Anything else?
Live Ref page
https://www.gov.uk/government/news/new-voluntary-calorie-guidelines-to-help-industry-tackle-obesity

image

image


alex-ju (approved):
Updating these examples makes sense after updating `govspeak` to replace the heading with a paragraph.

However, I think there is a tradeoff with [the change itself](https://github.com/alphagov/govspeak/pull/206). I understand there were problems with the heading level, but with this change, people will find it harder to reach the contact details using the landmark navigation.

chris-gds:
> However, I think there is a tradeoff with [the change itself](https://github.com/alphagov/govspeak/pull/206). I understand there were problems with the heading level, but with this change, people will find it harder to reach the contact details using the landmark navigation.

🤔  I take your point, ideally this entire contact section needs _something_ so a user can easily navigate to it.  First thoughts thinking `<section>` & `<address>` would aid with this. Will revisit.

alex-ju:
> > However, I think there is a tradeoff with [the change itself](https://github.com/alphagov/govspeak/pull/206). I understand there were problems with the heading level, but with this change, people will find it harder to reach the contact details using the landmark navigation.
> 
> 🤔 I take your point, ideally this entire contact section needs _something_ so a user can easily navigate to it. First thoughts are thinking `<section>` & `<address>` would aid with this. Will revisit.

Indeed, the contact section could be improved with semantic markup and microdata, but I'm not sure if it's in the scope of this work.

chris-gds:
Captured suggestions for prioritisation by PM

PR #1054, opened 2021-04-01
Add smart_answer schema

theseanything:
Previously, content items used to represent a Smart Answer were a transaction schema for the start page and a generic_with_external_related_links for the flow. This is because start pages were rendered by Frontend. However, we are changing Smart Answers to render the start pages and represent flows with a single content item. This adds schema to use for that content item.


alex-ju (approved):
`content-data-api` seems to be unhappy with this, but we may only need to try again

theseanything:
Thanks @alex-ju! I've done this PR to fix content data api https://github.com/alphagov/content-data-api/pull/1598

PR #1055, opened 2021-04-12
Deprecate supertypes that aren't used downstream

sihugh:
These supertypes used to be consumed by downstream apps (eg search-api and content-data). We've done some work to remove this dependency, so that the only app still left relying on finding supertypes via a notification is email-alert-api.

Our strategy is that supergroups are a point in time feature of a document based on its type. We should therefore devolve working these out to apps that wish to use them, and minimise the complexity of everything else.

- [Content-data-api not longer looks for them](alphagov/content-data-api@48a5c97#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3)
- [Search-api works them out for itself](https://github.com/alphagov/search-api/blob/987fa2313c0b35528eb87b892dccbd1cfd692dca/lib/indexer/document_preparer.rb#L155)
- [Publishing-api still sends them, but we don't want it to](https://github.com/alphagov/publishing-api/blob/a6ec79b0a8d77b991e0dd192f89dd88f9fe0b687/app/presenters/edition_presenter.rb#L52)

Email alert api still uses government_document_supertype and email_document_supertype so we'll leave them for another day.

https://trello.com/c/hp5BJD2i/117-unused-incorrect-supertypes-in-content-items

### DIFF
--- a/dist/formats/answer/notification/schema.json
+++ b/dist/formats/answer/notification/schema.json
@@ -15,7 +15,6 @@
     "govuk_request_id",
     "links",
     "locale",
-    "navigation_document_supertype",
     "payload_version",
     "phase",
     "public_updated_at",
@@ -25,8 +24,7 @@
     "routes",
     "schema_name",
     "title",
-    "update_type",
-    "user_journey_document_supertype"
+    "update_type"
   ],
   "additionalProperties": false,
   "properties": {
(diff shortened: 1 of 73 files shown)


hannako (approved):
Grepping alphagov for `navigation_document_supertype` and `user_journey_document_supertype` I came across a few references, mainly specs and documentation. But [this](https://github.com/alphagov/content-tagger/blob/a998e33c69c257303502200c773cc0e71e4fcddc/app/services/taxonomy/taxon_tree_export.rb#L47) one looks like a thing... is that fine?

PR #1058, opened 2021-05-07
Specify defaultBranch when calling pushTag

thomasleese:
The [`pushTag` method](https://github.com/alphagov/govuk-jenkinslib/blob/7d9befbef4b81e889fb7fb9f4562ec9afd730f30/vars/govuk.groovy#L700) accepts an argument for the default branch, which we need to change to be `main` rather than the default of `master`, otherwise the tag doesn't actually get pushed.


hannako (approved):
👍

PR #1060, opened 2021-05-18
Add devolved_administration_availability to local_transaction schema

Rosa-Fox:
# What?

This schema has been added as part of a task to enable publishers to specify devolved administration (DA) service availability for local transactions. 

# Why?

In the Frontend application, by default, users will be shown a link to a local authority website when inputting a postcode for a local transaction.

In some cases, the service is handled by a devolved administration. In other cases the service is unavailable to devolved administrations.

Currently, the devolved administration availability cannot be set by publishers and the logic is defined in the Frontend application.

As part of the overall task, [functionality is being added to the Publisher application](https://github.com/alphagov/publisher/pull/1451) to allow this information to be set within the publishing tool.

Therefore, this schema validates the DA availability values set in Publisher.

Only the 'unavailable' and 'devolved_administration_service' related values will be presented from Publisher. This will be enough information to determine what to display on the Frontend as it will fall back to displaying the local authority results page.

# Useful links

Final naming decisions were agreed as a team during Dev chat on 26/05/2021. 

We changed the `type` value from `alternative_url` to `devolved_administration_service` as we felt it was confusing having `alternative_url` as a field name and as a `type` value. We also felt it better described the type of service.

[Modelling PR in Publisher](https://github.com/alphagov/publisher/pull/1456).

[Spike content schema branch](https://github.com/alphagov/govuk-content-schemas/pull/1060). 

[Spike write up](https://docs.google.com/document/d/1vOwGK_X4zUVuaMr4_KU4o8Jpfdn7dskqDq2cFjbQrBI/edit?usp=sharing).

[Spike PR for modelling in Publisher](https://github.com/alphagov/publisher/pull/1451).

[Trello](https://trello.com/c/Kgr3I4Tz/401-store-devolved-administration-local-transaction-data-in-model-and-govuk-content-schema).

Paired with @kevindew

### DIFF
--- a/dist/formats/local_transaction/frontend/schema.json
+++ b/dist/formats/local_transaction/frontend/schema.json
@@ -349,14 +349,63 @@
         "need_to_know": {
           "$ref": "#/definitions/body_html_and_govspeak"
         },
+        "northern_ireland_availability": {
+          "$ref": "#/definitions/devolved_administration_availability"
+        },
+        "scotland_availability": {
+          "$ref": "#/definitions/devolved_administration_availability"
+        },
         "service_tiers": {
           "description": "List of local government tiers that provide the service",
           "type": "array",
           "items": {
             "type": "string"
           }
+        },
+        "wales_availability": {
+          "$ref": "#/definitions/devolved_administration_availability"
         }
       }
+    },
+    "devolved_administration_availability": {
+      "description": "Used to indicate that a particular devolved administration has a different handling process for the service",
+      "oneOf": [
+        {
+          "description": "A value that indicates a service is unavailable",
+          "type": "object",
+          "required": [
+            "type"
+          ],
+          "additionalProperties": false,
+          "properties": {
+            "type": {
+              "enum": [
+                "unavailable"
+              ]
+            }
+          }
+        },
+        {
+          "description": "A value that indicates the service is available through an alternative url",
+          "type": "object",
+          "required": [
+            "type",
+            "alternative_url"
+          ],
+          "additionalProperties": false,
+          "properties": {
+            "alternative_url": {
+              "type": "string",
+              "format": "uri"
+            },
+            "type": {
+              "enum": [
+                "devolved_administration_service"
+              ]
+            }
+          }
+        }
+      ]
     },
     "external_link": {
       "type": "object",
--- a/examples/local_transaction/frontend/local_transaction_with_devolved_administration_availability.json
+++ b/examples/local_transaction/frontend/local_transaction_with_devolved_administration_availability.json
@@ -1,0 +1,331 @@
+{
+  "analytics_identifier": null,
+  "base_path": "/report-stray-dog",
+  "content_id": "06465dbe-dcac-4e0b-8143-ae5f8526d493",
+  "document_type": "local_transaction",
+  "first_published_at": "2016-02-29T09:24:10.000+00:00",
+  "locale": "en",
+  "phase": "live",
+  "public_updated_at": "2015-11-09T15:52:07.000+00:00",
+  "publishing_app": "publisher",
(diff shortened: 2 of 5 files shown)


dist/formats/local_transaction/frontend/schema.json
```diff
@@ -349,15 +349,64 @@
         "need_to_know": {
           "$ref": "#/definitions/body_html_and_govspeak"
         },
+        "northern_ireland_availability": {
```
leenagupte, line 352: Are there separate "availability" values because it's possible to have more than one? Is it possible to have  a list of "devolved_administration_availability" entries in the content item, that have an enum for the country?

dist/formats/local_transaction/frontend/schema.json
```diff
@@ -349,15 +349,64 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "alternative_url": {
+              "type": "string",
+              "format": "uri"
+            },
+            "type": {
+              "enum": [
+                "alternative_url"
+              ]
```
leenagupte, line 404: I think having "alternative_url" as the enum value and as a property is little confusing. I thought they were the same thing!

dist/formats/local_transaction/frontend/schema.json
Rosa-Fox: yeah there are more than one, it could be all 3.

formats/shared/definitions/devolved_administration_availability.jsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  devolved_administration_availability: {
+    description: "An object specifying the availability of service",
```
kevindew, line 3:
```suggestion
    description: "Used to indicate that a particular devolved administration has a different handling process for the service",
```

kevindew (commented):
This is looking good - I think you've got the syntax good now.

The few things I'd suggest are:

- split the generation of schema files into a separate commit (it's a common convention in this repo for example: https://github.com/alphagov/govuk-content-schemas/pull/1061/commits)
- Create an example schema that demonstrates the devolved administration fields (to pair with https://github.com/alphagov/govuk-content-schemas/blob/main/examples/local_transaction/frontend/local_transaction.json) this will allow us to test these schema changes against that example and potentially have something we can work with on the frontend in testing.

formats/shared/definitions/devolved_administration_availability.jsonnet
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
Rosa-Fox: thanks, I made the change.

formats/shared/definitions/devolved_administration_availability.jsonnet
Rosa-Fox: Thanks, I updated!

Rosa-Fox:
Thanks @kevindew. I split up the commits and added an example file. The schema regeneration commit seems to be failing to build and I am not sure if it is related. I'll look at it first thing tomorrow.

examples/local_transaction/frontend/local_transaction_with_devolved_administration_availability.json
```diff
@@ -0,0 +1,331 @@
...
+      "type": "devolved_administration_service",
+      "alternative_url": "https://www.gov.scot/stray-dog"
+    },
+    "wales_availability": {
+      "type": "unavailable"
+    },
+    "northern_ireland_availability": {
+      "type": "devolved_administration_service",
+      "alternative_url": "https://www.nidirect.gov.uk/stray-dog"
+    },
```
kevindew, line 326: Looks good 👍

kevindew (approved):
Nice one @Rosa-Fox

PR #1065, opened 2021-07-15
Move away from startup.sh in favour of govuk-docker

ChrisBAshton:
Sister PR: https://github.com/alphagov/govuk-docker/pull/503
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh


README.md
```diff
@@ -40,6 +18,20 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
...
+### Building the schemas
+
+```
+bundle exec rake build
+```
+
+### Running the tests
+
+```
+bundle exec rake spec
```
benthorner, line 30:
```suggestion
bundle exec rake
```

We should encourage people to run the default, as this includes linting (and mirrors what CI will run).

README.md
```diff
@@ -40,6 +18,12 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
 
 **Use GOV.UK Docker to run any commands that follow.**
 
+### Building the schemas
```
benthorner, line 21: Optional: since this is a non-standard section, it would be helpful to explain when someone should run this command e.g. to get validate a change to the schemas, without running the tests (which are separate).

README.md
```diff
@@ -2,35 +2,13 @@
...
-            └── schema.json
-
-# fixtures to support testing in apps
-examples
-└── case_study
-    └── frontend
-        ├── archived.json
-        ├── case_study.json
-        └── translated.json
+The [Jsonnet](https://jsonnet.org/) source files live in `formats`, and the schemas live in `dist` (which should not be edited manually). There are also example JSON files in `examples`, to support testing in apps.
```
benthorner, line 5:
```suggestion
The actual JSON schema files live in `dist` and are generated from [Jsonnet templates](https://jsonnet.org) in `source`. Using templates makes it easier to duplicate common blocks across multiple schemas e.g. format and URL fields. **Do not edit files in `dist` manually**, as they will be overwritten.
```

Optional: we can be clearer about the relationship between the two directories, and explain why we have them.

README.md
```diff
@@ -48,3 +40,7 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
 * [Why do contract testing?](docs/why-contract-testing.md)
 * [Running your frontend against the examples and random content (content-store not needed)](docs/running-frontend-against-examples.md)
 * [Deployment](docs/deployment.md)
```
benthorner, line 42: Optional: I think we can delete this file, since it only contains one line! We could explain this repo is deployed like any other app, but we still don't need a whole separate file to explain this.

README.md
```diff
@@ -48,3 +40,7 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
 * [Why do contract testing?](docs/why-contract-testing.md)
 * [Running your frontend against the examples and random content (content-store not needed)](docs/running-frontend-against-examples.md)
```
benthorner, line 41: Optional: this file still contains references to the dev VM. Do you think it's worth tidying it up as part of this PR?

README.md
```diff
@@ -40,6 +18,20 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
...
+
+### Running the tests
+
+```
+bundle exec rake spec
+```
+
+### Further documentation
+
 * [How to change a schema](docs/changing-a-schema.md)
```
benthorner, line 35: Very optional: this file and the one below seem to overlap with the "Suggested workflows" one (it's actually quite confusing which one to read first). We could just leave them as-is, but I thought I'd mention in case you feel inclined to merge them together.

benthorner (approved):
Nice one - nearly at the end now 🎉.

I left a few more minor comments. One thing to check is that the '###' headings don't render properly on my machine, but that might just be my markdown previewer.

Going out-of-scope, I notice there are a couple of files about contract testing in here that seem very unrelated to the repo itself. Since I know you/'re team have been working on this, it might be a good time to move / combine it into the dev docs.

README.md
ChrisBAshton: I did write that initially, but then adding the "build schemas" section felt odd, as I was specifying commands that are already run as part of the default task. I've tweaked the commit a bit to add this back in, as well as explain how to run the tests in isolation.

ChrisBAshton:
Thanks, as ever, for the very useful and thorough review, @benthorner. I've addressed your comments 👍 
Interesting note about the contract tests documentation.

I'd kind of forgotten that govuk-content-schemas has its own variation of contract tests, which are completely outside of the usual Pact Broker workflow. I think it's worth keeping this documentation close to the source, which will now be imported into the Developer Docs anyway, and I've tried to clarify in these docs that other alternatives exist.

There probably is a piece of work to do in terms of comparing the two competing contract testing types, and considering whether one should be consolidated into the other. That would be out of scope for this PR though!

ChrisBAshton:
The `contacts` contract test seems to be failing quite consistently on all the open PRs. I'm [looking into it](https://gds.slack.com/archives/CAB4Q3QBW/p1626778976182200).

docs/running-frontend-against-examples.md
```diff
@@ -26,22 +26,6 @@ To start an instance of the dummy content store, from this repo, run:
 $ bundle exec dummy_content_store
```
benthorner, line 26: This doesn't exist. I think it's actually just the `bundle exec ruby app.rb` command we've used in GOV.UK Docker. Even then, the guidance won't work if you're running with GOV.UK Docker. I'm not aware that anyone actually uses this functionality for local development, so one option would just be delete the last section.

docs/changing-an-existing-content-schema.md
```diff
@@ -0,0 +1,48 @@
...
+* populate the field in any records inside content-store
+* follow the General workflow, adding the field to the `required` attribute
+* you can then update the frontend app to expect the field to always be present
+
+For example, imagine that you need to add a new optional field to the details hash of the `case_study` schema. The steps would be:
+
+1. edit [`formats/case_study.jsonnet`](/formats/case_study.jsonnet) to
+   add the new optional field
+1. run `rake`. This will:
+  1. regenerate the publisher [`schema.json`](/dist/formats/case_study/publisher/schema.json) to incorporate the changes you made to the `details`
```
benthorner, line 28: Style issue: these show up as part of the top-level list when I preview. Maybe it's just my local previewer, but it might be worth adding a newline here. Also worth noting this section uses numbers, whereas others use bullets.

On the other hand, I don't think we need to say all this here. If anything it should be in the README section about rebuilding schemas.

docs/changing-an-existing-content-schema.md
```diff
@@ -0,0 +1,48 @@
...
+  1. regenerate the notification [`schema.json`](/dist/formats/case_study/notification/schema.json) to incorporate the changes
+  1. regenerate the frontend [`schema.json`](/dist/formats/case_study/frontend/schema.json) to incorporate the same changes
+  1. revalidate all example files to check if they are still valid after this change. This will pass, because the new field is optional
+1. [Optional step] you could add an additional example to illustrate how your new field should be used. You can add a new file in [examples/case_study/frontend](/examples/case_study/frontend)
+1. create a new branch and commit and push your changes
+  - this will run a branch build of govuk-content-schemas. This includes running the contract tests for each application which relies on the schemas. You'll get immediate feedback about whether publishing applications generate content items compatible with the new schema.
+1. once the tests pass, someone will merge your pull request and the new schemas will be available to use
```
benthorner, line 35: Style issue: numbering restarts for me here. In the past I've had to just number everything manually, as the auto-numbering doesn't always work.

Also, the bit about "someone will merge" is unnecessarily confusing - people should merge their own PRs. We could just lose these points as they're obvious.

docs/contract-testing-against-schemas.md
```diff
@@ -1,46 +1,49 @@
...
 
-The definition of the contracts are provided by [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas).
+We wanted to be able to evolve our publishing systems and have confidence that the changes we make will not break something. We could do this by running all the systems at the same time and testing end-to-end, but setting up such a configuration would be tricky, and keeping it working reliably in a reproducible way which avoids test interference would be very difficult.
 
-For more information and background please read the [README.md](../README.md)
+The idea of contract testing is to gain more confidence that the whole system works together properly, but without needing to run all the services at the same time. Instead we can test each service in isolation using a 'contract' which describes the expectations on each interacting service.
 
-## Overview of publishing
+## How govuk-content-schemas contract testing works
```
benthorner, line 15: I think this is actually about how the different schemas are generated, rather than the contract testing itself. Could we change this to clarify that?

You could move the last two paragraphs to the next section, as they're more about the principle of the tests.

docs/contract-testing-against-schemas.md
```diff
@@ -1,46 +1,49 @@
...
 
-Note that the *frontend representation* differs slightly from the *publishing
-representation*. The main difference is that the `links` hash is expanded so
+The first arrow is the "publishing representation", and the second arrow is the "frontend representation". The latter is more detailed than the former, as the `links` hash is expanded so
 that it contains full details about links, wheras the publisher representation
-only contains content ids (for full info see [frontend_schema_generator_spec.rb](spec/lib/frontend_schema_generator_spec.rb)).
-
-The *contract* between the publishing application and the frontend application
-is defined by [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas).
+only contains content ids (for full info see [frontend_schema_generator_spec.rb](../lib/schema_generator/frontend_schema_generator.rb)).
```
benthorner, line 25: This is a link to the code, rather than the test.

docs/contract-testing-against-schemas.md
```diff
@@ -1,46 +1,49 @@
...
-The idea of contract testing is to gain more confidence that the whole system works together properly, but without needing to run all the services at the same time. Instead we can test each service in isolation using a 'contract' which describes the expectations on each interacting service.
+## Background
+
+govuk-content-schemas contract testing was proposed in [RFC 4](https://gov-uk.atlassian.net/wiki/display/WH/RFC+4+%3A+Enabling+the+independent+iteration+of+formats+on+government-frontend) and described in the GOV.UK blog post "[Validating a distributed architecture with JSON Schema](https://gdstechnology.blog.gov.uk/2015/01/07/validating-a-distributed-architecture-with-json-schema/)".
 
-The definition of the contracts are provided by [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas).
+We wanted to be able to evolve our publishing systems and have confidence that the changes we make will not break something. We could do this by running all the systems at the same time and testing end-to-end, but setting up such a configuration would be tricky, and keeping it working reliably in a reproducible way which avoids test interference would be very difficult.
```
benthorner, line 11: I don't think these two paragraphs add anything. We could keep the previous one as part of the intro - that should be enough for anyone who wants to read the background.

benthorner (approved):
Nice one. I didn't realise we had two competing definitions of "contract test" (I just assumed this repo had the normal ones, which doesn't make much sense now I think about it.)

Totally agree with keeping the documentation here. If we move it somewhere else we should think about renaming the term so it's distinct from normal contract tests. Arguably testing against examples isn't a "contract".

I found a few lingering issues in the newly changed files. Apart from a couple of minor style things, I think this is still a net improvement, so happy for it to go in if you don't want to spend any more time on it.

docs/changing-an-existing-content-schema.md
ChrisBAshton: Fixed up the style issue, good spot.

docs/changing-an-existing-content-schema.md
ChrisBAshton: It works for me in my markdown previewer, so assume it will be fine 👍

docs/running-frontend-against-examples.md
ChrisBAshton: I've now removed this section 👍

ChrisBAshton:
Thanks for the review, @benthorner. I'm going to avoid making further changes to the 'contract testing' doc, now that the PR has been open for a month and the unrelated tests are finally passing! I suspect we'll do some contract testing documentation consolidation at some point, so this will probably be improved soon as part of a wider plan.

PR #1069, opened 2021-08-31
Add TakePart pages to linkable contents

Tetrino:
This PR allows us to patch-link Take Part pages to the Get Involved page
with intent of showing the needed data once the latter is moved to the
Government Frontend repo.

Trello: https://trello.com/c/7BjH9PyC/2083-8-migrate-government-get-involved-to-government-frontend

### DIFF
--- a/dist/formats/get_involved/frontend/schema.json
+++ b/dist/formats/get_involved/frontend/schema.json
@@ -261,6 +261,10 @@
         }
       ]
     },
+    "body": {
+      "description": "The main content provided as HTML rendered from govspeak",
+      "type": "string"
+    },
     "change_history": {
       "type": "array",
       "items": {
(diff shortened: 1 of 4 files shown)


dist/formats/get_involved/frontend/schema.json
```diff
@@ -304,6 +308,42 @@
         },
         "change_history": {
           "$ref": "#/definitions/change_history"
+        },
+        "take_part_pages": {
```
brucebolt, line 312: Should these have been added to the `links` section, rather than `details`?

Tetrino: Everything except for the .jsonnet is autogenerated as per the documentation, and the .jsonnet is correct based on other files, so this is just how messy these can become...

brucebolt: Thanks for confirming!  This appears to be consistent with other links too.

brucebolt (approved):
Looks good to me :+1:

PR #1072, opened 2021-09-13
Simplify the URI required for redirect destinations

kevindew:
Trello: https://trello.com/c/OE0mGWzN/1630-create-pharmacy-collect-short-urls-for-13-september

This changes the schema validation for redirect destinations from
something quite restrictive to something much more permissive. The
reason for taking a permissive approach here is because of duplication
of validation efforts.

Currently Publishing API has an involved redirect destination validation
process, which is used for both unpublishing and individual redirects [1].
This schema has not kept up with this validation and is currently out of
sync. Rather than trying to continue a process of keeping a schema and a
Rails validator in sync, it seems simpler and more accurate to make the
schema permissive and rely on the Publishing API validator as the
validating entity.

The prompt for this change was trying to adjust Short URL Manager to
allow redirects to hosts off GOV.UK (nhs.uk) this had the pain point
that Short URL Manager validates the URL, then Publishing API validates
the URL via Content Schema, and then Publishing API validates the URL
again through its Rails validator. The distribution of this logic is a
pain point.

[1]: https://github.com/alphagov/publishing-api/blob/3717c4057c0c05f9846a122ceb5972e559117d7c/app/validators/routes_and_redirects_validator.rb#L122

### DIFF
--- a/dist/formats/redirect/publisher_v2/schema.json
+++ b/dist/formats/redirect/publisher_v2/schema.json
@@ -114,11 +114,6 @@
     }
   },
   "definitions": {
-    "absolute_fullpath": {
-      "description": "A path with optional query string and/or fragment.",
-      "type": "string",
-      "pattern": "^/(([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})+(/([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)*)?(\\?([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?(#([a-zA-Z0-9._~!$&'()*+,;=:@-]|%[0-9a-fA-F]{2})*)?$"
-    },
     "absolute_path": {
       "description": "A path only. Query string and/or fragment are not allowed.",
       "type": "string",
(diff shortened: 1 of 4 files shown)


formats/shared/definitions/routes_redirects.jsonnet
```diff
@@ -62,17 +62,7 @@
...
-        anyOf: [
-          {
-            "$ref": "#/definitions/absolute_fullpath",
-          },
-          {
-            "$ref": "#/definitions/govuk_subdomain_url",
-          },
-        ],
-      },
+      destination: { type: "string", format: "uri" },
```
leenagupte, line 65: So we still have a little validation to make sure we're not just adding a generic string. Nice!

I was a bit concerned that perhaps relative paths wouldn't be allowed, but it looks like they are: https://datatracker.ietf.org/doc/html/rfc3986

formats/shared/definitions/routes_redirects.jsonnet
```diff
@@ -62,17 +62,7 @@
           "exact",
         ],
       },
-      destination: {
-        type: "string",
-        anyOf: [
-          {
-            "$ref": "#/definitions/absolute_fullpath",
```
leenagupte, line 69: This might be an enhancement for another time, but I wonder if all other uses of this pattern could be changed to  "uri" as well?

dist/formats/redirect/publisher_v2/schema.json
```diff
@@ -302,14 +292,7 @@
...
           "type": "string",
-          "anyOf": [
-            {
-              "$ref": "#/definitions/absolute_fullpath"
-            },
-            {
-              "$ref": "#/definitions/govuk_subdomain_url"
-            }
-          ]
+          "format": "uri"
```
leenagupte, line 295: Nice! Ignore my previous comment!

leenagupte (approved):
LGTM! 🥇

formats/shared/definitions/routes_redirects.jsonnet
kevindew: Yeah - I'm not sure if they're always validated to the same rigor as this one is via Publishing API. I think for most cases the validation in the schemas is stricter than Publishing API, but I think this case is special as there's either none or very loose schema validation for the unpublishing requests (which share the same redirect validation as this)

PR #1080, opened 2021-11-19
Add Account API to the list of rendering apps

huwd:
The Account API can be used to personalise pages for logged in users.
Following [Account API ADR 2][1] this is intended to approached as a
progressive enhancement via javascript in a browser.

As such these endpoints can now be called and will return payloads of
JSON, sometimes containing rendered HTML to be added into application
frontends.

As we wish to proxy these routes to account-api we will need to identify
account-api a rendering app.

[1]: https://github.com/alphagov/account-api/blob/main/docs/adr/002-progressive-enhancement.md

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -616,6 +616,7 @@
       "description": "The application that renders this item.",
       "type": "string",
       "enum": [
+        "account-api",
         "calculators",
         "calendars",
         "collections",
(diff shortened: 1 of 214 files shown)


alex9smith:
According to [the docs](https://github.com/alphagov/govuk-content-schemas/blob/ae16865a109d115bf96661b47b832c3978a10fb6/docs/changing-a-schema.md) you'll also need to run `rake` and commit the updated schemas for the tests to pass

alex9smith (approved):

PR #1087, opened 2022-03-14
Enable Additional Properties for specialist topic schema

Tetrino:
This will allow us to set a flag to determine if a topic was
created as a duplicated Mainstream Browse page for the ongoing
merge and migration away from said Browse pages.

As per usual, the main change is the .jsonnet, with other changes
being automatically generated.


chao-xian:
Great stuff! My only thought atm is that these eventually wouldn't be a _copy_ but I understand why we're trying to differentiate these from the other tree. Thoughts?

Tetrino:
> Great stuff! My only thought atm is that these eventually wouldn't be a _copy_ but I understand why we're trying to differentiate these from the other tree. Thoughts?

I agree, but it depends on how we're handling these after the fact. Are we intending on removing them entirely (migrate content off them into existing and new specialist topics) or are we intending on "switching them on" at some point and make them visible. If the latter, then yes a new name for the field would be best. `mainstream_browse_derived` perhaps?

Tetrino:
`mainstream_browse_type` is sensible, making the change now!

Tetrino:
@chao-xian 🎉 CI runs passing

chao-xian (approved):

PR #1089, opened 2022-04-08
Replace mainstream_browse_type with mainstream_browse_origin in Topic schema

AgaDufrat:
We want to reference the Content id of Mainstream browse page the topic originated from as a part of unified topic system work.

Until the work is completed we will need to keep the Mainstream browse page and
the cloned Topic is sync. Since other attributes of the Mainstream browse page
such as title or slug can change, this adds a unique reference we can use to
identify the cloned Topic.

The presence of this attribute can still acts as a flag.

It may also be useful to include reverse association in the Mainstream browse page schema to the cloned Topic's content_id.

### DIFF
--- a/dist/formats/topic/frontend/schema.json
+++ b/dist/formats/topic/frontend/schema.json
@@ -309,8 +309,9 @@
         "internal_name": {
           "$ref": "#/definitions/taxonomy_internal_name"
         },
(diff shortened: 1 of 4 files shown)


chao-xian (approved):
Looks good! I'd just add to the commit message/PR description a bit more context about why we need that reference.

PR #1090, opened 2022-04-20
Update manual examples

1pretz1:
This PR updates the manual related example schemas for frontend. This is needed as some schemas are missing data (links, timestamps etc), which should be present. I've also updated some of the data to make life easier in testing, which is taken advantage of in https://github.com/alphagov/government-frontend/pull/2424.

Trello:
https://trello.com/c/YmmLFXu5/1127-manuals-migrate-manuals-frontend-code-to-government-frontend

### DIFF
--- a/examples/hmrc_manual_section/frontend/vatgpb2000.json
+++ b/examples/hmrc_manual_section/frontend/vatgpb2000.json
@@ -9,7 +9,9 @@
   "details": {
     "body": "\n",
     "section_id": "VATGPB2000",
-    "breadcrumbs": [],
+    "breadcrumbs": [
+
+    ],
     "child_section_groups": [
       {
         "child_sections": [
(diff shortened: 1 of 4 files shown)


examples/hmrc_manual_section/frontend/vatgpb2000.json
```diff
@@ -1,55 +1,53 @@
 {
   "content_id": "2189bae1-a94e-4f90-917b-ffe25ec9ac4d",
-  "base_path": "/hmrc-internal-manuals/vat-government-and-public-bodies/vatgpb1000",
+  "base_path": "/hmrc-internal-manuals/vat-government-and-public-bodies/vatgpb2000",
```
edwardkerry, line 3: looks fine to me - but out of interest why does changing this to `2000` make it easier for testing?

edwardkerry (approved):

examples/hmrc_manual_section/frontend/vatgpb2000.json
1pretz1: Thanks for the review Ed and good question. It makes it easier as the section groups are ordered and can be navigated through with previous and next page links (example page: https://www.gov.uk/hmrc-internal-manuals/vat-government-and-public-bodies/vatgpb2000). As this was the first section (vatgpb1000), only the next link was generated from the tests. I could've altered the data directly but thought it would be cleaner to just update the schema example.

PR #1094, opened 2022-05-17
Add social media links to topical event schema

brucebolt:
We are in the process of migrating the rendering of Topical Events out of Whitehall.

As part of this, we need to include all the relevant information in the content item.

This will allow us to store social media links, in the same way we do for organisations.

[Trello card](https://trello.com/c/DOsCLwQL)


JonathanHallam (approved):
LGTM! Thanks

PR #1096, opened 2022-05-23
Add schema for html publication with history mode

davidgisbey:


davidgisbey:
The Whitehall build will only pass when https://github.com/alphagov/whitehall/pull/6585 is merged and deployed

beccapearce (approved):

PR #1099, opened 2022-05-30
Add 'about page link text' to Topical Event schema

brucebolt:
This text is needed to allow us to include a link (with custom text, as defined in Whitehall) in the topical event.

The page is currently rendered by Whitehall, but will be migrated to Collections, which is why this now needs to be included in the content item.

[Trello card](https://trello.com/c/zpFXy34H)


JonathanHallam (approved):
Exciting!

PR #1100, opened 2022-06-09
Add html publication schema with alternative national applicability url

davidgisbey:


ollietreend (approved):
Looks good to me. It's [exactly the same as a parent Publication](https://github.com/alphagov/govuk-content-schemas/blob/f2d94c81121feeee1283a0a3e2dc73306f0abd91/formats/publication.jsonnet#L74-L76), so that makes sense.

PR #1101, opened 2022-06-13
Add Topical Event image

brucebolt:
This is needed to allow us to include the topical event's logo or image in the content item.

The page is currently rendered by Whitehall, but will be migrated to Collections, which is why this now needs to be included in the content item.

[Trello card](https://trello.com/c/uNbOV24p)


BeckaL (approved):
LGTM 👍

PR #1104, opened 2022-06-20
Add emphasised organisations to topical event

callumknights:
This is needed to allow us to include emphasised organisations in the
content item for a topical event.

The page is currently rendered by whitehall, but will be rendered by
collections, which is why this now needs to be included.

Trello: https://trello.com/c/iIfXqVKi/71-add-rendering-of-topical-event-organisation-names-and-logos-to-collections


JonathanHallam (approved):
Looks good to me :)

PR #1106, opened 2022-07-07
Add title to manuals block of HMRC manuals section schema

KludgeKML:
Add manual title to HMRC manuals section schema, needed to render HMRC manuals more efficiently now they've been moved to government-frontend.

Note that this field will be required when tasks have been run to add titles to all existing and new documents.

https://trello.com/c/NToovfjr/1349-add-parent-manuals-title-to-the-hmrc-section-content-item

### DIFF
--- a/formats/shared/definitions/hmrc_manual.jsonnet
+++ b/formats/shared/definitions/hmrc_manual.jsonnet
@@ -95,4 +95,20 @@
       },
     },
   },
+  hmrc_manual_section_parent: {
+    description: "The parent manual for a manual section",
+    type: "object",
+    additionalProperties: false,
+    required: [
+      "base_path"
+    ],
+    properties: {
+      base_path: {
+        "$ref": "#/definitions/absolute_path",
+      },
+      title: {
+        type: "string",
+      }
+    },
+  },
 }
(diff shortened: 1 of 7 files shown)


formats/shared/definitions/hmrc_manual.jsonnet
```diff
@@ -95,4 +95,20 @@
       },
     },
   },
+  hmrc_manual_section_parent: {
+    description: "The parent manual for a manual section",
+    type: "object",
+    additionalProperties: false,
+    required: [
```
1pretz1, line 102: We should probably require the `title` too as it will be a required field when we actually start using it

1pretz1 (commented):
Changes look good, nice work Keith! I've only left one comment which should be quick to resolve.

Think the tests that are failing (collections) are flakey and probably need another kick for the to pass 🥾

formats/shared/definitions/hmrc_manual.jsonnet
KludgeKML: I explicitly didn't make it required in this step because it causes the search-api and hmrc-manuals-publisher tests to fail, but I've put a checklist step into the card to make it a required property once the other tasks are done (ie when all the existing data actually would actually adhere to the schema with that property required).

1pretz1 (approved):
LGTM!

PR #1107, opened 2022-07-18
Make title required for manual block of hmrc manual section.

KludgeKML:
Now that all HMRC Manual sections have title blocks (except for a small number that are broken for other reasons), and newly-made ones will have the title block, we can make this required in the schema so that the rendering app (government-frontend) can rely on it.

https://trello.com/c/NToovfjr/1349-add-parent-manuals-title-to-the-hmrc-section-content-item


1pretz1:
Changes look good 👍 We do have a failing test though, do we need to merge the [spec changes](https://github.com/alphagov/search-api/pull/2414) into this PR?

1pretz1 (approved):

PR #1110, opened 2022-07-27
Add format for world location news

BeckaL:
This was previously using a placeholder schema, but this PR introduces a schema for world location news.

There are several elements that are shared with organisations, which have been moved to the shared whitehall definitions so that both schemas can use the same definitions.

I have made the elements of the details hash mandatory. This doesn't seem to be a standard approach across our schemas, but making them mandatory does seem to be more helpful in a schema, guaranteeing the presence of these fields. This makes sense for the featured links and featured documents as they are always present (although can be empty, which is allowable by the schema). Mission statement can in theory be nil (they are nullable in the database and there are two world locations for which this is the case, but they are old) although I think the Whitehall UI currently always creates a mission statement, but it can be an empty string. In the whitehall work I've defaulted it to an empty string for these cases.

[Trello](https://trello.com/c/FUNRgTWI/168-get-world-news-content-into-content-item)

### DIFF
--- a/dist/formats/organisation/frontend/schema.json
+++ b/dist/formats/organisation/frontend/schema.json
@@ -463,46 +463,7 @@
           }
         },
         "ordered_featured_documents": {
-          "description": "A set of featured documents to display for the organisation. Turn into proper links once organisations are fully migrated.",
-          "type": "array",
-          "items": {
-            "type": "object",
-            "required": [
-              "title",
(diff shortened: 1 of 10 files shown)


brucebolt (approved):
Looks good to me 👍 .

Thanks for the explanation on why these fields are being made mandatory.

PR #1111, opened 2022-07-29
Remove facet groups and facet values

leenagupte:
Related to: 
* https://github.com/alphagov/finder-frontend/pull/2835
* https://github.com/alphagov/search-api/pull/2422

Trello: https://trello.com/c/EOLde8PE

`facet_groups` and `facet_values` were created to support the EU Exit Business Support Finder [1]. 
This finder was decommissioned in February 2020 [2], so support for these types is no longer needed.

[1]: https://github.com/alphagov/content-tagger/blob/main/docs/arch/adr-001-links-based-facets.md
[2]: https://github.com/alphagov/finder-frontend/pull/1949

### DIFF
--- a/dist/formats/answer/frontend/schema.json
+++ b/dist/formats/answer/frontend/schema.json
@@ -67,14 +67,6 @@
           "description": "Link type automatically added by Publishing API",
           "$ref": "#/definitions/frontend_links_with_base_path"
         },
-        "facet_groups": {
-          "description": "Prototype-stage metadata tagging tree roots for this content item. A content item my belong to many facet groups without having any specific facet_values links.",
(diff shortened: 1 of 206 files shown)


kevindew (approved):
Great stuff. Are we confident there is nothing still live in Publishing API that uses them? May need a bit of deleting if there is. Otherwise looks good.

leenagupte:
> Great stuff. Are we confident there is nothing still live in Publishing API that uses them? May need a bit of deleting if there is. Otherwise looks good.

Publishing API was updated 2.5 years ago 😬. See: https://github.com/alphagov/publishing-api/commit/2185b561ebdab1d15dc52ccc400bb7f6cb3ff7a6

PR #1114, opened 2022-08-17
Add location type to World Location news page

brucebolt:
This is required so we can add a label to the World Location News page telling users whether the news relates to a World Location or International Delegation.

[Trello card](https://trello.com/c/zxDhAkI3)

### DIFF
--- a/dist/formats/world_location_news/frontend/schema.json
+++ b/dist/formats/world_location_news/frontend/schema.json
@@ -484,6 +484,13 @@
         },
         "ordered_featured_links": {
           "$ref": "#/definitions/ordered_featured_links"
+        },
+        "world_location_news_type": {
+          "type": "string",
+          "enum": [
+            "international_delegation",
+            "world_location"
+          ]
         }
       }
     },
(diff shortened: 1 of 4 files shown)


BeckaL (approved):
👍

BeckaL (commented):
Thinking about this some more, I'm super confused re international delegation vs world location. It seems like even news pages that relate to international delegations get displayed as 'World Location News' e.g. https://www.gov.uk/world/uk-mission-to-asean/news. So is there any point in adding to the schema vs hard coding on the frontend?

I am extremely confused by all things world location news at the moment, so possibly this is an extremely confused comment, can discuss.

brucebolt:
> Thinking about this some more, I'm super confused re international delegation vs world location. It seems like even news pages that relate to international delegations get displayed as 'World Location News' e.g. https://www.gov.uk/world/uk-mission-to-asean/news. So is there any point in adding to the schema vs hard coding on the frontend?

As per our discussion yesterday, I've updated https://github.com/alphagov/whitehall/pull/6742 to not have the type hardcoded (which was a mistake 🤦‍♂️).

I think we'll need some product direction on how to deal with the international delegation pages, which appear to be identical to International Delegation news pages, but with additional contact details.

BeckaL (approved):
Cool, re-approving now I've understood everything a bit more 😄

PR #1125, opened 2022-09-29
Upgrade Ruby to 3.1.2

brucebolt:
Upgrades Ruby to 3.1.2. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.

[Trello](https://trello.com/c/0nKMZjma/312-upgrade-ruby-for-govuk-content-schemas)


lib/schema_generator/frontend_schema_generator.rb
```diff
@@ -81,13 +81,14 @@ def change_multiple_content_types(definitions)
     end
 
     def replace_multiple_content_types(object)
-      if object == { "$ref" => "#/definitions/multiple_content_types" }
+      case object
```
JonathanHallam, line 84: Seriously, rubocop is getting rid of if/elses for us now?

PR #1131, opened 2022-10-25
Add puma as a dependency

BeckaL:
This was removed from the standard lib in ruby 3. It's required for the preview app.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "rspec"
 
 # Preview app for examples
 gem "govuk_schemas"
+gem "puma"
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Code looks good, have you been able to test the web server starts or try it on heroku? (Aside: I wonder what could be automated)

BeckaL:
Thanks @kevindew  - yup I've spun it up locally. Kind of fortuitous timing as I'm looking at CD for content schemas and feel like at least a cursory check that this can be spun up as part of our test suite would be worth it.

PR #1133, opened 2022-11-07
Add additional links for International Delegations

brucebolt:
International Delegations have a number of additional links required to present their content outside of Whitehall.

Presentation of these links was added in https://github.com/alphagov/whitehall/pull/7023.

We do not need to add `organisations` as this is already present in the base links we inherit.

[Trello card](https://trello.com/c/4FWd2mRD/363-add-international-delegation-page-content-to-their-content-item)

### DIFF
--- a/formats/world_location_news.jsonnet
+++ b/formats/world_location_news.jsonnet
@@ -28,4 +28,8 @@
       },
     },
   },
+  links: (import "shared/base_links.jsonnet") + {
+    ordered_contacts: "Contact details for this World Location (only used for International Delegations)",
+    worldwide_organisations: "Linked Worldwide Organisations (only used for International Delegations)",
+  },
 }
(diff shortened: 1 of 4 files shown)


formats/world_location_news.jsonnet
```diff
@@ -28,4 +28,8 @@
       },
     },
   },
+  links: (import "shared/base_links.jsonnet") + {
+    ordered_contacts: "Contact details for this World Location",
```
BeckaL, line 32: Could we signpost in the description here that this is only used for international delegations? I'm thinking forward to a future in which world location news != international delegation

BeckaL (approved):
LGTM, just a small comment here

formats/world_location_news.jsonnet
brucebolt: Good idea.  I've added this to the comment.

PR #1134, opened 2022-11-09
Remove mainstream browse origin from Topic schema

AgaDufrat:
It's no longer needed as we will not be merging the two topic systems.

[Trello](https://trello.com/c/4CjEsirD/117-remove-code-related-to-topic-merging)


hannako (approved):
I think @BeckaL is migrating this app. So maybe check in with her before merging. Not sure on status.

AgaDufrat:
Checked with @BeckaL who confirmed we can go ahead and merge it. The change will be copied it over into publishing api as content schemas are not quite ready to freeze yet.

PR #1135, opened 2022-11-16
Remove dummy content store

BeckaL:
We are retiring the dummy content store - this was an app deployed on heroku used by a couple of frontend apps to aid reviewing PRs, since it would provide some dummy content to check rendering against. However, the live content store can instead be used for this purpose, and this is how the majority of apps worked.

We recently completed the work to ensure that the last two apps using this dummy content store are no longer using it - see merged PRs [here](https://github.com/alphagov/government-frontend/pull/2591) and [here](https://github.com/alphagov/frontend/pull/3423). The application has also been manually deleted in heroku for the last week, so this code should be safe to remove.

[Trello](https://trello.com/c/z08hU3Ov/360-retire-the-dummy-content-store)


JonathanHallam (approved):
Thanks for this, incredibly tiny comment, PR description says it was a heroku app but it's really a sinatra app. There isn't a pedantic emoji but I would like to have one for this message 🙈

BeckaL:
Ha, a pedantic emoji woudl be great
> PR description says it was a heroku app but it's really a sinatra app
Sorry, what I meant by 'heroku app' was that it was deployed on heroku as opposed to what framework it was using. Will update

